PR #399, opened 2014-11-19
Add a shouldSkipBulk for string metrics

pquerna:
this is more a proof of concept than a real PR... I mean I coded it in sublime text... so much bad formatting. much TODO left.


chinmay-gupte:
Closing this because of lack of activity. @pquerna thanks for providing a proof of concept. But we changed the trajectory for the string metrics, and this might not be relevant anymore. Please re-open if you think I am wrong.

PR #410, opened 2014-12-24
Add repairStatus.sh

stackedsax:
We needed a way to check on the status of various repairs.  The best way seemed to be to ask Elasticsearch for how many range repairs were complete out of the total.


contrib/repairStatus.sh
```diff
@@ -0,0 +1,74 @@
...
+    echo "Node and a Hostname for Elasticsearch is required"
+    exit 2 
+fi 
+
+shift "$((OPTIND-1))" # Shift off the options and optional --.
+
+NOW=$(date +%s000)
+START=$(date -v-${DAYS}d +%s000)
+
+TOTAL_SUCCESSFUL_RANGE_REPAIRS=$(curl -s -XGET 'http://'$ES_HOST'/_count' -d '{"query":{"filtered":{"query":{"bool":{"should":[{"query_string":{"query":"level:\"info\" AND (*)"}}]}},"filter":{"bool":{"must":[{"range":{"@timestamp":{"from":'$START',"to":'$NOW'}}},{"fquery":{"query":{"query_string":{"query":"host:(\"'${NODE}'\")"}},"_cache":true}},{"fquery":{"query":{"query_string":{"query":"method:(\"syncComplete\")"}},"_cache":true}},{"fquery":{"query":{"query_string":{"query":"message:(\"is fully synced\")"}},"_cache":true}}]}}}}}'|  grep -oP 'count\"\:\K[[:digit:]]+(?=,)')
```
GeorgeJahad: fwiw: i wasn't able to find/build a version of grep that supports PCRE on newer versions of osx.

stackedsax: @GeorgeJahad: yeah, and the date function only has the -v option on Macs, so I rewrote the thing in python.  See if it works better now.

contrib/repairStatus.py
```diff
@@ -0,0 +1,199 @@
...
+                }
+            }
+        }
+    }
+  }
+
+  return json.dumps(query_json)
+
+
+# Execute an ES _count to return the number of ranges that have finished repairing
```
glyph: Rather than a comment before the function, the general convention is to put a docstring in the function itself, like

```python
def calculate_range_repairs(host, port, node, start_time, current_time):
    """
    Execute an ES _count to return the number of ranges that have finished repairing
    """
```

contrib/repairStatus.py
```diff
@@ -0,0 +1,199 @@
...
+import json
+import argparse
+import requests
+import re
+import logging as log
+from datetime import datetime, timedelta
+
+
+def generate_range_repair_query (node, start_time, current_time):
+  query_json = {
```
glyph: Definitely go with a 4-space indent :).  "pip install pep8" can tell you about other style-guide stuff.

contrib/repairStatus.py
```diff
@@ -0,0 +1,199 @@
+from __future__ import print_function
```
glyph: another good `__future__` import is `unicode_literals`, which makes all of your quoted strings be really actual strings, and not bytes-masquerading-as-strings (which is what "str" is in python 2).  This might necessitate fixing up some things in your code, but those are almost always good things to fix. Given that 99% of your quoted strings here are in JSON structures, and JSON structures are really supposed to be unicode anyway, this is probably the right way to go.

(You can still type bytes literals with the b-prefix syntax, like `b"foo"`.)

contrib/repairStatus.py
```diff
@@ -0,0 +1,199 @@
...
+
+  # Calculate how many 
+  total_successful_range_repairs = calculate_range_repairs(args.host, args.port, args.node, start_time, current_time)
+  total_to_repair = calculate_total_ranges(args.host, args.port, args.node, start_time, current_time)
+
+  # Print totals
+  print_totals(args.node, args.days, total_successful_range_repairs, total_to_repair)
+
+
+if __name__ == '__main__': main()
```
glyph: also a minor coding-style nitpick; most projects have a "no simple suites" rule (a bunch of statements following a ":" is called a "suite") so this would normally be written:
```python
if __name__ == '__main__':
    main()
```

contrib/repairStatus.py
```diff
@@ -0,0 +1,199 @@
...
+      }
+    }
+  }
+
+  return json.dumps(query_json)
+
+
+# Execute an ES _search to return the total number of ranges that need repairing
+def calculate_total_ranges(host, port, node, start_time, current_time):  
+  url = 'http://%(host)s:%(port)s/_search' % dict(host=host, port=str(port))
```
glyph: The new-style "format" syntax is a little less error prone than the old-style "%" syntax.  It would look like this:

```python
'http://{host}:{port}/_search'.format(host=host, port=str(port))
```

Notice that you don't need a "s" after the closing braces there - this way the brace-matching in your editor is sufficient to make sure the syntax is right.

contrib/repairStatus.py
```diff
@@ -0,0 +1,199 @@
...
+  if r.status_code == 200:
+    response = json.loads(r.text)
+    message = json.dumps(response['hits']['hits'][0]['fields']['message'])    
+    total_ranges = int(re.search('(?<=repairing\s)\d+(?=\sranges)', message).group(0))
+    return total_ranges
+
+
+def print_totals(node, days, total_successful_range_repairs, total_to_repair):
+  percent_complete = total_successful_range_repairs * 100 / total_to_repair
+  print ("Checking repair messages for", node, "for the past", days, "days")
```
glyph: Another minor coding-style thing: generally a space before the paren on a function call is discouraged.

contrib/repairStatus.py
```diff
@@ -0,0 +1,199 @@
...
+  
+  args = parser.parse_args()
+
+  if args.verbose:
+      log.basicConfig(format="%(levelname)s: %(message)s", level=log.DEBUG)
+  else:
+      log.basicConfig(format="%(levelname)s: %(message)s")
+  
+  # Set the current time and the point in time to start searching from
+  current_time=datetime.utcnow().strftime('%Y-%m-%dT%H:%M:%S.000Z')
```
glyph: It's not really important for this particular code since it's dealing with such big ranges, but one thing to keep in mind when generating multiple subsequent timestamps is that you should generally call `utcnow` only once and store it in a temporary.

I also did a little digging and found that you *don't* in fact need to copy and paste that hideous format string all over the place; `datetime` has a handy `isoformat()` method.

```
>>> datetime.datetime.utcnow().isoformat()
'2015-01-09T08:32:07.112879'
```

contrib/repairStatus.py
stackedsax: ah, nice.  I had looked at .format, but Dominic had suggested the other way and I just went with it.

contrib/repairStatus.py
stackedsax: and it looks weird.  Who does that, even?

contrib/repairStatus.py
```diff
@@ -0,0 +1,199 @@
...
+          }
+        },
+        "query": {
+          "query_string": {
+            "query": 'message: "Starting repair"'
+          }
+        }
+      }
+    }
+  }
```
glyph: This is more of a personal taste thing, nothing to do with Python style, but I found it much easier to read to compress this into fewer lines. Emacs helped a lot with this, but what I came up with was:
```python
    query_json = {
        "fields": "message",
        "query":
        {"filtered":
         {"filter":
          {"bool": {"must": [
              {"range":
               {"@timestamp": {"from": start_time, "to": current_time}}},
              {"fquery": {"_cache": True,
                          "query": {"query_string":
                                    {"query": 'host:("' + node + '")'}}}}
          ]}},
      "query": {"query_string": {"query": 'message: "Starting repair"'}}}}
    }
```
This one you can really take or leave :).

contrib/repairStatus.py
```diff
@@ -0,0 +1,199 @@
...
+  percent_complete = total_successful_range_repairs * 100 / total_to_repair
+  print ("Checking repair messages for", node, "for the past", days, "days")
+  print ("Count is: " + str(total_successful_range_repairs))
+  print ("Total to repair is: " + str(total_to_repair))
+  print ("Percent Complete: " + str(percent_complete) + "%")
+
+
+def main():
+  parser = argparse.ArgumentParser(
+    description = '''This script shows the current status of a repair running on a Cassandra node.
```
glyph: Since this is a keyword argument and not a variable being assigned, it is conventional to avoid the spaces around `=` and just have it be `description='''`.

contrib/repairStatus.py
```diff
@@ -0,0 +1,199 @@
...
+def calculate_range_repairs(host, port, node, start_time, current_time):
+  url = 'http://%(host)s:%(port)s/_count' % dict(host=host, port=port)
+  query = generate_range_repair_query(node, start_time, current_time)
+  
+  log.info("Ranges completed: curl %(url)s -d '%(query)s'" % dict(url=url, query=query)) 
+  r = requests.get(url, data=query)
+  
+  if r.status_code == 200:
+    response = json.loads(r.text)
+    return response['count']
```
glyph: Checking the return code this way means that when the remote end encounters any sort of error, you won't see what it was.  I think it would be better to just blow up, like `raise ValueError("invalid status code returned: {code} {text}".format(code=response.code, text=response.text)).  This way you have something to look at if the query goes wrong because the cluster is in bad shape, and more diagnostics to act on later.

contrib/compactionEta.sh
```diff
@@ -0,0 +1,49 @@
+#/bin/bash
```
glyph: It would be nice if this one were Python too :).

glyph:
Hopefully my style-guide comments were helpful.  The main comment though - no unit tests! :).  Would it make sense to put this into a separate repository so it could have separate CI?  I have set up 3 projects in the last couple of months so I'm starting to get pretty good at rolling out a Python template with travis, readthedocs, and so on.

glyph:
One other thought before I'm done - would it be easier to track and graph the output from this command if it itself produced a structured format, like JSON?  It might be nice to have a command-line flag to specify that instead of english text output, for easy parsing and future processing.

contrib/compactionEta.sh
stackedsax: And get rid of ```sort -n -k 5 | tail -1 | awk '{print $7}' | tr -d %```???  Whyever?!

stackedsax:
Yeah, style comments were super helpful.  

Regarding tests, I was researching the best way to mock out Elasticsearch.  I was planning to use something like HTTPretty to just spit back some fake responses from Elasticsearch. Suggestions?  But yes, I'm working on more mundane tests.

Regarding output, I think that's lower on my priority list than a couple other features, but it's simple enough.  I'll add that to my list.

More important, I believe, is to identify the timestamp of the most recent 'total ranges' message in Elasticsearch and then search from that point forward for the total number of successful repairs.  In addition, I need to add a search for repair failures so that the calculation of range repairs that have completed is more accurate.

glyph:
My suggestion with testing would always be to go with the simplest thing first, so you have some testing before really working out something elaborate.  Unit tests before integration tests.  So, for a first pass, just copy/paste some JSON into a test file, and fake out the requests library calls to just return that canned data and test against that.  I assume this is what you already meant by "more mundane tests".

As far as mocking - spinning up an ELK in a Docker container is probably a tractable thing for integration testing, I'd probably jump straight to that rather than set up a service like HTTPretty, HTTPBin or Mimic, since you mostly would want to verify against a reali-ish API for the second layer of integration testing anyway, and as fiddly as elasticsearch can be there are some good community hub images for getting it running fast and repeatably.

As far as my suggestion about structured output, that was scope creep, sorry :smile: . I should have suggested it explicitly be a separate issue.

contrib/repairStatus.py
glyph: As per my comment about scope creep below, adding `unicode_literals` is another good candidate for later improvement, probably not something you should try to do before landing this.

contrib/compactionEta.sh
```diff
@@ -0,0 +1,49 @@
...
+#    your hosts like my team does, but hopefully you get the idea.
+#
+#       for x in {00..31}; do ssh db$(printf "%2.2d" $x).domain.com 'bash -s' < ./compactionEta.sh [port]; done
+#
+# c. Or, to run on multiple machines a little faster, install GNU parallel 
+#    and run the command below. Again, replace 'domain.com' and also <start> 
+#    and <end> with the first and last servers you want to check.
+# 
+#       parallel -j20 --no-notice "ssh db{1}.domain.com 'bash -s' < ./compactionEta.sh [port]" ::: $(seq -w <start> <end>) | sort
+
```
mburns: `set -euo pipefail`, for conservative behavior (fail on non-zero command exit, fail on undefined variable, etc)?

mburns:
The BASH script reads ok to me. There are places [shellcheck](http://www.shellcheck.net/#) recommends quoting variables, to be conservative, and we could rewrite `ps auxx` for a `pgrep`, but I don't see either as being a terribly high priority.

:+1: from me fwiw

usnavi:
@stackedsax  - this is over a year old and now needs to be merged.  Shall we close this?

usnavi:
@stackedsax - this PR is over a year old?  May we close it?

PR #412, opened 2015-01-07
WIP: Metadata cache optimization

tilogaat:
Currently, Blueflood stores the following metadata for numeric and string metrics:
1. Type
2. Unit
3.  RollupType

However for numeric metrics, which is our most common use case, we do not want to store metadata information in the metadata cache and infer it automatically ie if type and rollup type are null, we assume a numeric type, Basic rollup type respectively. Units will be obtained from ElasticSearch.

### DIFF
--- a/blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/io/AstyanaxReaderIntegrationTest.java
+++ b/blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/io/AstyanaxReaderIntegrationTest.java
@@ -22,12 +22,10 @@ import com.rackspacecloud.blueflood.rollup.Granularity;
 import com.rackspacecloud.blueflood.types.*;
 import org.junit.Assert;
 import org.junit.Test;
-import org.junit.rules.ExpectedException;
 
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
-import java.util.UUID;
 
 public class AstyanaxReaderIntegrationTest extends IntegrationTestBase {
     
--- a/blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/io/AstyanaxReaderIntegrationTest.java
+++ b/blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/io/AstyanaxReaderIntegrationTest.java
@@ -22,10 +22,12 @@ import com.rackspacecloud.blueflood.rollup.Granularity;
 import com.rackspacecloud.blueflood.types.*;
 import org.junit.Assert;
 import org.junit.Test;
+import org.junit.rules.ExpectedException;
 
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.UUID;
 
 public class AstyanaxReaderIntegrationTest extends IntegrationTestBase {
     
@@ -93,5 +95,46 @@ public class AstyanaxReaderIntegrationTest extends IntegrationTestBase {
             MetricData metrics = results.get(locator);
             Assert.assertEquals(1, metrics.getData().getPoints().size());
         }
+    }
+
+    @Test
+    public void testCanRetrieveNumericMetricsEvenIfNoMetaDataStored() throws Exception {
+        // Write metrics and also persist their types.
+        List<Locator> locatorList = new ArrayList<Locator>();
+        Metric metric = writeMetric("string_metric", "version 1.0.43342346");
+        MetadataCache.getInstance().put(metric.getLocator(), MetricMetadata.TYPE.name().toLowerCase(), DataType.STRING.toString());
+        locatorList.add(metric.getLocator());
+
+        metric = writeMetric("int_metric", 45);
+        locatorList.add(metric.getLocator());
+
+        metric = writeMetric("long_metric", 67L);
+        locatorList.add(metric.getLocator());
+
+        // Test batch reads
+        AstyanaxReader reader = AstyanaxReader.getInstance();
+        Map<Locator, MetricData> results = reader.getDatapointsForRange(locatorList, new Range(metric.getCollectionTime() - 100000,
+                metric.getCollectionTime() + 100000), Granularity.FULL);
+
+        Assert.assertEquals(locatorList.size(), results.size());
+
+        for (Locator locator : locatorList) {
+            MetricData metrics = results.get(locator);
+            Assert.assertEquals(1, metrics.getData().getPoints().size());
+        }
+    }
+
+    @Test
+    public void test_StringMetrics_WithoutMetadata_NotRetrieved() throws Exception {
+        List<Locator> locatorList = new ArrayList<Locator>();
+        Metric metric = writeMetric("string_metric", "version 1.0.43342346");
+        locatorList.add(metric.getLocator());
+
+        // Test batch reads
+        AstyanaxReader reader = AstyanaxReader.getInstance();
+        Map<Locator, MetricData> results = reader.getDatapointsForRange(locatorList, new Range(metric.getCollectionTime() - 100000,
+                metric.getCollectionTime() + 100000), Granularity.FULL);
+
+        Assert.assertEquals(0, results.size());
     }
 }
--- a/blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/io/AstyanaxReaderIntegrationTest.java
+++ b/blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/io/AstyanaxReaderIntegrationTest.java
@@ -127,7 +127,7 @@ public class AstyanaxReaderIntegrationTest extends IntegrationTestBase {
     @Test
     public void test_StringMetrics_WithoutMetadata_NotRetrieved() throws Exception {
         List<Locator> locatorList = new ArrayList<Locator>();
-        Metric metric = writeMetric("string_metric", "version 1.0.43342346");
+        Metric metric = writeMetric("string_metric_1", "version 1.0.43342346");
         locatorList.add(metric.getLocator());
 
         // Test batch reads
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxIO.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxIO.java
@@ -108,12 +108,16 @@ public class AstyanaxIO {
     }
 
     protected AbstractSerializer serializerFor(RollupType rollupType, DataType dataType, Granularity gran) {
-        if (dataType.equals(DataType.STRING)) {
-            return StringSerializer.get();
-        } else if (dataType.equals(DataType.BOOLEAN)) {
-            return BooleanSerializer.get();
-        } else {
+       try {
+           if (dataType.equals(DataType.STRING)) {
+               return StringSerializer.get();
+           } else if (dataType.equals(DataType.BOOLEAN)) {
+               return BooleanSerializer.get();
+           } else {
+               return NumericSerializer.serializerFor(RollupType.classOf(rollupType, gran));
+           }
+       } catch (Exception ex) {
            return NumericSerializer.serializerFor(RollupType.classOf(rollupType, gran));
-        }
+       }
     }
 }
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxIO.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxIO.java
@@ -108,16 +108,16 @@ public class AstyanaxIO {
     }
 
     protected AbstractSerializer serializerFor(RollupType rollupType, DataType dataType, Granularity gran) {
-       try {
-           if (dataType.equals(DataType.STRING)) {
-               return StringSerializer.get();
-           } else if (dataType.equals(DataType.BOOLEAN)) {
-               return BooleanSerializer.get();
-           } else {
-               return NumericSerializer.serializerFor(RollupType.classOf(rollupType, gran));
-           }
-       } catch (Exception ex) {
-           return NumericSerializer.serializerFor(RollupType.classOf(rollupType, gran));
-       }
+        if (dataType == null) {
+            return NumericSerializer.serializerFor(RollupType.classOf(rollupType, gran));
+        }
+
+        if (dataType.equals(DataType.STRING)) {
+            return StringSerializer.get();
+        } else if (dataType.equals(DataType.BOOLEAN)) {
+            return BooleanSerializer.get();
+        } else {
+            return NumericSerializer.serializerFor(RollupType.classOf(rollupType, gran));
+        }
     }
 }
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
@@ -43,7 +43,6 @@ import com.rackspacecloud.blueflood.types.*;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import javax.xml.crypto.Data;
 import java.io.IOException;
 import java.util.*;
 
@@ -335,7 +334,6 @@ public class AstyanaxReader extends AstyanaxIO {
             if (rollupType == null) {
                 rollupType = RollupType.BF_BASIC;
             }
-
             if (type == null) {
                 return getNumericOrStringRollupDataForRange(locator, range, gran, rollupType);
             }
@@ -372,11 +370,9 @@ public class AstyanaxReader extends AstyanaxIO {
             try {
                 RollupType rollupType = RollupType.fromString((String)
                             metaCache.get(locator, MetricMetadata.ROLLUP_TYPE.name().toLowerCase()));
-
                 if (rollupType == null) {
                     rollupType = RollupType.BF_BASIC;
                 }
-
                 //TODO: If we stop processing string and boolean, we can always hardcode this to numeric
                 DataType dataType = getDataType(locator, MetricMetadata.TYPE.name().toLowerCase());
 
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
@@ -222,7 +222,7 @@ public class AstyanaxReader extends AstyanaxIO {
         return columns == null ? new EmptyColumnList<Long>() : columns;
     }
 
-    private Map<Locator, ColumnList<Long>>  getColumnsFromDB(List<Locator> locators, ColumnFamily<Locator, Long> CF,
+    private Map<Locator, ColumnList<Long>> getColumnsFromDB(List<Locator> locators, ColumnFamily<Locator, Long> CF,
                                                             Range range) {
         if (range.getStart() > range.getStop()) {
             throw new RuntimeException(String.format("Invalid rollup range: ", range.toString()));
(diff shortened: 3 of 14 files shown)


blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/io/AstyanaxReaderIntegrationTest.java
```diff
@@ -22,10 +22,12 @@
 import com.rackspacecloud.blueflood.types.*;
 import org.junit.Assert;
 import org.junit.Test;
+import org.junit.rules.ExpectedException;
```
chinmay-gupte: NIT: unused import?

blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/io/AstyanaxReaderIntegrationTest.java
```diff
@@ -22,10 +22,12 @@
...
 
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.UUID;
```
chinmay-gupte: Ditto

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxIO.java
```diff
@@ -108,12 +108,16 @@ protected static Keyspace getKeyspace() {
...
+            return NumericSerializer.serializerFor(RollupType.classOf(rollupType, gran));
+        }
+
         if (dataType.equals(DataType.STRING)) {
             return StringSerializer.get();
         } else if (dataType.equals(DataType.BOOLEAN)) {
             return BooleanSerializer.get();
         } else {
-           return NumericSerializer.serializerFor(RollupType.classOf(rollupType, gran));
+            return NumericSerializer.serializerFor(RollupType.classOf(rollupType, gran));
```
chinmay-gupte: NIT: redundant space.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
```diff
@@ -43,6 +43,7 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import javax.xml.crypto.Data;
```
chinmay-gupte: Unused again?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
```diff
@@ -222,7 +223,7 @@ public String getLastStringValue(Locator locator) {
         return columns == null ? new EmptyColumnList<Long>() : columns;
     }
 
-    private Map<Locator, ColumnList<Long>> getColumnsFromDB(List<Locator> locators, ColumnFamily<Locator, Long> CF,
+    private Map<Locator, ColumnList<Long>>  getColumnsFromDB(List<Locator> locators, ColumnFamily<Locator, Long> CF,
```
chinmay-gupte: Extra space.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
```diff
@@ -364,9 +371,19 @@ public MetricData getDatapointsForRange(Locator locator, Range range, Granularit
...
-                        metaCache.get(locator, MetricMetadata.TYPE.name().toLowerCase()));
+                            metaCache.get(locator, MetricMetadata.ROLLUP_TYPE.name().toLowerCase()));
+
+                if (rollupType == null) {
+                    rollupType = RollupType.BF_BASIC;
+                }
+
+                //TODO: If we stop processing string and boolean, we can always hardcode this to numeric
+                DataType dataType = getDataType(locator, MetricMetadata.TYPE.name().toLowerCase());
+
```
chinmay-gupte: NIT: extra blank lines around the if block.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/IncomingMetricMetadataAnalyzer.java
```diff
@@ -72,6 +72,7 @@ private IncomingMetricException checkMeta(Locator locator, String key, String in
             String existing = cache.get(locator, key, String.class);
 
             // always update the cache. it is smart enough to avoid needless writes.
+            //if (dataType.equals(DataType.STRING) }})
```
chinmay-gupte: @tilogaat can you explain why this has been commented?

chinmay-gupte:
@tilogaat Can you clean up the extra spaces and new lines introduced before making the ready?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/IncomingMetricMetadataAnalyzer.java
tilogaat: it was something that I was trying, will get rid of it

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionServer.java
```diff
@@ -172,7 +172,7 @@ private void buildProcessingChains() {
 
         this.defaultProcessorChain = AsyncChain
                 .withFunction(typeAndUnitProcessor)
-                .withFunction(rollupTypeCacher)
+                //.withFunction(rollupTypeCacher)
                 .withFunction(batchSplitter)
```
GeorgeJahad: Why is this line commented out?  Are you getting rid of the rollupTypeCacher too?

tilogaat: @GeorgeJahad We are not caching the rollup type for numeric and string metrics. But we will continue to cache the rollup type in the statsdProcessorChain. Does that make sense?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxIO.java
tilogaat: The earlier spacing was wrong. It's fixed now.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
```diff
@@ -484,6 +486,16 @@ private MetricData transformColumnsToMetricData(Locator locator, ColumnList<Long
...
     }
 
+    private DataType getDataType(Locator locator, String dataTypeCacheKey) throws CacheException{
+        String meta = metaCache.get(locator, dataTypeCacheKey);
+        DataType dataType = null;
+        if (meta != null) {
+            dataType = new DataType(meta);
+        }
+
+        return dataType;
```
chinmay-gupte: May be consolidate this as,

```
String meta = ...
if (meta != null)
    return new DataType(meta);
return null;

File: blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java [hunk 10]
```diff
@@ -398,7 +398,7 @@ public MetricData getHistogramsForRange(Locator locator, Range range, Granularit
 
         ColumnFamily cf = CassandraModel.getColumnFamily(HistogramRollup.class, granularity);
         Points<HistogramRollup> histogramRollupPoints = getDataToRoll(HistogramRollup.class, locator, range, cf);
-        return new MetricData(histogramRollupPoints, getUnitString(locator), MetricData.Type.HISTOGRAM);
+        return new MetricData(histogramRollupPoints, null, MetricData.Type.HISTOGRAM);
```
chinmay-gupte: @tilogaat We should refactor `MetricData` to stop taking `unit` as a field, so that we can remove it from its constructor as well.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
```diff
@@ -298,19 +298,6 @@ public String getLastStringValue(Locator locator) {
         return points;
     }
 
-    public static String getUnitString(Locator locator) {
```
chinmay-gupte: @tilogaat Why are we removing this method? I thought the users will still have an option of getting the unit from `MetadataCache`, if they do not have ES?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
```diff
@@ -415,7 +415,7 @@ private MetricData getStringMetricDataForRange(Locator locator, Range range, Gra
             }
         }
 
-        return new MetricData(points, getUnitString(locator), MetricData.Type.STRING);
+        return new MetricData(points, null, MetricData.Type.STRING);
```
chinmay-gupte: Ditto for all the places where MetricData object gets initialized.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
```diff
@@ -364,9 +356,17 @@ public MetricData getDatapointsForRange(Locator locator, Range range, Granularit
...
-                        metaCache.get(locator, MetricMetadata.TYPE.name().toLowerCase()));
+                            metaCache.get(locator, MetricMetadata.ROLLUP_TYPE.name().toLowerCase()));
+                if (rollupType == null) {
+                    rollupType = RollupType.BF_BASIC;
+                }
+                //TODO: If we stop processing string and boolean, we can always hardcode this to numeric
+                DataType dataType = getDataType(locator, MetricMetadata.TYPE.name().toLowerCase());
+
+                if (dataType == null) {
+                    dataType = DataType.INT;
```
chinmay-gupte: 2:22 PM <chinmay_g_> tilogaat: this seems to be really hacky https://github.com/rackerlabs/blueflood/pull/412/files#diff-d3454fc0815b5e706f9fabec298fdc48R367
2:23 PM <tilogaat> chinmay_g_: you mean not returning the locator if number of datapoints is 0
2:23 PM <tilogaat> ?
2:23 PM <chinmay_g_> because data we are going to retrieve can be Long, Double or BigInt as well
2:23 PM <chinmay_g_> no, sec, I will explain
2:24 PM <chinmay_g_> dominiclobue: can you make the rollup thread settings ASAP, we are still falling back on rollup work.....
2:25 PM <tilogaat> chinmay_g_: yes I agree, but it does not really matter because the serializers are self describin
2:25 PM <tilogaat> they dont depend on the datatype to return the serializer
2:25 PM <tilogaat> datatype is just used to return the serializer-deserializer for string or bool
2:26 PM <chinmay_g_> yeah, should we refactor the code then to have just have an umbrella Numeric type?
2:26 PM <tilogaat> maybe i can add tests to show that even if data type is int, we return the correct serializer
2:26 PM <tilogaat> chinmay_g_: yes, we can 
2:26 PM <chinmay_g_> like there will be just 3 types 1) Numeric 3) String 3) Boolean
2:26 PM <chinmay_g_> cool
2:27 PM <chinmay_g_> tilogaat: this is turning into more contrived task than we imagined, no?
2:27 PM <chinmay_g_> do you want me to add some commits?
2:27 PM <tilogaat> chinmay_g_: yes .. i was thinking of just making another PR and remove the rollup type in that 
2:27 PM <tilogaat> so that its easier to deploy and review 
2:27 PM <tilogaat> then anotehr for units and anotehr for type
2:27 PM <tilogaat> basically cherry pick the commits but deploy and review them separately
2:28 PM <chinmay_g_> tilogaat: yea, that is a good idea...
2:28 PM <tilogaat> ok i will do that

blueflood-core/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/RollupHandler.java
```diff
@@ -206,4 +268,28 @@ protected MetricData getHistogramsByGranularity(String tenantId,
 
         return data;
     }
+
+    private class UnitsGetter implements Callable {
```
chinmay-gupte: Its totally up to you, but `UnitsFetcher` seems nicer and prettier than `UnitsGetter`?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/RollupHandler.java
```diff
@@ -206,4 +268,28 @@ protected MetricData getHistogramsByGranularity(String tenantId,
...
+        public UnitsGetter(String tenantId, String metricName) {
+            this.tenantId = tenantId;
+            this.metricName = metricName;
+        }
+
+        public String call() {
+            String unit = null;
+            try {
+                List<SearchResult> results = discoveryHandler.search(tenantId, metricName);
+                for (SearchResult res : results) {
```
chinmay-gupte: @tilogaat instead of this iteration, can we just do `results.get(0).getUnit()`? Also, it will be good to assert in some way that the results' size is always 1, may be have a block of code which tests that?

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionServer.java
chinmay-gupte: @tilogaat Lets remove the line completely.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/RollupHandler.java
```diff
@@ -53,20 +60,51 @@
     protected final Histogram numHistogramPointsReturned = Metrics.histogram(RollupHandler.class, "Histogram points returned");
 
     private static final boolean ROLLUP_REPAIR = Configuration.getInstance().getBooleanProperty(CoreConfig.REPAIR_ROLLUPS_ON_READ);
+    protected final DiscoveryIO discoveryHandler = loadDiscoveryModule();
```
chinmay-gupte: We have this cruft code in a lot of places, which loads the discovery modules. May be we should have a separate class which does this for us or something.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/RollupHandler.java
```diff
@@ -128,13 +166,37 @@ protected MetricData getRollupByGranularity(
         return metricData;
     }
 
+    private MetricData getMetricData(long from, long to, Granularity g, Locator locator) {
+
+        String unit = null;
+        Callable<String> unitsGetter = new UnitsGetter(locator.getTenantId(),locator.getMetricName());
```
chinmay-gupte: @tilogaat what do you think about consolidating this into,
```
String unit = executorService.submit(new Callable () {
            String unitInt;

            @Override
            public Object call() throws Exception {
                try {
                    List<SearchResult> results = discoveryHandler.search(locator.tenantId(), locator.getMetricName();
                    for (SearchResult res : results) {
                        unitInt = res.getUnit();
                        break;
                    }
                } catch (Exception ex) {
                    log.info(ex.getMessage());
                }
                return unitInt;

            }
        }).get();
```

blueflood-core/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/RollupHandler.java
```diff
@@ -206,4 +268,28 @@ protected MetricData getHistogramsByGranularity(String tenantId,
...
+        public String call() {
+            String unit = null;
+            try {
+                List<SearchResult> results = discoveryHandler.search(tenantId, metricName);
+                for (SearchResult res : results) {
+                    unit = res.getUnit();
+                    break;
+                }
+            } catch (Exception ex) {
+                log.info(ex.getMessage());
```
chinmay-gupte: May be we should just let the callable throw an exception and catch it in the block around `.get()`

blueflood-core/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/RollupHandler.java
```diff
@@ -128,13 +166,37 @@ protected MetricData getRollupByGranularity(
...
+
+        MetricData metricData;
+        metricData = AstyanaxReader.getInstance().getDatapointsForRange(
+                locator,
+                new Range(g.snapMillis(from), to),
+                g);
+
+        try {
+            unit = unitsGetterFuture.get();
+        } catch (InterruptedException ie) {
```
chinmay-gupte: If there are problems while getting units, out, should be set it to `UNKNOWN` instead of null or may be just not return unit? What do you think?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/RollupRunnable.java
```diff
@@ -65,7 +69,38 @@ public RollupRunnable(RollupExecutionContext executionContext, SingleRollupReadC
         this.rollupBatchWriter = rollupBatchWriter;
         startWait = System.currentTimeMillis();
     }
-    
+
+    private static DiscoveryIO loadDiscoveryModule() {
```
chinmay-gupte: Loading the discovery module, seems to be a bit out of place in `RollupRunnable`...can you look into if we can refactor `RollupEvent` and  REE so that we do a check if there are current subscribers to the rollup event and get units from MetadataCache, (yes, without concerning about ES in that case) there?
Let me know if you want some more explanation.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
```diff
@@ -431,7 +431,7 @@ private MetricData getBooleanMetricDataForRange(Locator locator, Range range, Gr
             }
         }
 
-        return new MetricData(points, getUnitString(locator), MetricData.Type.BOOLEAN);
+        return new MetricData(points, null, MetricData.Type.BOOLEAN);
```
dlobue: this seems like a bad code smell... if the unit isn't required, perhaps remove it from the constructor and have it optionally supplied via a method?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
dlobue: I'm kinda in favor of keeping the curly brackets: https://www.imperialviolet.org/2014/02/22/applebug.html

blueflood-core/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/RollupHandler.java
dlobue: We do this exact same thing for all our modular code. We should look into using a real configuration library, and into a real plugin system.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/RollupHandler.java
dlobue: this whole bit is going to be incredibly slow! all these searches will kill the cluster! I'm not quite sold on using ES for the fetching of metadata, but if we were, this should be a simple `get`; it will be much faster.

will probably require a method added to the DiscoveryIO interface.

dlobue:
@tilogaat  I'm not so sure about using ES this way. What's the pro argument?

tilogaat:
Closing in favor of #428 and #432

PR #414, opened 2015-01-12
Move cobertura to the top-level pom and configure it to generate aggregate results.

dreid:
This makes

```console
$ mvn cobertura:cobertura -P all-modules
```

generate an aggregate coverage report at `target/site/cobertura/index.html`.


gdusbabek:
+1

PR #415, opened 2015-01-12
It's unsafe to mutate defaultResponse concurrently.

dreid:
I've replaced the unsafe mutation with a new instance for every call to respond.

There are no existing unit tests for any of this HTTP framework stuff, but there is an existing integration test that covers the changed code path.

Of course it doesn't actually test anything in a way that would help with a thread run order bug such as this.  I suspect findbugs would correctly identify it but I haven't tried configuring findbugs yet.

There are only 4 usages of this method though, 
 * DefaultHandler
 * NoRouteHandler
 * QueryStringDecoderAndRouter
 * UnsupportedVerbsHandler

So the impact of these changes is probably pretty minimal.


gdusbabek:
+1 good catch.

PR #416, opened 2015-01-13
Removal of overly complicated AsyncChain code, in order to facilitate future maintenance

GeorgeJahad:
1. Removed AsyncChain class and associated tests.
2. Replaced AsyncFunctionWithThreadPool class with FunctionWithThreadPool.
3. Converted AsyncFunctionWithThreadPool's subclasses to use FunctionWithThreadPool instead
4. Converted the users of AsyncChain to invoke each step of the chain individually.
5. Added/fixed unit tests for the individual functions in the chain, (which didn't exist.)
6. Added integration test for preaggregated metrics.
7. Increased the http request timeout from 5 seconds to 1 minute, and changed the status
   returned on timeout from 202 to 504, (GATEWAY_TIMEOUT).
8. My understanding is the Udp ingestor code is basically deprecated.  Since it uses
   AsyncChain as well, I had to update it, just to keep it compiling.  It didn't have
   any tests, and I didn't add any.


blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionServer.java
```diff
@@ -219,4 +127,74 @@ public void exceptionCaught(ChannelHandlerContext ctx, ExceptionEvent e) throws
             return pipeline;
         }
     }
+    static class Processor {
```
chinmay-gupte: What is the reasoning behind declaring this class `static`?

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionServer.java
```diff
@@ -62,51 +60,32 @@
 
 public class HttpMetricsIngestionServer {
     private static final Logger log = LoggerFactory.getLogger(HttpMetricsIngestionServer.class);
-    private static TimeValue DEFAULT_TIMEOUT = new TimeValue(5, TimeUnit.SECONDS);
-    private static int WRITE_THREADS = Configuration.getInstance().getIntegerProperty(CoreConfig.METRICS_BATCH_WRITER_THREADS); // metrics will be batched into this many partitions.
-
+    private static TimeValue DEFAULT_TIMEOUT = new TimeValue(1, TimeUnit.MINUTES);
```
chinmay-gupte: This might have a side effect of making ingestion slower. Is this change necessary because, now some operations in the processor chain happen synchronously and they need this limit raised to finish?

chinmay-gupte:
@GeorgeJahad Is it a proper analysis of the code that the async nature of the processor chain has been reduced to just the `BatchWriter`? I still see that we are going to wait for the `ListenableFuture` returned by the BatchWriter.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionServer.java
GeorgeJahad: It made sense to me as a subclass in the same namespace, and it didn't need a reference to an instance of the parent class.  But I have no problem moving it out, if you think it isn't clear.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionServer.java
GeorgeJahad: No, it's just a part of the 202 removal.  I switched the timeoutException handler to now return a 504 instead of a 202; but I thought we agreed to give it more time before returning an error.  How long should we give it?

GeorgeJahad:
> Is it a proper analysis of the code that the async nature of the
>  processor chain has been reduced to just the BatchWriter?

In reality, the asyncNature of the processing stream hasn't changed at all.

The TypeAndUnitProcessor, RollupTypeCacher, and DiscoveryWriter processing always ran in a threadpool and still do.  They were always asynchronous WRT the batchWriter and they still are.  In the past, to satisfy each step of the chain, the prior step would return a copy of the input in a NoOp future so that the next step could proceed immediately.  Now instead of manipulating the input through the NoOp futures, I pass the input in to each step directly.  Does that make sense?

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionServer.java
GeorgeJahad: I guess should have asked, how would you have implemented it?  Is there a better way to do it?

chinmay-gupte: Seems reasonable. +1.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionServer.java
chinmay-gupte: @GeorgeJahad I think this is out of the scope of this PR. Returning 504 is going to change our API and we should discuss how we want to do it before making the change. 
As for the separate issue of changing the ingestion timeout, lets make it configurable as a first step and set it to current default value of 5 secs, if someone wants to change it, they will have to test it in their staging environment first to find the optimal setting.

GeorgeJahad:
Ran some perf tests to confirm that ingest rate/cpu usage remains the same before and after these changes:

Here are the before graphs:
cpu-usage-before
ingest-before


Here are the after graphs:
cpu-usage-after
ingest-after

chinmay-gupte:
@GeorgeJahad Now that we have removed `AsyncChain`, is there a reason to keep `NoOpFuture` around? I see it getting used only in tests and udp modules...

blueflood-core/src/test/java/com/rackspacecloud/blueflood/inputs/processors/BatchWriterTest.java
```diff
@@ -0,0 +1,115 @@
...
+                    writer,
+                    timeout,
+                    bufferedMetrics,
+                    context);
+        
+        ListenableFuture<List<Boolean>> futures = batchWriter.apply(allTestdata);
+        List<Boolean> persisteds = futures.get(timeout.getValue(), timeout.getUnit());
+
+        //Confirm correct number of futures, (the 2 is because we have both
+        //  simple and PreaggregatedMetric's)
```
chinmay-gupte: NIT: Indentation looks a bit weird over here, dunno what you can with that...

GeorgeJahad:
Fixed indentation and removed no longer used NoOpFuture class.  Thanks Chinmay!

chinmay-gupte:
+1. Please squash the commits before merging.

PR #418, opened 2015-01-15
Use bounded queue for metadata batch threadpool

dlobue:
When `unboundedQueue` became an alias for `SynchronousQueue`, it
improved performance in asyncChain, and almost everywhere in blueflood.
Except for the metadata batch threadpool. This is one of those areas
where a large queue is required in order for work to build up. The more
work that builds up, the bigger the batches, the more efficient the meta
cache batch operations get.

This should speed up blueflood ramp up during cache loads.


chinmay-gupte:
```
When unboundedQueue became an alias for SynchronousQueue
```
@dlobue `unboundedQueue` should never have become an alias for `SynchronousQueue`, it is a huge renaming overlook. We should refactor the method to explicitly have a method called `syncQueue` or something like that.

dlobue:
@chinmay-gupte I don't disagree. That is out of the scope of this PR though.

dreid:
+1

I don't know why coveralls failed to figure out the coverage change but I verified that it does know the lines you added were covered so *shrug*.

PR #420, opened 2015-01-21
Use JaCoCo to get combined unit and integration test coverage.

dreid:
This tries to use jacoco to enable combined coverage reports of integration and unit tests across all modules.

It doesn't address cross-module coverage, so code that is uncovered in blueflood-core, but is executed when blueflood-http tests are run, is counted as uncovered.

This is strictly speaking correct as the coverage provided is accidental and illusory. However there is an issue to provide that functionality in https://github.com/jacoco/jacoco/pull/97

Since we're nowhere near enforcing 100% coverage it's probably fine to just wait until that is addressed.

### DIFF
--- a/blueflood-core/pom.xml
+++ b/blueflood-core/pom.xml
@@ -101,6 +101,7 @@
               <!-- do not require zookeeper for these tests. -->
               <argLine>-DZOOKEEPER_CLUSTER=NONE</argLine>
               <argLine>-Dlog4j.configuration=file://${basedir}/src/test/resources/log4j-test.properties</argLine>
+              <argLine>${jacoco.agent.it.argLine}</argLine>
               <!-- Skips integration tests if the value of skip.integration.tests property is true -->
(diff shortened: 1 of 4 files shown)


blueflood-core/pom.xml
```diff
@@ -101,6 +101,7 @@
               <!-- do not require zookeeper for these tests. -->
               <argLine>-DZOOKEEPER_CLUSTER=NONE</argLine>
               <argLine>-Dlog4j.configuration=file://${basedir}/src/test/resources/log4j-test.properties</argLine>
+              <argLine>${jacoco.agent.it.argLine}</argLine>
```
chinmay-gupte: Where does this argument come from?

dreid: The parent POM https://github.com/rackerlabs/blueflood/pull/420/files#diff-600376dffeb79835ede4a0b285078036R166

dlobue:
+1

PR #421, opened 2015-01-21
Consolidating all DataTypes of all numeric metrics into a single NUMERIC datatype

tilogaat:
Another smaller PR that is a part of the larger whole PR #412 

This PR gets rid of usages of the following DataTypes in bf-core:
- DataType.INT
- DataType.LONG
- DataType.Double
-DataType.Bigint

All these are replaced by a single data type - DataType.Numeric

The DataType value is used in blueflood-core, only for serializing/de-serializing String and Boolean metrics and does not serve any purpose for numeric metrics. However, for UDPSerialization it is required to deserialize the UDP datagram. 

This is the first step towards stopping caching of DataType information in the metadata cache
This PR is also a response to the review comments made here https://github.com/rackerlabs/blueflood/pull/412#discussion_r23191294.

### DIFF
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/types/DataType.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/types/DataType.java
@@ -32,20 +32,21 @@ public class DataType {
     public final static DataType DOUBLE = new DataType("D");
     public final static DataType BOOLEAN = new DataType("B");
     public final static DataType BIGINT = new DataType("BI");
+    public final static DataType NUMERIC = new DataType("N");
 
     public static DataType getMetricType(Object metricValue) {
         if (metricValue instanceof String) {
             return STRING;
         } else if (metricValue instanceof Integer) {
-            return INT;
+            return NUMERIC;
         } else if (metricValue instanceof Long) {
-            return LONG;
+            return NUMERIC;
         } else if (metricValue instanceof Double) {
-            return DOUBLE;
+            return NUMERIC;
         } else if (metricValue instanceof Boolean) {
             return BOOLEAN;
         } else if (metricValue instanceof BigInteger) {
-            return BIGINT;
+            return NUMERIC;
         } else {
             throw new RuntimeException("Unknown metric value type " + metricValue.getClass());
         }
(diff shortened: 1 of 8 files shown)


blueflood-core/src/main/java/com/rackspacecloud/blueflood/types/DataType.java
```diff
@@ -32,20 +32,21 @@ public DataType(String type) {
     public final static DataType DOUBLE = new DataType("D");
     public final static DataType BOOLEAN = new DataType("B");
     public final static DataType BIGINT = new DataType("BI");
+    public final static DataType NUMERIC = new DataType("N");
 
     public static DataType getMetricType(Object metricValue) {
         if (metricValue instanceof String) {
             return STRING;
         } else if (metricValue instanceof Integer) {
```
chinmay-gupte: Can we remove all these cases and just have `metricValue instanceOf Number`?

blueflood-udp/src/main/java/com/rackspacecloud/blueflood/service/udp/UDPMetricSerialization.java
```diff
@@ -105,26 +107,44 @@ private static int computeBinarySize(Collection<Metric> metrics) {
...
+        else if (dataType.equals(DataType.LONG))
+            out.writeRawVarint64((Long) metric.getMetricValue());
+        else if (dataType.equals(DataType.DOUBLE))
             out.writeDoubleNoTag((Double) metric.getMetricValue());
         else if (metric.getDataType().equals(DataType.BOOLEAN))
             out.writeBoolNoTag((Boolean) metric.getMetricValue());
     }
-    
+
+    private static DataType getDataType(Metric metric) {
```
chinmay-gupte: We can remove this block of code as well, I think. and then use `instanceOf` operator to get the type of the object in the `write` block.

tilogaat: This method is used in write and computeBinarySize which is why I extracted it into a method. Also, the datatype itself is plugged into the output stream.

tilogaat:
Closing in favor of #428

PR #425, opened 2015-01-23
Instrument mark slot dirty

dlobue:
Instrument the amount of time spent marking slots dirty as suggested by @dreid on irc yesterday.

This instruments the amount of time spent marking all the slots in a single batch as dirty, as well as the amount of time spent marking a single slot as dirty.

It also adds a metric to record how long it took to persist a single batch. This last metric will be useful for sense of scale since the current `BatchWriter` duration metric is for the all the batches summed.


dreid:
+1

As an aside it might be nice to start using metrics-annotations for whole method metrics.

PR #426, opened 2015-01-23
Add toggle for token aware load balancing to Cassandra

dlobue:
Add toggle for token aware load balancing strategy to Cassandra in
Astyanax. This will allow us to bypass coordinator nodes and talk
directly to the node that owns the data.


chinmay-gupte:
I would like to test this in staging first to see the results of this change. +1 to staging.

chinmay-gupte:
@dlobue Closing this due to lack of activity. Please re-open if need be.

PR #427, opened 2015-01-23
Get rid of unnecessary coordination in Instrumentation.

dreid:
Ok, these ``synchronized`` blocks are in place to prevent two concurrent requests making 2 separate instances of ``Meter`` or ``Timer`` in the associated ``HashMap``.

However… ``com.rackspacecloud.blueflood.utils.Metrics$timer`` (and ``$meter``) uses a ``com.codahale.metrics.MetricRegistry`` instance which is backed by a ``ConcurrentHashMap``.

So this change just deletes all of ``Instrumentation``'s ``HashMap``s and the ``synchronized`` blocks and uses ``com.rackspace.blueflood.utils.Metrics`` directly.


dlobue:
+1
and as a bonus, we reduce our memory footprint. yay!

PR #433, opened 2015-01-30
Enable checkstyle

dreid:
This enables the standard maven checkstyle plugin.

http://maven.apache.org/plugins/maven-checkstyle-plugin/


chinmay-gupte:
+1

PR #440, opened 2015-02-19
RFR: Changing the precision of sum field in timer to double

tilogaat:
The statsd payload being sent by ele https://gist.github.com/tilogaat/4d44019dd93d53fac6ce 
is erroring out because it has a timer whose sum is of greater precision than  long. 
Assuming long precision for timer.sum is not enough. Therefore, changing the precision to be double instead.


blueflood-core/src/test/resources/serializations/timer_version_0.bin
```diff
@@ -1 +1 @@
-ACrZAlK4HoXrUTdAAABsOANs2wcCbAIBbmq8dJOYqopAAgNmb29uw/UoXI8qh0ADYmFybmFUUiegiaA/
+AAAAAAAAAEVA2QJSuB6F61E3QAAAbDgDbNsHAmwCAW5qvHSTmKqKQAIDZm9vbsP1KFyPKodAA2Jhcm5hVFInoImgPw==
```
GeorgeJahad: sorry, i'm not sure what this is?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/serializers/NumericSerializer.java
```diff
@@ -332,14 +333,38 @@ private static SetRollup deserializeV1SetRollup(CodedInputStream in) throws IOEx
...
-    
-    private static void serializeTimer(TimerRollup rollup, byte[] buf) throws IOException {
+
+    private static void serializeV1Timer(TimerRollup rollup, byte[] buf) throws IOException {
         CodedOutputStream out = CodedOutputStream.newInstance(buf);
         timerRollupSize.update(buf.length);
         out.writeRawByte(Constants.VERSION_1_TIMER);
+
+        // sum, count, countps, avg, max, min, var
+        out.writeRawVarint64((long)rollup.getSum());
```
GeorgeJahad: Is there a reason why you are changing the old code?  (the cast to long?)

tilogaat: rollup.getSum() has been converted into a double. So to serialize it in the old way we need to cast it to long

chinmay-gupte: @GeorgeJahad The old code is not getting changed, per se. GH diff is confusing, but the method name for `serializeTimer` has been changed to `serializeV1Timer` and the same code has been added to another method called `serializeV2Timer`, with just a change of writing `sum` as double instead of long. I wonder if we can refactor it better, instead of duplicating most of the code in another method.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/serializers/NumericSerializer.java
```diff
@@ -613,21 +795,37 @@ public ByteBuffer toByteBuffer(TimerRollup o) {
...
+        }
+
+
         @Override
         public TimerRollup fromByteBuffer(ByteBuffer byteBuffer) {
             CodedInputStream in = CodedInputStream.newInstance(byteBuffer.array());
             try {
                 byte version = in.readRawByte();
-                if (version != VERSION_1_TIMER)
+                if (version > VERSION_2_TIMER)
```
chinmay-gupte: Can you make the checking for versions more explicit like,
```
if (version == V1) {...}
else if (version == V2) {...}
else SerializationException()
```

chinmay-gupte:
LGTM other than a few refactoring comments. Especially, the part where we are duplicating a lot of code, by putting them in different method, where only a single line needs to be adapted https://github.com/rackerlabs/blueflood/pull/440#discussion_r25188173.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/serializers/NumericSerializer.java
tilogaat: it seems strange to refactor the methods and just pass a long sum into it. I'd rather pass all fields into it. Also if you have common methods shared between 2 methods of serialization, then if it changes for 1 version, we have to make sure that it is compatible with the other version too.

GeorgeJahad:
Tilo asked me to finish this one while she was out of town.   It is continued here:
https://github.com/rackerlabs/blueflood/pull/442

PR #442, opened 2015-02-25
Timer sum double

GeorgeJahad:
This is the continuation of:
https://github.com/rackerlabs/blueflood/pull/440

Chinmay asked for some changes, and Tilo asked me to do them while she is at Rax.io for the week.

The only part of this PR that still needs to be reviewed is this:
https://github.com/rackerlabs/blueflood/commit/24ecd6ed98ca559830c8937c6a7da48d050ae681


blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/serializers/NumericSerializer.java
```diff
@@ -535,7 +556,7 @@ private static AbstractRollupStat getStatFromRollup(byte statType, BasicRollup b
         public ByteBuffer toByteBuffer(Object o) {
             try {
                 byte type = typeOf(o);
-                byte[] buf = new byte[sizeOf(o, type)];
+                byte[] buf = new byte[sizeOf(o, type, VERSION_2_TIMER)];
```
chinmay-gupte: It seems a bit strange to pass in the `timer` version while serializing non-timer rollups like gauges, counters etc. What do you think about having two overloaded methods named `sizeOf` with one accepting just two arguments i.e `(Object, Type)` which will be used by all non-timer rollup types and then the other one with `timerVersion` as the third argument, which gets called by the timer serializer?

GeorgeJahad: This is much cleaner.  I originally thought there was a problem doing it this way, but it seems much better.  I'll push some changes shortly.

chinmay-gupte:
+1

PR #444, opened 2015-03-06
George/multiplot simple number bug

GeorgeJahad:
1. Fixed AstyanaxReader.getPointsFromColumns() to use pointFromColumn()
   	 instead of generating points with "new".  This fixes
	 multiplot query because it causes it to return SimpleNumber's
	 instead of Integers.
2. Removed extraneous gran param from pointFromColumn()
3. Improved error messages in JSONBasicRollupsOutputSerializer.java
4. Increased Multiplot test coverage in BatchMetricsQueryHandlerIntegrationTest
     to test Full resolution metrics and confirm that it is returning
     SimpleNumbers instead of integers.
5. Create metricLocator and strLocator vars in BatchMetricsQueryHandlerIntegrationTest
     to make the code a bit more readable.


GeorgeJahad:
Just to be clear, the bug that occurs is that Multiplot queries use AstyanaxReader.getPointsFromColumns() which was returning integers for full res metrics, (instead of SimpleNumbers, which is what SinglePlot full res returns.

That causes this method to break because it can't handle Integers:
com.rackspacecloud.blueflood.outputs.serializers.JSONBasicRollupsOutputSerializer.toJSON()

chinmay-gupte:
+1. Nice catch!

PR #451, opened 2015-04-14
WIP: Increasing the Metrics List search result limit

tilogaat:
https://github.com/rackerlabs/blueflood/commit/dfec1a9c6e2c83598680379cd0d7c70d42ed07ed had previously increased it to 100k from 500. 

Therefore, the finder does not return statsd metrics and shows only monitoring metrics in grafana. This PR doubles the result set size.


tilogaat:
This has been a problem just once. Closing PR.

PR #452, opened 2015-04-16
WIP: Fix samplecount bug

tilogaat:
in PreaggregateConversions (https://github.com/rackerlabs/blueflood/compare/master...tilogaat:fix-samplecount-bug?expand=1#diff-b643ac2bbdc21b81d8434e67b8c56026L73) withCount was being invoked by passing the sampleCount and the Count was always being set to the sampleCount value in the counterRollup class. A side effect of this is that sampleCount should have been long, but was mistakenly set to int.


tilogaat:
Closing it in favor of #453

PR #453, opened 2015-04-16
Fixing sampleCount bug 

tilogaat:
convertCounters is wrongly invoking withCount for setting sampleCount. As a result, all count values are getting overridden to the sampleCount values which is incorrect. 

#452 has additional commits that probably should be made because sampleCount field in CounterRollup class is an int, but that is probably insufficient precision that has gone unnoticed because of this bug.


GeorgeJahad:
+1

PR #455, opened 2015-04-16
WIP: Ignoring ShardStateIntegrationTest 

tilogaat:
Ignoring ShardStateIntegrationTest for now because it's causing build failures. 

This is the failure that's being caused:
http://maven.apache.org/surefire/maven-surefire-plugin/faq.html#vm-termination


GeorgeJahad:
+1

PR #456, opened 2015-04-20
Adding counterValue field to counter query result

tilogaat:
To post a counter to BF via statsd backend or directly via the http endpoint the payload is:
```
curl -i -XPOST 'http://localhost:19000/v2.0/5405533/ingest/aggregated' -d '
  {
  "tenantId": "5405533",
  "timestamp": 1429567619000,
  "counters": [
    {
      "name": "tilo_counter",
      "rate": 1,
      "value": 32
    },
    {
      "name": "tilo_another_counter",
      "rate": 1,
      "value": 4424
    }
  ]
}'
```
But while querying the data back the field that is returned is numPoints as shown below:

```
curl -i -X GET "http://localhost:20000/v2.0/5405533/views/tilo_counter?from=1429564800000&to=1429568400000&points=20"
HTTP/1.1 200 OK
Content-Length: 208

{
  "unit": "unknown",
  "values": [
    {
      "numPoints": 32,
      "timestamp": 1429567500000
    }
  ],
  "metadata": {
    "limit": null,
    "next_href": null,
    "count": 1,
    "marker": null
  }
}
```

This is confusing to say the least and is not quite correct because CounterRollup.getCount does not really represent the number of points. Fixing that by replacing the numPoints field with counterValue field which is more intuitive and is more correct too. Query result after this PR is:

```
url -i -X GET "http://localhost:20000/v2.0/5405533/views/tilo_counter?from=1429564800000&to=1429568400000&points=20"
HTTP/1.1 200 OK
Content-Length: 234

{
  "unit": "unknown",
  "values": [
    {
      "numPoints": 32,
      "timestamp": 1429567500000,
      "counterValue": 32
    }
  ],
  "metadata": {
    "limit": null,
    "next_href": null,
    "count": 1,
    "marker": null
  }
}
```

### DIFF
--- a/blueflood-http/src/main/java/com/rackspacecloud/blueflood/outputs/serializers/BasicRollupsOutputSerializer.java
+++ b/blueflood-http/src/main/java/com/rackspacecloud/blueflood/outputs/serializers/BasicRollupsOutputSerializer.java
@@ -156,6 +156,21 @@ public interface BasicRollupsOutputSerializer<T> {
                 return rawSample;
             }
         },
+        COUNTER_VALUE("counterValue") {
+            @Override
+            Object convertRollupToObject(Rollup rollup) throws Exception {
+                if (rollup instanceof CounterRollup)
+                    return ((CounterRollup) rollup).getCount();
+                else
+                    // gauge, set, basic
+                    throw new Exception(String.format("counter value not supported for this type: %s", rollup.getClass().getSimpleName()));
+            }
+
+            @Override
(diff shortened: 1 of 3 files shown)


GeorgeJahad:
Isn't numPoints supposed to equal sampleCount?  Should we fix that as well?

GeorgeJahad:
Tilo and I agreed we would fix numPoints and TimerRollups in a separate PR, because we want to get this out now

GeorgeJahad:
+1

goru97:
+1

PR #460, opened 2015-04-30
Multiplot/fetch_multi changes

GeorgeJahad:
Made a number of fixes/enhancements to the finder:

1. Graphite requires the datapoints returned by the finder to be 
equidistant in time, (interleaved with Null datapoints if necessary.)
We weren't doing this so the timescale was off on graphs with sparse data.

2. Improved performance significantly by switching to fetch_multi/multiplot.
(Note this required updating the graphite-api code from version 1.0.1 to 
Head of master.  Since we also had to patch that code I created a new repo
for it here: https://github.com/rackerlabs/graphite-api/tree/george/fetch_multi_with_patches

3. One other detail about fetch_multi is that it has to take the multiplot limits into
account.  Multplot allows only 100 metrics per call, and has implicit limit of around
7000 json characters. So my changes batch up each of graphite-apis fetch_multi() calls
into to the minimum number of multiplot calls that still abide by those limits.

4. Now supports specificy specific submetric names in the config file, (rather
than just assuming "average", "num_points" or "latest" as we were before.
This required changes on both the search side, and the fetch_multi side.
This work was based on the changes suggested and created by Alex Weeks:
https://github.com/aweeks

5. Finally added a bunch of tests to get coverage of this file up to 93%:
https://23282e1499f9e238db58e26751fc6d96be63cc17-www.googledrive.com/host/0B28IWTWKORCkfnlLYmljbW1NQ295NWhWbmtfbDJ1UXYwQm9xdlVuMjN0NFV1QTNmcDltY2M/x/blueflood.html

### DIFF
--- a/contrib/graphite/blueflood.py
+++ b/contrib/graphite/blueflood.py
@@ -52,10 +52,6 @@ class TenantBluefloodFinder(object):
   __fetch_multi__ = 'tenant_blueflood'
   def __init__(self, config=None):
     print("Blueflood Finder v22")
-#    if os.path.isfile("/root/pdb-flag"):
-#      import remote_pdb
-#      remote_pdb.RemotePdb('127.0.0.1', 4444).set_trace()
-
     if config is not None:
       bf_config = config.get('blueflood', {})
       urls = bf_config.get('urls', bf_config.get('url', '').strip('/'))
--- a/contrib/graphite/blueflood.py
+++ b/contrib/graphite/blueflood.py
@@ -88,8 +88,12 @@ class TenantBluefloodFinder(object):
     print("BF finder submetrics enabled: ", enable_submetrics)
 
   def complete(self, metric, query_depth):
-    #is submetric name complete?
-    return len(metric.split('.')) == (query_depth - 1)
+    metric_parts = metric.split('.')
+    if query_depth < len(metric_parts):
+      return False
+    if self.enable_submetrics and (query_depth == len(metric_parts)):
+      return False
+    return True
 
   def make_request(self, url, payload, headers):
     if auth.is_active():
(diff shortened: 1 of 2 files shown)


contrib/graphite/blueflood.py
```diff
@@ -43,37 +81,105 @@ def __init__(self, config=None):
...
+  def find_metrics(self, query):
+    print "BluefloodClient.find_metrics: " + str(query)
+    payload = {'query': query}
+    headers = auth.headers()
+    endpoint = self.find_metrics_endpoint(self.bf_query_endpoint, self.tenant)
+    r = self.make_request(endpoint, payload, headers)
+
+    if r.status_code == 200:
+        return [m['metric'] for m in r.json()]
+    else:
```
chinmay-gupte: Is it worth adding retries over here?

contrib/graphite/blueflood.py
```diff
@@ -15,25 +18,60 @@
...
+    num_points = (stop - start) / secs_per_res['MIN1440']
+    res = 'MIN1440'
+  return res
+
+class TenantBluefloodFinder(object):
+  __fetch_multi__ = 'tenant_blueflood'
   def __init__(self, config=None):
-    authentication_module = None
+    print("Blueflood Finder v21")
+#    if os.path.isfile("/root/pdb-flag"):
```
chinmay-gupte: What are these commented out lines for?

contrib/graphite/blueflood.py
GeorgeJahad: People have been using it without a problem I think, and I've had to work on so many other issues that I'm not going to do this one yet.

contrib/graphite/blueflood.py
GeorgeJahad: I'll remove them

contrib/graphite/blueflood.py
```diff
@@ -43,37 +77,133 @@ def __init__(self, config=None):
...
     self.bf_query_endpoint = urls[0]
+    self.enable_submetrics = enable_submetrics
+    self.submetric_aliases = submetric_aliases
+    self.client = BluefloodClient(self.bf_query_endpoint, self.tenant, 
+                                  self.enable_submetrics, self.submetric_aliases)
+    print("BF finder submetrics enabled: ", enable_submetrics)
+
+  def complete(self, metric, query_depth):
+    metric_parts = metric.split('.')
+    if query_depth < len(metric_parts):
```
chinmay-gupte: Totally up to you, but what do you think about consolidating these if statements into just one block like, `if (query_depth < len(metric_parts)) or (self.enable_submetrics and (query_depth == len(metric_parts)))`

GeorgeJahad:
These files have been updated per the review from chinmay-gupte.

1. Tried to simplify the find_nodes() method a bit by splitting
it out into 2 separate methods depending on whether submetrics are
used or not.

2. Changed the term "cache" to "dictionary"

3. No longer sorts timestamps, because they are sorted from BF.

4. Chinmay suggested that we rollup metrics that coexist in the same
"step", (instead of dropping all but the first.)  However, this
problem only occurs when full res metrics have sub-second frequencies.
Since that seems to be a rare case, I've just documented the problem
for now, and will fix it when it actually becomes a problem


Note that if you only want to see the diffs since the initial
version of this PR, you can see them like so:
git diff 6e7e811 c21dd40

contrib/graphite/blueflood.py
```diff
@@ -43,37 +77,133 @@ def __init__(self, config=None):
...
+    query_parts = query.pattern.split('.')
+    query_depth = len(query_parts)
+
+    if query_depth > 1:
+      #  In this if clause we are searching for complete metric names
+      #  followed by a ".*". If so, we are requesting a list of submetric
+      #  names, so return leaf nodes for each submetric alias that is
+      #  enabled
+      if query_parts[-1] == '*':
+        for metric in self.find_metrics('.'.join(query_parts[0:-1])):
```
chinmay-gupte: If the query end with a `*` but its not for a leaf node, it will make a call to `find_metrics` here and because its not complete, it will make a call again to `find_metrics` to return a branch node. Are these actual API calls or the data returned is cached?

chinmay-gupte: If its doing the former, can we make just one API call in this function to `find_metrics` and cache the data in a local variable?

contrib/graphite/blueflood.py
```diff
@@ -43,37 +77,133 @@ def __init__(self, config=None):
...
+          metric_parts = metric.split('.')
+          if self.complete(metric, query_depth):
+            for alias, _ in self.submetric_aliases.items():
+              yield TenantBluefloodLeafNode('.'.join(metric_parts + [alias]),
+                                            TenantBluefloodReader(metric, self.tenant, 
+                                                                  self.bf_query_endpoint, 
+                                                                  self.enable_submetrics, 
+                                                                  self.submetric_aliases))
+
+      #if searching for a particular submetric alias, create a leaf node for it
```
chinmay-gupte: The more I think about it, suppose someone makes a call to `a.b.*` and the complete metric is `a.b.c.d`, it will make a call first to get the metrics, then check if it complete in order to append sub-metric aliases to it, and if its not complete, then again make a call for the branch nodes. It seems like a toggled approach to me. What do you think about making these checks for sub-metric aliases only if the metrics are complete? Let me know if it makes sense.

contrib/graphite/blueflood.py
```diff
@@ -43,37 +77,133 @@ def __init__(self, config=None):
...
+    endpoint = self.find_metrics_endpoint(self.bf_query_endpoint, self.tenant)
+    r = self.make_request(endpoint, payload, headers)
+
+    if r.status_code == 200:
+        return [m['metric'] for m in r.json()]
+    else:
+      return []
+
+  def find_nodes_with_submetrics(self, query):
+    # By definition, when using submetrics, the names of all Leafnodes must end in a submetric alias
```
chinmay-gupte: Nice comment!

contrib/graphite/blueflood.py
```diff
@@ -87,105 +217,207 @@ def fetch(self, start_time, end_time):
...
+      'to': end_time * 1000,
       'resolution': res
     }
-    #print 'USING RES ' + res
+    if self.enable_submetrics:
+      payload['select'] = ','.join(self.submetric_aliases.values())
+    return payload
+
+  def gen_keys(self, node, metrics):
+    # returns metrics_key and data_key
```
chinmay-gupte: NIT: formatting seems to be a bit off in some of the comments.

contrib/graphite/blueflood.py
```diff
@@ -87,105 +217,207 @@ def fetch(self, start_time, end_time):
...
+        ret_arr.append(v_iter[0][data_key])
+      else:
+        ret_arr.append(None)
+
+    return ret_arr
+
+  def get_multi_endpoint(self, endpoint, tenant):
+    return "%s/v2.0/%s/views" % (endpoint, tenant)
+
+  def get_metric_list(self, endpoint, tenant, metric_list, payload, headers):
```
chinmay-gupte: This name of the function might be a bit confusing, as it prompts to metrics_list api call. Can we have a better name of it, may be `get_metric_data` or something like that?

contrib/graphite/blueflood.py
GeorgeJahad: Sorry, I don't see the problem.  Can you be more explicit?

contrib/graphite/blueflood.py
GeorgeJahad: i'll fix

GeorgeJahad:
The lastest diffs:
https://github.com/rackerlabs/blueflood/compare/c21dd403edc118ffe614324f143c433185acd5d3...george/add_fetch_multi2

contrib/graphite/blueflood.py
```diff
@@ -43,37 +78,149 @@ def __init__(self, config=None):
...
       
-      for obj in values:
-        metric = obj['metric']
-        parts = metric.split('.')
-        metric_depth = len(parts)
-        if metric_depth > query_depth:
-          yield BranchNode('.'.join(parts[:query_depth]))
+      # First modify the pattern to get a superset that includes already complete
+      #  submetrics
+      new_pattern = complete_pattern + '*'
```
chinmay-gupte: Is the variable `new_pattern` same as original `query.pattern`?

chinmay-gupte:
+1 after my last comment has been addressed.

contrib/graphite/blueflood.py
GeorgeJahad: no it isn't.  It is a superset which is what I was trying to explain in the comment.  For example, if query.pattern is "a.b.*", it won't return "a.b",  so new_pattern is "a.b*", with no second ".".  Does that make sense?  I'm trying to eliminate the second api call by combining the two.

PR #461, opened 2015-05-04
Adding support for average, min, max, variance to Gauge rollups

tilogaat:
GaugeRollups are inherited from BasicRollup. They can provide the functionality of min, max, average and variance out of the box. Those values just needed to be returned in BasicRollupsOutputSerializer.


tilogaat:
Closing for now, since the issue cannot be replicated

PR #463, opened 2015-05-07
Remove DropWizard for now

stackedsax:
We haven't touched the DropWizard code in ages.  We haven't implemented it.  It's getting in the way of builds, and probably confusing people who are looking at our codebase.

Let's just remove this for the time being.  We have source control -- if we need to deliver this at some future point we'll resurrect it.

(sorry, @gdusbabek!)


gdusbabek:
Ponies

chinmay-gupte:
heh. sorry @gdusbabek but one of the dependency is really breaking our builds.
@stackedsax +1

PR #468, opened 2015-05-26
RFR: Fixing the points to Granularity mapping such that it takes into account the ttls 

tilogaat:
Currently, when a query arrives for points, BF tries to internally map it to resolution and invokes GetByResolution. But sometimes the interval is for older data and BF fails to take into account that the data may have been TTLd out. 

For example, if the query is for 1000 points that have occurred in the last month, Full Granularity should not be the best match because the data for Full resolution has TTLd out and will not be available. 
However currently BF just returns Full Granularity because that's the way we can return the max points (close to 1000), but when BF checks into the metrics_full CF, there will not be anything in it because data has TTLd. 

This PR tries to make the points to granularity mapping smarter by taking into account the ttls. 

Main code changed is here in [Granularity.java](https://github.com/rackerlabs/blueflood/compare/master...tilogaat:fix-points2granularity?expand=1#diff-f6ab010e2a8c7c85dd54384bd3a64b37R261)

Todo: 
1. Tests, tests, tests
2. Code cleanup


blueflood-core/src/main/java/com/rackspacecloud/blueflood/rollup/Granularity.java
```diff
@@ -274,9 +288,29 @@ private static Granularity granularityFromPointsGeometric(double requestedDurati
...
+            gran = Granularity.LAST;
+        }
+
         return gran;
     }
 
+    private static long getTTL(String tenantid, Granularity g) {
+        long ttl;
+        try {
+            ttl = (long)TTL_PROVIDER.getTTL(
```
GeorgeJahad: @chinmay-gupte points out that this is not correct for full res metrics.  See here:
https://github.com/rackerlabs/blueflood/blob/master/blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionHandler.java#L178

GeorgeJahad: Also this is a useful utility method than @chinmay-gupte is going to want for rollupsOnRead().  So it probably go somewhere else, maybe one of the ttl classes.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/rollup/Granularity.java
```diff
@@ -248,11 +253,20 @@ private static Granularity granularityFromPointsLinear(double requestedDuration,
...
         double minimumPositivePointRatio = Double.MAX_VALUE;
         Granularity gran = null;
+        if (TTL_PROVIDER == null) {
+            TTL_PROVIDER = SafetyTtlProvider.getInstance();
+        }
 
         for (Granularity g : Granularity.granularities()) {
+            long ttl = getTTL(tenantid, g);
+
+            if (to < Calendar.getInstance().getTimeInMillis() - ttl) {
```
GeorgeJahad: Should this be "from <" instead of "to <"??

blueflood-core/src/main/java/com/rackspacecloud/blueflood/rollup/Granularity.java
GeorgeJahad: Is the (long) cast needed here?  Doesn't toMillis() return a long?

blueflood-core/src/test/java/com/rackspacecloud/blueflood/rollup/GranularityTest.java
```diff
@@ -22,31 +22,91 @@
...
 public class GranularityTest {
 
-    final long baseMillis = 1335820166000L;
+    //final long baseMillis = 1335820166000L;
+    final long fromBaseMillis = Calendar.getInstance().getTimeInMillis();
+    final long toBaseMillis = fromBaseMillis+ 604800000;
+
+    //An old timestamp that signifies ttld out data.
+    // 619200000 is 8 days, just 1 day more than 7 days ie ttl limit of full gran
+    final long oldToBaseMillis_FullGran = fromBaseMillis - 619200000;
```
GeorgeJahad: Minor Nit:  I don't love putting "_" in variable names unless they are all caps.  Is there a reason why you are doing that?  It looks a bit odd to me.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/rollup/GranularityTest.java
```diff
@@ -22,31 +22,91 @@
...
-        Assert.assertEquals(Granularity.MIN_1440.name(), Granularity.granularityFromPointsInInterval(0, 86400000, 1).name());
-
-        Assert.assertEquals(Granularity.FULL.name(), Granularity.granularityFromPointsInInterval(0, 43200000, 800).name());
-        Assert.assertEquals(Granularity.MIN_5.name(), Granularity.granularityFromPointsInInterval(0, 43200000, 288).name()); // 144 5m points vs 1440 full points.
-        Assert.assertEquals(Granularity.MIN_5.name(), Granularity.granularityFromPointsInInterval(0, 43200000, 144).name());
-        Assert.assertEquals(Granularity.MIN_20.name(), Granularity.granularityFromPointsInInterval(0, 43200000, 35).name());
-        Assert.assertEquals(Granularity.MIN_60.name(), Granularity.granularityFromPointsInInterval(0, 43200000, 11).name());
-        Assert.assertEquals(Granularity.MIN_240.name(), Granularity.granularityFromPointsInInterval(0, 43200000, 3).name());
-
-        Assert.assertEquals(Granularity.MIN_5.name(), Granularity.granularityFromPointsInInterval(baseMillis, baseMillis + 86400000, 287).name());
```
GeorgeJahad: Is there a reason why you dropped this assert?

blueflood-core/src/test/java/com/rackspacecloud/blueflood/rollup/GranularityTest.java
```diff
@@ -22,31 +22,91 @@
...
+    public void testFromPointsInInterval_LessThanADayInterval() throws Exception {
```
GeorgeJahad: Is there a reason why you split the old testFromPointsInInterval() method into two separate methods?

blueflood-core/src/test/java/com/rackspacecloud/blueflood/rollup/GranularityTest.java
```diff
@@ -22,31 +22,91 @@
...
 
+import java.util.Calendar;
 import java.util.HashMap;
-import java.util.Iterator;
 import java.util.Map;
 
 public class GranularityTest {
 
-    final long baseMillis = 1335820166000L;
+    //final long baseMillis = 1335820166000L;
```
GeorgeJahad: Is there a reason why you commented this out instead of deleting?

blueflood-core/src/test/java/com/rackspacecloud/blueflood/rollup/GranularityTest.java
```diff
@@ -22,31 +22,91 @@
...
+    final long fromBaseMillis = Calendar.getInstance().getTimeInMillis();
+    final long toBaseMillis = fromBaseMillis+ 604800000;
+
+    //An old timestamp that signifies ttld out data.
+    // 619200000 is 8 days, just 1 day more than 7 days ie ttl limit of full gran
```
GeorgeJahad: You can make the code clearer and more self documenting by using SECS_PER_DAY, (from Granularity.java),
    and using multiples of it like (8 * SECS_PER_DAY), instead of magic numbers like 619200000
    // 619200000 is 8 days, just 1 day more than 7 days ie ttl limit of full gran

blueflood-core/src/main/java/com/rackspacecloud/blueflood/rollup/Granularity.java
tilogaat: @GeorgeJahad It's possible for from to be before the TTL period has passed, but as long as to falls within, there should be some points in that interval. Do you agree?

blueflood-core/src/test/java/com/rackspacecloud/blueflood/rollup/GranularityTest.java
tilogaat: I put the _ in to make it less of a mouthful (eyeful?), I thought oldToBaseMillisFullGran is a little harder to read.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/rollup/GranularityTest.java
tilogaat: There is a whole test for it .. AWeekInterval

blueflood-core/src/test/java/com/rackspacecloud/blueflood/rollup/GranularityTest.java
tilogaat: Yes that test was getting too long and it was hard to debug where exactly the test was failing.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/rollup/GranularityTest.java
tilogaat: No I should delete it.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/rollup/GranularityTest.java
tilogaat: Agreed. Will fix.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/rollup/Granularity.java
tilogaat: I am confused about which methods we should be using [getTTL](https://github.com/rackerlabs/blueflood/blob/master/blueflood-core/src/main/java/com/rackspacecloud/blueflood/cache/ConfigTtlProvider.java#L125) or [getTTLForIngestion](https://github.com/rackerlabs/blueflood/blob/master/blueflood-core/src/main/java/com/rackspacecloud/blueflood/cache/ConfigTtlProvider.java#L148) and should we consider [areTTLsForced](https://github.com/rackerlabs/blueflood/blob/master/blueflood-core/src/main/java/com/rackspacecloud/blueflood/cache/ConfigTtlProvider.java#L152)

blueflood-core/src/main/java/com/rackspacecloud/blueflood/rollup/Granularity.java
GeorgeJahad: The thing I'm concerned about is there may only be one or two points there and it will look wierd in the finder.
My preference is for the finder to look as good as possible even if it means we use less granular data.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/rollup/Granularity.java
GeorgeJahad: yeh, maybe you should get his opinion.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/rollup/Granularity.java
tilogaat: @GeorgeJahad ok that makes sense. will fix.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/rollup/Granularity.java
tilogaat: fixed in 01dcdc0

blueflood-core/src/main/java/com/rackspacecloud/blueflood/rollup/Granularity.java
tilogaat: fixed in 01dcdc0

blueflood-core/src/test/java/com/rackspacecloud/blueflood/rollup/GranularityTest.java
tilogaat: fixed in 01dcdc0

blueflood-core/src/main/java/com/rackspacecloud/blueflood/cache/ConfigTtlProvider.java
```diff
@@ -140,11 +140,23 @@ public void setTTL(String tenantId, Granularity gran, RollupType rollupType, Tim
         throw new RuntimeException("Not allowed to override TTL values specified in configs.");
     }
 
+    public long getTTLForGranularityAndTenant(String tenantid, Granularity g) throws Exception{
```
GeorgeJahad: Is this method still used anywhere?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/cache/TenantTtlProvider.java
```diff
@@ -26,4 +26,6 @@
     public void setTTL(String tenantId, Granularity gran, RollupType rollupType, TimeValue ttlValue) throws Exception;
 
     public TimeValue getTTLForStrings(String tenantId) throws Exception;
+
+    public TimeValue getConfigTTLForIngestion() throws Exception;
```
GeorgeJahad: Maybe I misunderstood, but I thought @chinmay-gupte said to expose the getFinal TTL method in the interface, instead of this one?

GeorgeJahad:
+1 after removing the extra method

PR #473, opened 2015-06-03
Query by points in Finder instead of by resolution

tilogaat:
We want the ability to query by points in the blueflood-finder for Grafana to take advantage of getByPoints smart querying. getByPoints maps points to resolution, but it proceeds to the next coarser resolution if it finds that the data belonging to one resolution has TTLd out. This is useful for graphing over a long period of time. 

TODO:
Test the finder manually using Grafana.


GeorgeJahad:
+1

tilogaat:
Old finder with getbyRes
New finder with getbyPoints

PR #485, opened 2015-06-23
Add grinder based distributed load testing for Blueflood

GeorgeJahad:
Grinder is a distributed load testing tool described here:
http://grinder.sourceforge.net/g3/getting-started.html

This PR defines implementations of grinder worker threads meant to repeatedly invoke the required number of BF api calls during each "reporting interval"

It also includes the infrastructure to divide the total work described in the properties file across all the workers in the distrubuted system.

These files are probably easiest reviewed in the following order:
grinder.py
utils.py
blueflood.py
query.py
tests.py


contrib/grinder/properties/grinder.properties
```diff
@@ -0,0 +1,20 @@
+grinder.script=../scripts/tests.py
+grinder.package_path=/Library/Python/2.7/site-packages
+grinder.runs=1
+grinder.threads=25
+grinder.useConsole=false
+grinder.logDirectory=resources/logs
+grinder.bf.name_fmt="t4.int.abcdefg.hijklmnop.qrstuvw.xyz.ABCDEFG.HIJKLMNOP.QRSTUVW.XYZ.abcdefg.hijklmnop.qrstuvw.xyz.met.%d"
+grinder.bf.report_interval=10000
```
chinmay-gupte: Is this the flush interval for the metrics generated by the tool?

contrib/grinder/scripts/grinder.py
```diff
@@ -0,0 +1,27 @@
+# A more complex HTTP example based on an authentication conversation
```
chinmay-gupte: Can you explain this comment in plain English? :)

contrib/grinder/properties/grinder.properties.staging
```diff
@@ -0,0 +1,20 @@
+grinder.script=../scripts/grinder.py
```
chinmay-gupte: Is it necessary to supply 2 sample configs?

contrib/grinder/scripts/utils.py
```diff
@@ -0,0 +1,172 @@
...
+    test various parts of blueflood.  The code is structured so that
+    the thread type is determined by the thread num.  The grinder
+    properties file determines how many of each type to create based
+    on the "ingest_concurrency" and "query_concurrency" options.
+
+    """
+    thread_type = None
+    server_num = thread_num
+
+    for x in self.types:
```
chinmay-gupte: Can you explain this logic with the thread number? I spent some time but cannot grok it.

contrib/grinder/scripts/utils.py
```diff
@@ -0,0 +1,172 @@
...
+
+
+RAND_MAX =  982374239
+
+class ThreadManager(object):
+  #keep track of the various thread types
+  types = []
+
+  @classmethod
+  def add_type(cls, type):
```
chinmay-gupte: What is exactly a `cls` object?

contrib/grinder/scripts/utils.py
```diff
@@ -0,0 +1,172 @@
...
+    if self.position >= max_position:
+      self.position = 0
+      sleep_time = self.finish_time - int(self.time())
+      self.finish_time += (default_config['report_interval'] / 1000)
+      if sleep_time < 0:
+        #return error
+        logger("finish time error")
+      else:
+        logger("pausing for %d" % sleep_time)
+        self.sleep(sleep_time)
```
chinmay-gupte: System dependent sleeps are not accurate, they depend a lot on the system load, just pointing it out. Also, I am not sure of any better way of doing it.

contrib/grinder/scripts/blueflood.py
```diff
@@ -0,0 +1,87 @@
...
+            'metricValue': random.randint(0, RAND_MAX),
+            'ttlInSeconds': (2 * 24 * 60 * 60),
+            'collectionTime': time}
+
+  def generate_payload(self, time, batch):
+    payload = map(lambda x:self.generate_metric(time,*x), batch)
+    return json.dumps(payload)
+
+  def ingest_url(self):
+    return "%s/v2.0/tenantId/ingest/multi" % default_config['url']
```
chinmay-gupte: Is there value in having single tenant ingestion endpoint tests as well?

contrib/grinder/scripts/blueflood.py
```diff
@@ -0,0 +1,87 @@
+import random
+try: 
+  from com.xhaus.jyson import JysonCodec as json
+except ImportError:
+  import json
+from utils import *
+from net.grinder.script import Test
+from net.grinder.plugin.http import HTTPRequest
+
+class IngestThread(AbstractThread):
```
chinmay-gupte: This class can have a better name may be `ingest.py` because it makes only ingestion requests?

contrib/grinder/scripts/query.py
```diff
@@ -0,0 +1,128 @@
...
+                        random.randint(0, default_config['metrics_per_tenant']))
+    metrics_list = map(self.generate_metric_name, range(metrics_count))
+    return json.dumps(metrics_list)
+
+  def generate_multiplot(self, time, logger):
+    tenant_id = random.randint(0, default_config['num_tenants'])
+    payload = self.generate_multiplot_payload()
+    to = time
+    frm = time - self.one_day
+    resolution = 'FULL'
```
chinmay-gupte: Is there value in making the resolution for mplot configurable? I think there is as we can test things like rollups on read only with non-raw granularities?

contrib/grinder/scripts/query.py
```diff
@@ -0,0 +1,128 @@
...
+    #divide the total work for this worker process into the work for this thread
+    start_query, end_query = self.generate_job_range(self.num_queries_for_current_node,
+                                                                  self.num_threads(),
+                                                                  thread_num)
+    self.num_queries_for_current_thread = end_query - start_query
+
+  def get_query_fn(self):
+    # Randomly select a query type based on the proportion of the various types
+    # to the total number of queries per interval
+    num = random.randint(0, self.total_queries)
```
chinmay-gupte: I am not sure this will generate queries deterministically...like suppose the total queries being generated are 100 with 10 splot, 10 mplot and 80 metrics list queries, the probability that splot queries will get generated is 0.1 which does not necessarily mean there will be 10 splot queries/min right?

contrib/grinder/scripts/tests.py
```diff
@@ -0,0 +1,230 @@
...
+    random.randint = self.real_randint
+    utils.AbstractThread.time = self.real_time
+    utils.AbstractThread.sleep = self.real_sleep
+
+
+#if __name__ == '__main__':
+unittest.TextTestRunner().run(unittest.TestLoader().loadTestsFromTestCase(BluefloodTests))
+
+
+
```
chinmay-gupte: NIT: extra new lines.

contrib/grinder/scripts/utils.py
```diff
@@ -0,0 +1,172 @@
...
+      if not entry.key.startswith("grinder.bf."):
+        continue
+      k = entry.key.replace("grinder.bf.","")
+      default_config[k] = self.convert(entry.value)
+
+  def __init__(self, grinder):
+    self.tot_threads = 0
+    self.concurrent_threads = 0
+    self.setup_config(grinder)
+    if self.tot_threads != self.concurrent_threads:
```
chinmay-gupte: If tot threads and concurrent threads are always equal, why have different config options for them?

contrib/grinder/scripts/utils.py
```diff
@@ -0,0 +1,172 @@
...
+  def generate_unit(self, tenant_id):
+    unit_number = tenant_id % 6
+    return units_map[unit_number]
+
+  def check_position(self, logger, max_position):
+    """Sleep if finished all work for report interval"""
+    if self.position >= max_position:
+      self.position = 0
+      sleep_time = self.finish_time - int(self.time())
+      self.finish_time += (default_config['report_interval'] / 1000)
```
chinmay-gupte: What happens if the time for this batch exceeds? Will the tool catch up?

contrib/grinder/README.md
```diff
@@ -0,0 +1,54 @@
+# Blueflood Grinder Integration
+##Intro
+Grinder is a distributed load testing tool described [here](http://grinder.sourceforge.net/g3/getting-started.html)
+
+This code defines implementations of grinder worker threads meant to repeatedly invoke the required number of BF api calls during each "reporting interval"
+
+It also includes the infrastructure to divide the total work described in the grinder properties file across all the workers in the distrubuted system.
```
chinmay-gupte: NIT: `distributed`

chinmay-gupte:
@GeorgeJahad I know we can track the metrics/queries getting generated after receiving them in blueflood using the instrumentation we have, but what do you think about having some instrumentation in the tool itself, will it be useful? 
One use case I can think of is that we can match the numbers getting generated in the tool vs those getting generated in blueflood and possibly be confident that the tool is not getting bottlenecked.

contrib/grinder/properties/grinder.properties
GeorgeJahad: No, the check_position method in utils.py maintains it

contrib/grinder/scripts/grinder.py
GeorgeJahad: It was a bad comment leftover from a cut and paste.  I've removed it.

contrib/grinder/properties/grinder.properties.staging
GeorgeJahad: Yes, one is to show how to run the tests.  the other to run in staging

contrib/grinder/scripts/utils.py
GeorgeJahad: I've hopefully fixed the doc here to make it clearer.

contrib/grinder/scripts/utils.py
GeorgeJahad: It's just a class object, (in this case, the ThreadManager class.)

contrib/grinder/scripts/utils.py
GeorgeJahad: It's ok because the check_position method takes that into account

contrib/grinder/scripts/blueflood.py
GeorgeJahad: there is but it's out of scope here.  I'll add a jira ticket

contrib/grinder/scripts/blueflood.py
GeorgeJahad: done

contrib/grinder/scripts/query.py
GeorgeJahad: there is but it's out of scope here.  I'll add a jira ticket

contrib/grinder/scripts/query.py
GeorgeJahad: you are right.  restructured code and confirmed that it now is deterministic on staging.

contrib/grinder/scripts/tests.py
GeorgeJahad: fixed

contrib/grinder/scripts/utils.py
GeorgeJahad: fixed comments to explain this

contrib/grinder/scripts/utils.py
GeorgeJahad: No it won't.  It basically means BF is overloaded and I need to stop and find out where the bottleneck is.

GeorgeJahad:
Adding more metrics for the tool itself is a good idea.  i'll add a jira ticket.

contrib/grinder/scripts/tests.py
```diff
@@ -0,0 +1,234 @@
...
+  def tearDown(self):
+    random.shuffle = self.real_shuffle
+    random.randint = self.real_randint
+    utils.AbstractThread.time = self.real_time
+    utils.AbstractThread.sleep = self.real_sleep
+
+#if __name__ == '__main__':
```
chinmay-gupte: Why is this block of code commented?

GeorgeJahad: another left over.  I'll remove it before merging

chinmay-gupte:
Do we need to add license files?  Please squash the commits before merging. Otherwise, LGTM. +1. Nice job!

PR #488, opened 2015-06-24
Refactored Module Loader Removed Discovery Module Loader and Events Module Loader Created a generic Module Loader

goru97:


goru97:
Tried Creating using Hub Command Line tool but it's trying to merge into master so closing it.

PR #487, opened 2015-06-24
Annotations: Pass 2

chinmay-gupte:
Reverts rackerlabs/blueflood#486

PR #492, opened 2015-06-29
Annotations Pass 2

tilogaat:
Continuing from #487 
Merging the same branch reviewed in #487 into the annotations feature branch.


tilogaat:
Merging because it has already been +1d

PR #489, opened 2015-06-24
Annotations module loader

goru97:
Deleted QueryModuleLoader and EventModuleLoader and created a Generic Module Loader for loading any Module.

PR #493, opened 2015-06-29
Annotations module loader : pass 2

tilogaat:
PR into the feature branch. 

Continuation from #489


tilogaat:
Merging because #489 has been +1d

PR #495, opened 2015-07-07
RFR: Adding checkstyle,findbugs and PMD rules to Blueflood

tilogaat:
Import the xml profile into Intellij and run it against your changeset to begin with. 

This ruleset may seem intimidating, but at first it's only a way to get thinking about style issues in your code. Though we want our code to obey good style, we do not want to create too much noise. 

Suggestions for use (to make it look less scary):
1. Run this profile using QAPlugs (or any other tool) and fix all issues against your changeset
2. Run it only against code classes (as opposed to Test)
3. Try to add only a single commit for style issues (to prevent noise).


tilogaat:
Not getting used , closing it for now.

PR #499, opened 2015-07-21
Annotations query for grinder

goru97:
Added annotations query endpoint to grinder.


contrib/grinder/scripts/query.py
```diff
@@ -96,13 +96,30 @@ def generate(self, time, logger):
...
+
+class AnnotationsQuery(AbstractQuery):
+  query_interval_name = 'annotations_queries_per_interval'
+  query_name = "AnnotationsQuery"
+  test_number = 5
+
+
+  def generate(self, time, logger):
+    tenant_id = random.randint(0, default_config['num_tenants'])
+    to = time
```
GeorgeJahad: I think Tilo has changed annotations ingest to use it's own num_tenants; if so you'll want to use that here as well.

contrib/grinder/scripts/tests.py
```diff
@@ -114,14 +116,23 @@ def test_init_process(self):
 
     # confirm that the number of queries is correctly distributed across
     #  each thread in this worker process
+
```
GeorgeJahad: use num_nodes instead of "2" in the calculations below.

goru97: Done

contrib/grinder/scripts/query.py
goru97: She's actually creating new variable annotations_per_tenant but using existing num_tenants.

tilogaat: @goru97 i am using the existing num_tenants currently, but planning to have a separate num_annotations_tenants

GeorgeJahad:
+1

goru97:
Merged using git! Closing it now.

PR #501, opened 2015-07-28
Fix time() to return time in ms so that metrics are properly timestamped in the grinder

GeorgeJahad:


contrib/grinder/scripts/utils.py
```diff
@@ -177,8 +177,8 @@ def check_position(self, logger, max_position):
 
   @classmethod
   def time(cls):
-    return time.time()
+    return int(time.time() * 1000)
 
   @classmethod
   def sleep(cls, x):
-    return time.sleep(x)
+    return time.sleep(x/1000)
```
tilogaat: why are we multiplying by 1000 https://github.com/rackerlabs/blueflood/pull/501/files#diff-413f7f2bf24ab7bf69b97eda82cdd766R66 and dividing by 1000 here?

GeorgeJahad: I wanted to keep the time functions consistent.  They both use milliseconds.  It seemed confusing to me to have one output milliseconds and the other take seconds.

tilogaat:
+1

PR #502, opened 2015-08-10
RFR: Multi tenant Aggregated endpoint

tilogaat:
This PR attempts to add another endpoint for sending aggregated metrics for multiple tenants in the same request.
The current single tenant aggregated endpoint already has the capability to include tenantid and timestamp as shown here:

```
{
         "tenantId":"5405532",
         "timestamp":1439231324000,
         "gauges":[ 
            {
               "name":"G200ms",
                "value":145
            }
        ],
        "counters":[
           {
                ...
          }
        ]
}

```
This PR takes in a list of the above payload and stores the metrics appropriately:

```
[
    {
        "tenantId": "5405532",
        "timestamp": 1439231324000,
        "gauges": [
            {
                "name": "G200ms",
                "value": 145
            }
        ]
    },
    {
        "tenantId": "706456",
        "timestamp": 1439231324001,
        "gauges": [
            {
                "name": "another_tenant_G200ms",
                "value": 200
            }
        ]
    }
]
```

TODO: 
~~Need to add integration tests~~


VinnyQ:
lgtm :+1:

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpAggregatedIngestionHandler.java
```diff
@@ -46,17 +46,17 @@
...
-public class HttpStatsDIngestionHandler implements HttpRequestHandler {
+public class HttpAggregatedIngestionHandler implements HttpRequestHandler {
     
-    private static final Logger log = LoggerFactory.getLogger(HttpStatsDIngestionHandler.class);
+    private static final Logger log = LoggerFactory.getLogger(HttpAggregatedIngestionHandler.class);
     
-    private static final Timer handlerTimer = Metrics.timer(HttpStatsDIngestionHandler.class, "HTTP statsd metrics ingestion timer");
-    private static final Counter requestCount = Metrics.counter(HttpStatsDIngestionHandler.class, "HTTP Request Count");
+    private static final Timer handlerTimer = Metrics.timer(HttpAggregatedIngestionHandler.class, "HTTP statsd metrics ingestion timer");
+    private static final Counter requestCount = Metrics.counter(HttpAggregatedIngestionHandler.class, "HTTP Request Count");
```
GeorgeJahad: The metric names have been changed here, so any dashboards will have to be updated

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpAggregatedMultiIngestionHandler.java
```diff
@@ -0,0 +1,103 @@
...
+        final String body = request.getContent().toString(Constants.DEFAULT_CHARSET);
+        try {
+            // block until things get ingested.
+            requestCount.inc();
+            List<Bundle> bundleList = createBundleList(body);
+            for (Bundle bundle : bundleList){
+                MetricsCollection collection = new MetricsCollection();
+                collection.add(PreaggregateConversions.buildMetricsCollection(bundle));
+                ListenableFuture<List<Boolean>> futures = processor.apply(collection);
+                List<Boolean> persisteds = futures.get(timeout.getValue(), timeout.getUnit());
```
GeorgeJahad: This looks like you are going to block for each bundle in the bundleList.  Isn't that going to be very slow?  It would probably be create a list of metrics, instead of a list of bundles and then simultaneously block on each individual metric, rather than each individual bundle.

blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/inputs/handlers/HttpAggregatedMultiIntegrationTest.java
```diff
@@ -0,0 +1,110 @@
+package com.rackspacecloud.blueflood.inputs.handlers;
```
GeorgeJahad: File Header block/copyright?

blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/inputs/handlers/HttpAggregatedMultiIntegrationTest.java
```diff
@@ -0,0 +1,110 @@
...
+        while (curLine != null) {
+            sb = sb.append(curLine);
+            curLine = reader.readLine();
+        }
+        String json = sb.toString();
+
+        return json;
+    }
+
+}
```
GeorgeJahad: This class has so much in common with the other; how hard would it be to make one the base class of the other?

GeorgeJahad: In fact, what if, instead of 2 classes we had the same old class that now either accepts an object or a list of objects?

tilogaat: @GeorgeJahad Can you clearly specify which 2 classes you are speaking about?

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpAggregatedMultiIngestionHandler.java
tilogaat: How so? Wouldn't the resultant time spent blocking be the same?

blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/inputs/handlers/HttpAggregatedMultiIntegrationTest.java
tilogaat: does 70b2122 fix your concern?

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpAggregatedMultiIngestionHandler.java
tilogaat: Fixed in  d2f92e9

GeorgeJahad: Looks much better!

GeorgeJahad:
Sorry for the confusion.  I was refering to the two implementation classes:
HttpAggregatedIngestionHandler.java and HttpAggregatedMultiIngestionHandler.java

They are so similar, wouldn't it be better to create one that either takes a single object, or a list of objects?

tilogaat:
@GeorgeJahad  The refactoring does not look great for the Json processing. 
https://gist.github.com/tilogaat/271f2dc8112849fe0e0f

We need to create a better structure for the base handler to enable this. I think it's worthwhile doing it similar to HttpMultitenantMetricsIngestionHandler. I will create another story for it though.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpAggregatedMultiIngestionHandler.java
```diff
@@ -0,0 +1,122 @@
+/*
```
GeorgeJahad: Since the major change in this endpoint is that each entry in the bundlelist has it's own timestamp and tenant, I would add a simple test for each of those.  Other than that, this PR looks ready to go.

GeorgeJahad:
+1

PR #505, opened 2015-08-19
Ingest enum payload from single-tenant aggregated endpoint

tilogaat:
*What*
Adds the entire pipeline for ingesting enums 

*How*
Adds a new rollup type EnumRollup, adds serialization-deserialization for Enums. With this code, we can ingest enums using the single tenant as follows:
```
curl -i -XPOST 'http://localhost:19000/v2.0/6076299/ingest/aggregated' -d '
{
    "tenantId": "5405532",
    "timestamp": 1438207529002,
    "enums": [
        {
            "name": "call_xyz_api",
            "value": "500"
        }
    ]
}'
```
You will observe that the enum call_xyz_api is stored in metrics_preaggregated_full 
The serialization format is the hashcode of the enum value followed by the count of the number of the times that value was seen. 

*Why*
This PR lays down the infrastructure to get enum metrics into the system. We can then add caching/validation logic of enums, rollups, queries, multi tenant ingestion on top of this.


blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/Constants.java
```diff
@@ -36,6 +36,8 @@
     public static final byte VERSION_1_COUNTER_ROLLUP = 0;
     public static final byte VERSION_1_SET_ROLLUP = VERSION_1_ROLLUP; // don't change this.
```
goru97: NIT: extra line.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/types/EnumRollup.java
```diff
@@ -0,0 +1,41 @@
+package com.rackspacecloud.blueflood.types;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class EnumRollup implements Rollup {
+    private Map<Long, Long> en2Value = new HashMap<Long, Long>();
+
+    public EnumRollup withObject(Long valueNameHashcode, Long value) {
```
GeorgeJahad: Don't we need to store the full string name here, and not just the hash code?  Other wise how does the query endpoint return it?

GeorgeJahad: NIT: withObject() seems like a wierd name here.  What is the "object"?

blueflood-core/src/test/java/com/rackspacecloud/blueflood/io/serializers/EnumRollupSerializationTest.java
```diff
@@ -0,0 +1,69 @@
...
+import java.io.*;
+import java.nio.ByteBuffer;
+
+public class EnumRollupSerializationTest {
+
+    @Test
+    public void testEnumV1RoundTrip() throws IOException {
+        EnumRollup e0 = new EnumRollup().withObject((long)"enumValue1".hashCode(), (long)1);
+        EnumRollup e1 = new EnumRollup().withObject((long)"t4.enum.abcdefg.hijklmnop.qrstuvw.xyz.ABCDEFG.HIJKLMNOP.QRSTUVW.XYZ.abcdefg.hijklmnop.qrstuvw.xyz.met".hashCode(), 34454722343L);
+
```
GeorgeJahad: When testing a container, (like EnumRollups), I prefer to see the test add at least two elements to the container.  Here you create 2 EnumRollups, with one element each.

GeorgeJahad:
why are there no tests for enum rollups here:
https://github.com/rackerlabs/blueflood/blob/master/blueflood-core/src/test/java/com/rackspacecloud/blueflood/io/serializers/SerializationTest.java

https://github.com/rackerlabs/blueflood/blob/master/blueflood-core/src/test/java/com/rackspacecloud/blueflood/io/serializers/IMetricSerializerTest.java

tilogaat:
@GeorgeJahad SerializationTest.java has tests only for basic rollups. But I agree that those tests seem comprehensive. I will add a story to our backlog to add more comprehensive tests for aggregated metrics.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/serializers/NumericSerializer.java
```diff
@@ -71,6 +66,7 @@
         static final byte B_TIMER = (byte)'T';
         static final byte B_SET = (byte)'S';
         static final byte B_GAUGE = (byte)'G';
+        static final byte B_ENUM = (byte) 'E';
```
VinnyQ: NIT:  should remove the white space between the cast and the letter like the other constants above.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/serializers/NumericSerializer.java
```diff
@@ -55,6 +48,7 @@
     public static AbstractSerializer<TimerRollup> timerRollupInstance = new TimerRollupSerializer();
     public static AbstractSerializer<SetRollup> setRollupInstance = new SetRollupSerializer();
     public static AbstractSerializer<GaugeRollup> gaugeRollupInstance = new GaugeRollupSerializer();
+    public static AbstractSerializer<EnumRollup> enRollupInstance = new EnumRollupSerializer();
```
VinnyQ: NIT:  why not just spell out "enumRollupInstance"?  it's only 2 letters more and it reads clearer.

VinnyQ:
:+1: looks good to me, @tilogaat !

PR #506, opened 2015-08-19
cassandra_connection_fix

goru97:
.setInitConnsPerHost() was used twice 
First time as .setInitConnsPerHost(connsPerHost)
Second time as .setInitConnsPerHost(connsPerHost / 2).

So, remove the first one, as second one was overriding.


tilogaat:
+1

PR #511, opened 2015-08-24
Multi-tenant enums ingestion

tilogaat:
*what:*
test and renaming to check that enums can be ingested using multi-tenant endpoint

*why:*
multi-tenant aggregated endpoint was developed almost in parallel with the enums-ingestion. Making sure that the twain can meet. 

*how:*
Renamed EnumRollup to BluefloodEnumRollup
Renamed sample_multi_bundle.json to sample_multi_aggregated_payload.json
Added enums payload to it.
Added a test to check enums serializability with the multi endpoint


VinnyQ:
:+1: 

Sorry I didn't catch the "BluefloodEnumRollup" naming in the previous commits.

VinnyQ:
good tests

tilogaat:
Thank you @VinnyQ

PR #514, opened 2015-08-31
Enum relational serialization

tilogaat:
*what*
Relational approach to serializing enums

*why*
If we serialize all the enum values as strings, each time an enum data point is ingested, we have an expensive disk operation on our hands. Therefore, this PR explores the idea of storing only the hashcode, rather than the whole string of the enum value. 

*how*
We serialize the hashcode in metrics_preaggregated_full rather than the whole string name. However, we also serialize the hashcode of the value, with the actual value in a separate column_family, as follows:

```
 key             | column1    | value
-----------------+------------+------------------
 123456.enum1 | -841539219 | anotherenumvalue
 123456.enum1 |  262599952 |        enumvalue
```
Since the column is the hashcode, while querying we will just get the value stored for the given locator and hashcode. 

Added a test for showing how the joins can be done across metrics_enum and metrics_preaggregated_full
~~Todo:~~
~~Working on a PoC to make sure that querying the actual string values is feasible~~


VinnyQ:
@tilogaat @GeorgeJahad This is a great idea tilo!  Going to look over your code and cassandra schema changes to see how you will implement this and add some comments.

src/cassandra/cli/load.script
```diff
@@ -8,6 +8,7 @@ CREATE COLUMN FAMILY metrics_discovery WITH column_type='Standard' AND comparato
 CREATE COLUMN FAMILY metrics_state WITH column_type='Standard' AND comparator='UTF8Type' AND key_validation_class='LongType' AND default_validation_class='LongType';
 CREATE COLUMN FAMILY metrics_metadata WITH column_type='Standard' AND comparator='UTF8Type' AND key_validation_class='UTF8Type' AND default_validation_class='BytesType';
 CREATE COLUMN FAMILY metrics_string WITH column_type='Standard' AND comparator='LongType' AND key_validation_class='UTF8Type' AND default_validation_class='UTF8Type';
+CREATE COLUMN FAMILY metrics_enum1 WITH column_type='Standard' AND comparator='LongType' AND key_validation_class='UTF8Type' AND default_validation_class='UTF8Type';
```
VinnyQ: Not too fond of the the column family name "metrics_enum1".  How about something like "metrics_enum_values" or "metrics_enum_hashes"?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxMetricsWriter.java
```diff
@@ -34,7 +34,7 @@ public void insertFullMetrics(Collection<Metric> metrics) throws IOException {
     }
 
     @Override
-    public void insertPreaggreatedMetrics(Collection<IMetric> metrics) throws IOException {
+    public void insertPreaggreatedMetrics(Collection<IMetric> metrics) throws Exception {
```
VinnyQ: What code changes in this class is throwing more exceptions other than IOException that cause you to change this throws definition?

blueflood-core/src/test/java/com/rackspacecloud/blueflood/io/serializers/IMetricSerializerTest.java
```diff
@@ -52,7 +52,7 @@ public void testCounterSerialization() throws IOException {
 
     @Test
     public void testEnumSerialization() throws IOException {
-        String enumRollupString = "{\"type\":\"BluefloodEnumRollup\",\"count\":2,\"en2Value\":{\"-1365468863\":2,\"233047280\":5},\"hashes\":{\"-1365468863\":2,\"233047280\":5}}";
+        String enumRollupString = "{\"type\":\"BluefloodEnumRollup\",\"count\":1,\"hashedEnum2Value\":{\"-563243779\":1},\"hashes\":{\"-563243779\":1},\"rawEnumValues\":[\"somevalue\"],\"rawValues\":[\"somevalue\"]}";
 
         BluefloodEnumRollup enumDeserialized = mapper.readValue(enumRollupString, BluefloodEnumRollup.class);
         String enumSerialized = mapper.writeValueAsString(enumDeserialized);
```
VinnyQ: Good serialization test.  Can we also have tests of inserting the hashed enum values and names, and retrieving the hashed values and deserializing them?

tilogaat: @VinnyQ yea that's a good point, I need to make some code changes to AstyanaxReader class for it.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxMetricsWriter.java
tilogaat: @VinnyQ this is an oversight on my part. I will change it back.

src/cassandra/cli/load.script
tilogaat: @VinnyQ how about metrics_enum?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxWriter.java
```diff
@@ -161,6 +161,12 @@ private final void insertLocator(Locator locator, MutationBatch mutationBatch) {
                 .putEmptyColumn(locator, LOCATOR_TTL);
     }
 
+    private final void insertEnumLocator(Locator locator, BluefloodEnumRollup rollup, MutationBatch mutationBatch) {
```
GeorgeJahad: Would insertEnumValueNames() be a better name for this method?

src/cassandra/cli/load.script
VinnyQ: @tilogaat that sounds good!

blueflood-core/src/test/java/com/rackspacecloud/blueflood/io/serializers/IMetricSerializerTest.java
tilogaat: Added a pretty comprehensive test in 2da9471

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxWriter.java
tilogaat: Fixed d4b4b3f

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxMetricsWriter.java
tilogaat: fixed in d4b4b3f

src/cassandra/cli/load.script
tilogaat: fixed d4b4b3f

blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/inputs/handlers/HttpHandlerIntegrationTest.java
```diff
@@ -267,6 +269,19 @@ public void testHttpAggregatedMultiIngestionHappyCase() throws Exception {
                 locator1, new Range(1439231323000L, 1439231325000L), CassandraModel.getColumnFamily(BluefloodCounterRollup.class, Granularity.FULL));
         Assert.assertEquals(1, points1.getPoints().size());
 
+        final Locator locator2 = Locator.createLocatorFromPathComponents("5405577", "call_xyz_api");
+        Points<BluefloodEnumRollup> points2 = new FakeEnumAstyanaxReader().getInstance().getDataToRoll(BluefloodEnumRollup.class,
+                locator2, new Range(1439231323000L, 1439231325000L), CassandraModel.getColumnFamily(BluefloodEnumRollup.class, Granularity.FULL));
+        Assert.assertEquals(1, points1.getPoints().size());
```
GeorgeJahad: This assert is the same as the one above, on line 270. Is it supposed to be for points2?

blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/inputs/handlers/HttpHandlerIntegrationTest.java
```diff
@@ -267,6 +269,19 @@ public void testHttpAggregatedMultiIngestionHappyCase() throws Exception {
...
+        for (Long timestamp : points1.getPoints().keySet()) {
```
GeorgeJahad: why are you cycling through points1?

blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/inputs/handlers/HttpHandlerIntegrationTest.java
```diff
@@ -267,6 +269,19 @@ public void testHttpAggregatedMultiIngestionHappyCase() throws Exception {
...
+            BluefloodEnumRollup er = points2.getPoints().get(timestamp).getData();
+            for (String value : er.getStringEnumValuesWithCounts().keySet()) {
+                Assert.assertEquals("500",value);
```
GeorgeJahad: where is the "500" coming from?  how did the CF_METRICS_ENUM table get set to '500' for this locator?

tilogaat: it's in the sample payload file (blueflood-http/src/test/resources/sample_multi_aggregated_payload.json). It's an integration test, ingesting the sample payload first and then querying for it, which is why it's in the metrics_enum CF.

blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/inputs/handlers/HttpHandlerIntegrationTest.java
tilogaat: yea it's a typo, it should cycle through points2. The tests pass because the timestamp is common for the payload and the number of points is the same = 1.

blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/inputs/handlers/HttpHandlerIntegrationTest.java
tilogaat: yup will fix.

GeorgeJahad:
+1

blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/inputs/handlers/HttpHandlerIntegrationTest.java
```diff
@@ -413,4 +427,66 @@ public static void shutdown() {
...
+            return hash2enumValues;
+        }
+
+        private ColumnList<Long> getColumnsFromEnumCF(final Locator locator) {
+            final Map<Locator, ColumnList<Long>> columns = new HashMap<Locator, ColumnList<Long>>();
+
+            try {
+                OperationResult<Rows<Locator, Long>> query = getKeyspace()
+                        .prepareQuery(CassandraModel.CF_METRICS_ENUM)
+                        .getKeySlice(locator)
```
GeorgeJahad: NIT: It probably makes sense to use getKey() instead of getKeySlice() here, since you only want to get the single locator row back.  Maybe that's how we should do it when we move the class to AstyanaxReader.

PR #515, opened 2015-09-03
CORS_support_for_events

goru97:
What:
This PR solves two changes:
1. Support for multiple methods (GET, POST, OPTIONS etc) for same URL.
2. Support for OPTIONS method for events endpoint, which is used by the browser as preflight query before sending actual GET request with Authorization token.

Why:
1. Blueflood plugin for Grafana is under-construction that will run in browser's engine and will directly contact blueflood. 
2. These changes were required to support Cross Origin Resource Sharing with blueflood.

How: 
1. Fixed RouteMatcher class.
2. Added options method in HttpMetricDataQueryServer class.

### DIFF
--- a/blueflood-http/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/HttpEventsQueryHandler.java
+++ b/blueflood-http/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/HttpEventsQueryHandler.java
@@ -5,6 +5,7 @@ import com.rackspacecloud.blueflood.http.DefaultHandler;
 import com.codahale.metrics.Timer;
 import com.rackspacecloud.blueflood.http.HTTPRequestWithDecodedQueryParams;
 import com.rackspacecloud.blueflood.http.HttpRequestHandler;
+import com.rackspacecloud.blueflood.http.HttpResponder;
 import com.rackspacecloud.blueflood.io.EventsIO;
 import com.rackspacecloud.blueflood.utils.DateTimeParser;
 import com.rackspacecloud.blueflood.utils.Metrics;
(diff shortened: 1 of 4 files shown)


blueflood-http/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/HttpEventsQueryHandler.java
```diff
@@ -27,35 +28,46 @@ public HttpEventsQueryHandler(EventsIO searchIO) {
...
+            }
+        }
+        else if(request.getMethod().toString().equalsIgnoreCase("OPTIONS")){
+            HttpResponse response = new DefaultHttpResponse(HttpVersion.HTTP_1_1, HttpResponseStatus.OK);
+            response.setHeader("Access-Control-Allow-Origin", "*");
+            response.setHeader("Access-Control-Allow-Methods", "GET");
+            response.setHeader("Access-Control-Allow-Headers", "X-Auth-Token, Accept");
+            response.setHeader("Access-Control-Max-Age", "1728000");
+            HttpResponder.respond(ctx, request, response);
         }
```
VinnyQ: Do we need a final else case here to handle where the request is neither GET or OPTIONS?

VinnyQ: Oh never mind, this class is only currently invoked in either of those two cases.  It's cool no need for a catchall else.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/HttpEventsQueryHandler.java
```diff
@@ -27,35 +28,46 @@ public HttpEventsQueryHandler(EventsIO searchIO) {
...
     public void handle(ChannelHandlerContext ctx, HttpRequest request) {
-        final String tenantId = request.getHeader("tenantId");
-        HttpResponseStatus status = HttpResponseStatus.OK;
 
-        ObjectMapper objectMapper = new ObjectMapper();
-        String responseBody = null;
-        final Timer.Context httpEventsFetchTimerContext = httpEventsFetchTimer.time();
-        try {
-            HTTPRequestWithDecodedQueryParams requestWithParams = (HTTPRequestWithDecodedQueryParams) request;
-            Map<String, List<String>> params = requestWithParams.getQueryParams();
```
VinnyQ: Not sure how it got missed here, but can you add the line "Tracker.track(request);" here for debugging issues in production via jolokia/jmx?  See HttpMetricsIngestionHandler.handle() method as an example.  Thanks!

tilogaat: good point @VinnyQ 
I think we added the annotations endpoint at the same time as you were adding the jmx stuff, which is why it got left out.

goru97: The response is being sent by DefaultHandler for the get request; There is "Tracker.trackResponse(request, response);" over there. I'm not sure if you still want this?

blueflood-http/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/HttpEventsQueryHandler.java
tilogaat: Can we check if the access-control-allow-origin and all other headers can be added to repose?

tilogaat:
+1 after the repose question is answered

PR #516, opened 2015-09-04
Initial commit Grafana Blueflood Plugin

goru97:
*What:*
Adding Blueflood as a datasource in grafana for annotations only (as of now).

*Why:*
To leverage our newly publisheds end-point for Annotations/Events.

*How:*
Adding the Blueflood plugin in grafana to talk directly to Blueflood DB.


tilogaat:
+1

PR #517, opened 2015-09-09
Created BadMetrics table and access primitives

GeorgeJahad:
What:
Table for storing bad metrics and the read/write primitives to update it.

Why:
To be able to check incoming metrics against a list of know bad metrics.

How:
A Cassandra table with a Locator per row, and a timestamp for the single column

### DIFF
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
@@ -17,6 +17,7 @@
 package com.rackspacecloud.blueflood.io;
 
 import com.codahale.metrics.Timer;
+import com.google.common.base.Function;
 import com.google.common.collect.ArrayListMultimap;
 import com.google.common.collect.HashBasedTable;
 import com.google.common.collect.ListMultimap;
@@ -27,6 +28,7 @@ import com.netflix.astyanax.connectionpool.exceptions.ConnectionException;
 import com.netflix.astyanax.connectionpool.exceptions.NotFoundException;
 import com.netflix.astyanax.model.*;
 import com.netflix.astyanax.query.RowQuery;
+import com.netflix.astyanax.recipes.reader.AllRowsReader;
 import com.netflix.astyanax.serializers.AbstractSerializer;
 import com.netflix.astyanax.serializers.BooleanSerializer;
 import com.netflix.astyanax.serializers.StringSerializer;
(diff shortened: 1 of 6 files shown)


blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/CassandraModel.java
```diff
@@ -67,6 +67,10 @@
             LongSerializer.get(),
             SlotStateSerializer.get());
 
+    public static final ColumnFamily<Locator, Long> CF_METRICS_BAD_METRICS = new ColumnFamily<Locator, Long>("metrics_bad_metrics",
```
tilogaat: should we make it clear that these are bad enum metrics?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
```diff
@@ -520,4 +523,40 @@ private Points getPointsFromColumns(ColumnList<Long> columnList, RollupType roll
...
+                return true;
+            }
+        };
+
+        ColumnFamily CF = CassandraModel.CF_METRICS_BAD_METRICS;
+        Timer.Context ctx = Instrumentation.getBatchReadTimerContext(CF);
+        try {
+            new AllRowsReader.Builder<Locator, Long>(keyspace, CassandraModel.CF_METRICS_BAD_METRICS)
+                    .withColumnRange(null, null, false, 0)
+                    .withPartitioner(null) // this will use keyspace's partitioner
```
tilogaat: Why do we need the partitioner?

VinnyQ:
:+1:  lgtm.  

You will be using the writeBadMetrics() method during rollup to add the locator (tenantId:metric_name) to this table if num distinct enum values for that metric are > threshold?

And getBadMetrics() method during ingestion to see if the metric is flagged as a bad metric?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/CassandraModel.java
GeorgeJahad: will do.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
GeorgeJahad: we don't.  i'll remove

GeorgeJahad:
Yes, writeBadMetric() and getBadMetrics() are to be used as you describe.

tilogaat:
+1

PR #518, opened 2015-09-11
Http Handler Improvement

goru97:
*What: 
Added Headers Parameter in sendResponse Method of DefaultHandler class.

*Why:
To enable extended handlers (like HttpAggregatedIngestionHandler, HttpEventsIngestionHandler etc) send custom headers in response.

### DIFF
--- a/blueflood-http/src/main/java/com/rackspacecloud/blueflood/http/DefaultHandler.java
+++ b/blueflood-http/src/main/java/com/rackspacecloud/blueflood/http/DefaultHandler.java
@@ -36,8 +36,9 @@ public class DefaultHandler implements HttpRequestHandler {
     }
 
     public static void sendResponse(ChannelHandlerContext channel, HttpRequest request, String messageBody, HttpResponseStatus status) {
-        sendResponse(channel, request, messageBody, null);
+        sendResponse(channel, request, messageBody, status, null);
     }
+    
     public static void sendResponse(ChannelHandlerContext channel, HttpRequest request, String messageBody, HttpResponseStatus status, Map<String, String> headers) {
 
         HttpResponse response = new DefaultHttpResponse(HttpVersion.HTTP_1_1, status);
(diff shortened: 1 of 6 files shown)


blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpAggregatedIngestionHandler.java
```diff
@@ -80,25 +80,25 @@ public void handle(ChannelHandlerContext ctx, HttpRequest request) {
             List<Boolean> persisteds = futures.get(timeout.getValue(), timeout.getUnit());
             for (Boolean persisted : persisteds) {
                 if (!persisted) {
-                    DefaultHandler.sendResponse(ctx, request, null, HttpResponseStatus.INTERNAL_SERVER_ERROR);
+                    DefaultHandler.sendResponse(ctx, request, null, HttpResponseStatus.INTERNAL_SERVER_ERROR, null);
```
VinnyQ: I am not fond of adding a new parameter argument to an existing functions and modifying any current code that needs it to have it call by passing "null".

May I recommend you create a wrapper method also called sendResponse() in the DefaultHandler class, only needing the original parameters, and have that called your modified method with the null as the headers parameter?

Then you can remove all these modification to the endResponse() calls and not have to pass in a "null" when it's not actually needed.

goru97: Can't agree more! :)

blueflood-http/src/main/java/com/rackspacecloud/blueflood/http/DefaultHandler.java
```diff
@@ -33,7 +36,18 @@ public void handle(ChannelHandlerContext ctx, HttpRequest request) {
     }
 
     public static void sendResponse(ChannelHandlerContext channel, HttpRequest request, String messageBody, HttpResponseStatus status) {
+        sendResponse(channel, request, messageBody, null);
+    }
```
VinnyQ: NIT:  spacing (need new line).  But wait and see if there's other changes you'll need to make from the other reviewers.

VinnyQ:
:+1: lgtm if all checks are passed.

PR #519, opened 2015-09-11
Enum rollups

tilogaat:
What:
Laying down the infrastructure for rolling up enums

How: 
Added the buildRollupFromEnumRollups method in EnumRollup class. Plugged it in with the RollupRunnable class. 

TODO:
~~add test to see if serialization works for rollups too~~
~~improve ingestion testing for rollups~~

### DIFF
--- a/blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/service/RollupRunnableIntegrationTest.java
+++ b/blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/service/RollupRunnableIntegrationTest.java
@@ -41,6 +41,7 @@ public class RollupRunnableIntegrationTest extends IntegrationTestBase {
     private final Locator gaugeLocator = Locator.createLocatorFromPathComponents("runnabletest", "gauge");
     private final Locator timerLocator = Locator.createLocatorFromPathComponents("runnabletest", "timer");
     private final Locator setLocator = Locator.createLocatorFromPathComponents("runnabletest", "set");
+    private final Locator enumLocator = Locator.createLocatorFromPathComponents("runnabletest", "enum");
     private final Locator normalLocator = Locator.createLocatorFromPathComponents("runnabletest", "just_some_data");
     
     private final Range range = new Range(0, 5 * 60 * 1000);
@@ -60,6 +61,7 @@ public class RollupRunnableIntegrationTest extends IntegrationTestBase {
         cache.put(gaugeLocator, cacheKey, RollupType.GAUGE.name());
         cache.put(timerLocator, cacheKey, RollupType.TIMER.name());
         cache.put(setLocator, cacheKey, RollupType.SET.name());
+        cache.put(enumLocator, cacheKey, RollupType.ENUM.name());
         // do not put normalLocator in the cache. it will constitute a miss.
         
         // put some full resolution data.
@@ -95,7 +97,11 @@ public class RollupRunnableIntegrationTest extends IntegrationTestBase {
             BluefloodSetRollup rollup = new BluefloodSetRollup().withObject(i);
             metric = new PreaggregatedMetric(time, setLocator, ttl, rollup);
             preaggregatedMetrics.add(metric);
-            
+
+            BluefloodEnumRollup enumRollup = new BluefloodEnumRollup().withHashedEnumValue((long)i, (long)5*(i+1));
+            metric = new PreaggregatedMetric(time, enumLocator, ttl, enumRollup);
+            preaggregatedMetrics.add(metric);
+
             metric = new Metric(normalLocator, i, time, ttl, "centipawns");
             normalMetrics.add(metric);
         }
@@ -159,6 +165,11 @@ public class RollupRunnableIntegrationTest extends IntegrationTestBase {
     public void testSetRollup() throws IOException {
         testRolledupMetric(setLocator, BluefloodSetRollup.class, BluefloodSetRollup.class);
     }
+
+    @Test
+    public void testEnumRollup() throws IOException {
+        testRolledupMetric(enumLocator, BluefloodEnumRollup.class, BluefloodEnumRollup.class);
+    }
     
     private void testRolledupMetric(Locator locator, Class fullResClass, Class rollupClass) throws IOException { 
         // full res has 5 samples.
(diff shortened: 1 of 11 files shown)


blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/service/RollupRunnableIntegrationTest.java
```diff
@@ -95,7 +97,11 @@ public void setUp() throws Exception {
             BluefloodSetRollup rollup = new BluefloodSetRollup().withObject(i);
             metric = new PreaggregatedMetric(time, setLocator, ttl, rollup);
             preaggregatedMetrics.add(metric);
-            
+
```
VinnyQ: You should move the hashed value calculation here to a separate that can be repeated call for other tests that you're going to be writing ;)

blueflood-core/src/main/java/com/rackspacecloud/blueflood/types/BluefloodEnumRollup.java
```diff
@@ -7,13 +10,22 @@
...
+
+        return this.withHashedEnumValue((long)valueName.hashCode(), value);
     }
 
     public BluefloodEnumRollup withHashedEnumValue(Long hashedEnumValue, Long value) {
-        this.hashedEnumValues2Count.put(hashedEnumValue, value);
+        long hashcount = 0;
+        if (this.hashedEnumValues2Count.containsKey(hashedEnumValue)) {
+            hashcount = this.hashedEnumValues2Count.get(hashedEnumValue);
+        }
```
VinnyQ: I don't understand the line below, why are we putting value+hashcount in hashedEnumValues2Count?  Shoudn't it just be the hashcount?

tilogaat: @VinnyQ 

value is a bad name for this variable, it should be incomingCount or something to that effect. 
This test case explains why we need to add the existingCount with the incomingCount. 
https://github.com/rackerlabs/blueflood/pull/519/files#diff-437d2681df3fc60b82f4ed2954b87af2R37

In a single 5 min rollup period, if we see the enum value 'call_xyz_api:OK' first and again after say 2 mins 'call_xyz_api:OK', we want the map to store 'OK' => 2 and hash('OK') => 2. Is that right? If we do not add the 2 values, then the last OK will overwrite the first and the only value stored in the maps will be 'OK' => 1

tilogaat: I fixed variable names now and hopefully it is clearer.

blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/service/RollupRunnableIntegrationTest.java
tilogaat: I just wanted to generate a long count for the values, so I just multiplied i+1 by 5 and cast it to long.

VinnyQ:
:+1:  lgtm

blueflood-core/src/main/java/com/rackspacecloud/blueflood/types/BluefloodEnumRollup.java
```diff
@@ -46,4 +58,24 @@ public boolean equals(Object obj) {
         return hashedEnumValues2Count.equals(other.hashedEnumValues2Count);
     }
 
+    public static BluefloodEnumRollup buildRollupFromEnumRollups(Points<BluefloodEnumRollup> input) throws IOException {
+        BluefloodEnumRollup enumRollup = new BluefloodEnumRollup();
+        for (Points.Point<BluefloodEnumRollup> point : input.getPoints().values()) {
+            BluefloodEnumRollup pointData = point.getData();
+            Map<Long, Long> incomingHashedEnums = pointData.getHashedEnumValuesWithCounts();
+            Map<Long, Long> currentHashedEnums = enumRollup.getHashedEnumValuesWithCounts();
```
GeorgeJahad: It seems redundant to put this line in the for loop.  Wouldn't it work the same if currentHashedEnums was set once outside the loop? or am i missing something

blueflood-core/src/main/java/com/rackspacecloud/blueflood/types/BluefloodEnumRollup.java
```diff
@@ -46,4 +58,24 @@ public boolean equals(Object obj) {
...
+
+            for (Long hash : incomingHashedEnums.keySet()) {
+                long incomingCount = incomingHashedEnums.get(hash);
+                if (currentHashedEnums.containsKey(hash)) {
+                    long currentCount = currentHashedEnums.get(hash);
+                    incomingCount+=currentCount;
+                }
+                enumRollup.getHashedEnumValuesWithCounts().put(hash, incomingCount);
```
GeorgeJahad: since you've created "currentHashedEnums" as a synonym for "enumRollup.getHashedEnumValuesWithCounts()", couldn't you use it here as well?  In other words, instead of this line, wouldn't it be clearer to just do:
           currentHashedEnums .put(hash, incomingCount);

blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/service/RollupRunnableIntegrationTest.java
```diff
@@ -187,6 +198,9 @@ private void testRolledupMetric(Locator locator, Class fullResClass, Class rollu
             } catch (InterruptedException e) {
             }
         }
+
+        reader.getDataToRoll(rollupClass, locator, range, CassandraModel.CF_METRICS_PREAGGREGATED_5M).getPoints();
```
GeorgeJahad: why is this line needed?

blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/BluefloodEnumRollupTest.java
```diff
@@ -0,0 +1,77 @@
+package com.rackspacecloud.blueflood.types;
```
GeorgeJahad: File header?

blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/BluefloodEnumRollupTest.java
```diff
@@ -0,0 +1,77 @@
...
+            else if (hash == 40L) {
+                Assert.assertEquals(40L, (long)hashCount);
+            }
+            else if (hash == 50L) {
+                Assert.assertEquals(10L, (long)hashCount);
+            }
+        }
+    }
+
+    private static <T> Points<T> asPoints(Class<T> type, long initialTime, long timeDelta, T... values) {
```
GeorgeJahad: You've added multiple copies of this method.   Can it be put in a utility class somewhere?

GeorgeJahad:
+1

PR #522, opened 2015-09-17
Added Query for Enums

goru97:
*What:
Added query feature for Enums

*Why:
So that users can query for Enums

*How:
By writing code


blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
```diff
@@ -265,12 +266,16 @@ public String getLastStringValue(Locator locator) {
...
     // todo: A better interface may be to pass the serializer in instead of the class type.
-    public <T extends Rollup> Points<T> getDataToRoll(Class<T> type, Locator locator, Range range, ColumnFamily<Locator, Long> cf) throws IOException {
+    public <T extends Rollup> Points<T> getDataToRoll(Class<T> type, final Locator locator, Range range, ColumnFamily<Locator, Long> cf) throws IOException {
         AbstractSerializer serializer = NumericSerializer.serializerFor(type);
         // special cases. :( the problem here is that the normal full res serializer returns Number instances instead of
         // SimpleNumber instances.
         // todo: this logic will only become more complicated. It needs to be in its own method and the serializer needs
         // to be known before we ever get to this method (see above comment).
+
+        ExecutorService taskExecutor = Executors.newSingleThreadExecutor();
```
GeorgeJahad: I think it would be better if we only did this when really needed.  Can you move this down to line 292?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
```diff
@@ -301,6 +307,16 @@ public String getLastStringValue(Locator locator) {
             log.error("Problem deserializing data for " + locator + " (" + range + ") from " + cf.getName(), ex);
             throw new IOException(ex);
         }
+
+        if (type.equals(BluefloodEnumRollup.class)) {
+                try {
```
GeorgeJahad: indentation seems off here

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
goru97: Done!

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
goru97: Done!

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
```diff
@@ -430,6 +450,26 @@ private MetricData getStringMetricDataForRange(Locator locator, Range range, Gra
         return new MetricData(points, getUnitString(locator), MetricData.Type.STRING);
     }
 
+    private MetricData getEnumMetricDataForRange(final Locator locator, Range range, Granularity gran, RollupType rollupType, DataType dataType) {
```
GeorgeJahad: Why are rollupType and dataType parameters?  Aren't they always the same for enumMetrics?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
```diff
@@ -301,6 +308,16 @@ public String getLastStringValue(Locator locator) {
...
             throw new IOException(ex);
         }
+
+        if (type.equals(BluefloodEnumRollup.class)) {
+            try {
+                ColumnList<Long> enumvalues = enumValuesFuture.get();
+                taskExecutor.shutdown();
+                return getEnumStringValuesFromHashes(points, enumvalues);
+            } catch (Exception e) {
+                log.error("Exception", e);
```
GeorgeJahad: We need better exception messages.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
GeorgeJahad: Also, the code in this method is doing basically the same as the code above in the getDataToRoll() method.  Is there any way you can merge most of the functionality into the same method?

VinnyQ:
@goru97 I think you need to rebase this PR against the enums-feature branch since we didn't a rebase of that against master a few days ago, and merged 3 other PRs in.  We can work together to resolve the conflicts that you'll most likely have with the rebase when you have a chance.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
goru97: Hey George! I've made another callable to get the points so as to remove redundancy.
Yes, type and rollup type are two different things.
Also I've repaired the exceptions.

tilogaat: I am not sure whether we need the hashes 2 string in getDataToRoll. I am looking into that currently. I think the method is used only for rollups.

tilogaat:
closed in favor of #527.

PR #524, opened 2015-09-17
fix failing tests in inputs.handlers

VinnyQ:
WHAT:  fix failing tests after ES upgrades to 1.7.1

HOW:  tests are broken because of invalid mappings and setups, which before was ok but now is not.

### DIFF
--- a/blueflood-elasticsearch/pom.xml
+++ b/blueflood-elasticsearch/pom.xml
@@ -75,7 +75,7 @@
     <dependency>
       <groupId>com.github.tlrx</groupId>
       <artifactId>elasticsearch-test</artifactId>
-      <version>1.1.0</version>
+      <version>1.2.1</version>
       <scope>test</scope>
       <optional>true</optional>
     </dependency>
--- a/blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/inputs/handlers/HttpAnnotationsEndToEndTest.java
+++ b/blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/inputs/handlers/HttpAnnotationsEndToEndTest.java
@@ -73,7 +73,7 @@ public class HttpAnnotationsEndToEndTest {
         esSetup.execute(EsSetup.deleteAll());
         esSetup.execute(EsSetup.createIndex(EventElasticSearchIO.EVENT_INDEX)
                 .withSettings(EsSetup.fromClassPath("index_settings.json"))
-                .withMapping("metrics", EsSetup.fromClassPath("events_mapping.json")));
+                .withMapping("graphite_event", EsSetup.fromClassPath("events_mapping.json")));
         eventsSearchIO = new EventElasticSearchIO(esSetup.client());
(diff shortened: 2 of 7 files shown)


VinnyQ:
@GeorgeJahad @tilogaat please review and merge.  You will need this to fix the failing tests from us upgrading to ES 1.7.1

blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/inputs/handlers/HttpAnnotationsEndToEndTest.java
```diff
@@ -73,7 +73,7 @@ public static void setUp() throws Exception {
         esSetup.execute(EsSetup.deleteAll());
         esSetup.execute(EsSetup.createIndex(EventElasticSearchIO.EVENT_INDEX)
                 .withSettings(EsSetup.fromClassPath("index_settings.json"))
-                .withMapping("metrics", EsSetup.fromClassPath("events_mapping.json")));
+                .withMapping("graphite_event", EsSetup.fromClassPath("events_mapping.json")));
         eventsSearchIO = new EventElasticSearchIO(esSetup.client());
```
VinnyQ: Note here that the document type to retrieve from events_mapping.json is actually called "graphite_event", not "metrics" 
(see https://github.com/rackerlabs/blueflood/blob/master/blueflood-elasticsearch/src/main/resources/events_mapping.json).

This was causing weird errors to occur later on, not at the time of esSetup execute for creating the index, because of the flexible nature of elastic search.

blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/inputs/handlers/HttpAnnotationsEndToEndTest.java
```diff
@@ -153,7 +153,12 @@ private URI getAnnotationsQueryURI() throws URISyntaxException {
     @AfterClass
     public static void tearDownClass() throws Exception{
         Configuration.getInstance().setProperty(CoreConfig.EVENTS_MODULES.name(), "");
-        esSetup.terminate();
-        httpQueryService.stopService();
+        if (esSetup != null) {
```
VinnyQ: I added the null check for these tears down methods because it was masking the real error when something failed during the test setups and these clients never got instantiate and configured, and when it failed and called the tear down it then tries to call the terminate or stopService methods etc which resulted in a NullPointerException, and the error that was getting as the cause of the test failures.

blueflood-elasticsearch/pom.xml
```diff
@@ -75,7 +75,7 @@
     <dependency>
       <groupId>com.github.tlrx</groupId>
       <artifactId>elasticsearch-test</artifactId>
-      <version>1.1.0</version>
+      <version>1.2.1</version>
```
VinnyQ: I decided to just upgraded the elasticsearch-test lib to version 1.2.1 (latest stable) here just because it's good to use the latest stable version, but it's not strictly necessary for this fix.

tilogaat:
+1

PR #526, opened 2015-09-18
Use excessEnumsTable to check enums at ingest time 

GeorgeJahad:
why:  To prevent bad metrics from being ingested.

how: 
1. Create in-memory excess enum metrics table which gets periodically read from Cassandra.
2. Use that table to check incoming metrics.
3. Fixed AstyanaxReader.getExcessEnumMetrics to return a set instead of a list.


blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/PreaggregateConversions.java
```diff
@@ -129,10 +130,12 @@
...
-            BluefloodEnumRollup rollup = new BluefloodEnumRollup();
-            rollup = rollup.withEnumValue(en.getValue(), 1L);
-            PreaggregatedMetric metric = new PreaggregatedMetric(timestamp, locator, DEFAULT_TTL, rollup);
-            list.add(metric);
+            if (!ExcessEnumReader.getInstance().isInExcessEnumMetrics(locator)) {
+                BluefloodEnumRollup rollup = new BluefloodEnumRollup();
+                rollup = rollup.withEnumValue(en.getValue(), 1L);
+                PreaggregatedMetric metric = new PreaggregatedMetric(timestamp, locator, DEFAULT_TTL, rollup);
+                list.add(metric);
+            }
```
VinnyQ: Where are we actually returning an error to the user and/or writing errors to a log that the metrics ingestions were rejected because it had reached the bad metrics count (determined by the isInExcessEnumMetrics call)?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/BluefloodServiceStarter.java
```diff
@@ -125,6 +125,10 @@ private static void startIngestServices(ScheduleContext context) {
                     System.exit(1);
                 }
             }
+
+            final Thread eerThread = new Thread(ExcessEnumReader.getInstance(), "Excess Enum Table Reader");
+            eerThread.start();
```
VinnyQ: Let's add another log.info here that the ExcessEnumRead thread was started so we can note it in the log like when other ingestion module was started in the loop before it.  We also might or might not increment services_started counter by 1, depending if you think if it should be consider part of the "ingestion services" started by this method.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
```diff
@@ -530,8 +530,8 @@ private Points getPointsFromColumns(ColumnList<Long> columnList, RollupType roll
      * Gets all the ExcessEnum Metrics Locators
      *
      */
-    public List<Locator> getExcessEnumMetrics() throws Exception{
-        final ArrayList<Locator> excessEnumMetrics = new ArrayList<Locator>();
+    public Set<Locator> getExcessEnumMetrics() throws Exception{
+        final Set<Locator> excessEnumMetrics = new HashSet<Locator>();
         Function<Row<Locator, Long>, Boolean> rowFunction = new Function<Row<Locator, Long>, Boolean>() {
             @Override
             public Boolean apply(Row<Locator, Long> row) {
```
VinnyQ: When was this method added?  A previous story?

VinnyQ:
You have a failing test in ExcessEnumsReaderIntegrationTest.  Need fixing.

blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/service/ExcessEnumsReaderIntegrationTest.java
```diff
@@ -0,0 +1,53 @@
...
+        Assert.assertFalse(ExcessEnumReader.getInstance().isInExcessEnumMetrics(dummyLocator));
+        // Start the thread to read the table from Cassandra
+        eerThread.start();
+        Thread.sleep(100);
+        // Confirm that the locator has been read from Cassandra
+        Assert.assertTrue(ExcessEnumReader.getInstance().isInExcessEnumMetrics(dummyLocator));
+
+    }
+}
+
```
VinnyQ: A good practice I've learned and try to do often is that when using assert statements in tests, if it's not immediately obvious from the value or test name why the assert failed, and what it was *really* testing, then I would actually add a more precise failure message to the assert.  e.g. for this assert, I would change it to: 

```
Assert.assertTrue("isInExcessEnumMetrics for dummyLocator should return true", ExcessEnumReader.getInstance().isInExcessEnumMetrics(dummyLocator));
```

The idea is to make it easy when a tests failed to be able to find where it failed (especially when a test has multiple assert statements) and a brief message/details to hint at what it was trying to test for.  Apply this principle to other asserts as you see fit.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
VinnyQ: When was this method added? A previous story?

blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/service/ExcessEnumsReaderIntegrationTest.java
GeorgeJahad: will do.  Thanks.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/PreaggregateConversions.java
GeorgeJahad: We haven't yet defined how we are going to expose those to ELE.  I don't think they want us to return errors.  We'll need to talk to them about it.  I've added a jira story to track it.  In the mean time I'll add a log mesg.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/BluefloodServiceStarter.java
GeorgeJahad: Good suggestions.  Will do.

GeorgeJahad:
@VinnyQ I can't reproduce the test failure.  how exactly is it failing for you?  and how are you running the integration tests?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
GeorgeJahad: yes

VinnyQ:
@GeorgeJahad I generally run "mvn clean test integration-test" as shown in the README.  Looks fine now though, no more failing tests.  Must have been something weird on my laptop.

VinnyQ:
:+1: lgtm @GeorgeJahad

PR #527, opened 2015-09-25
Adding queries for enums

tilogaat:
What:
Adding query features for enums. The query and result that will be supported by this PR are:
```
curl -XGET 'localhost:20000/v2.0/99988877/views/call_xyz_api?from=1443231812000&to=1443231814000&points=200'

{
  "unit": "unknown",
  "values": [
    {
      "numPoints": 2,
      "timestamp": 1444071734905,
      "enum_values": {
        "enumValueAhxeB": 1,
        "enumValueIehOS": 1
      },
      "type": "ENUM"
    }
  ],
  "metadata": {
    "limit": null,
    "next_href": null,
    "count": 1,
    "marker": null
  }
}
```

How:
The crux of this PR is in AstyanaxReader. All others are wiring changes. 
metrics_enum CF holds the raw string values mapped to the hashcodes. In rollups, we only use hashcodes for rollups. At the time of the query, we need to map these hashcodes with the actual raw strings. In AstyanaxReader a threadpool will retrieve enum values from metrics_enum CF, simultaneously it will retrieve the actual points from metrics_pre-aggregated_full (or the corresponding CFs for rollups). It will create a relational join using the string and hash values. The end result for the query will be obtained from stringEnumValues2Count datastructure in BluefloodEnumRollup class.


blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
```diff
@@ -559,4 +612,91 @@ public Boolean apply(Row<Locator, Long> row) {
...
+                System.out.println("CF Not found");
+            } else {
+                System.out.println("Some other error");
+            }
+        } finally {
+        }
+        return columns.get(locator);
+    }
+
+    class ColumnFromEnumCF implements Callable{
```
VinnyQ: I am not a fan of inner classes.  It makes the top AstyanaxReader class very bloated, and hard to reuse in the future (as you can see from all the class extraction refactoring we have had to do recently).  Any reason why this and the PointsFromDB class below can't be standalone classes in the com.rackspacecloud.blueflood.io package?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
```diff
@@ -559,4 +612,91 @@ public Boolean apply(Row<Locator, Long> row) {
         return excessEnumMetrics;
     }
 
+    private <T extends Rollup> Points<T> getEnumStringValuesFromHashes(Points<T> points, ColumnList<Long> enumvalues) {
```
VinnyQ: Let's made these and the two other function below public.  They're usable for other stories.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
```diff
@@ -559,4 +612,91 @@ public Boolean apply(Row<Locator, Long> row) {
...
+        ColumnFromEnumCF(Locator locator){
+            this.locator = locator;
+        }
+        @Override
+        public ColumnList<Long> call() throws ScriptException {
+            return getColumnsFromEnumCF(locator);
+        }
+    }
+
+    class PointsFromDB implements Callable{
```
VinnyQ: Should extract this inner class to be standalone.  See above comment for class ColumnFromEnumCF.

blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/inputs/handlers/HttpHandlerIntegrationTest.java
```diff
@@ -305,7 +305,7 @@ public void testHttpAggregatedMultiIngestion_WithMultipleEnumPoints() throws Exc
         verify(context, atLeastOnce()).update(anyLong(), anyInt());
 
         final Locator locator2 = Locator.createLocatorFromPathComponents("99988877", "call_xyz_api");
-        Points<BluefloodEnumRollup> points2 = new FakeEnumAstyanaxReader().getDataToRoll(BluefloodEnumRollup.class,
+        Points<BluefloodEnumRollup> points2 = AstyanaxReader.getInstance().getDataToRoll(BluefloodEnumRollup.class,
                 locator2, new Range(1439231323000L, 1439231325000L), CassandraModel.getColumnFamily(BluefloodEnumRollup.class, Granularity.FULL));
         Assert.assertEquals(2, points2.getPoints().size());
```
VinnyQ: We definitely need more integration tests to test out the new query for enum values features implemented in this story.

VinnyQ:
Can we update the PR overview with a bit more details?  e.g. what are the end points that this story is for?  Example of queries and expected results would be nice.  And a general overview of how this will be accomplished (e.g. adding retrieval methods in AstyanaxReader that retrieve hash from x column family and values from y column family and/or deserializing?)

Thanks!

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
```diff
@@ -342,6 +346,9 @@ public MetricData getDatapointsForRange(Locator locator, Range range, Granularit
             if (rollupType == null) {
                 rollupType = RollupType.BF_BASIC;
             }
+            if (type == null && rollupType == RollupType.ENUM){
+                return getEnumMetricDataForRange(locator, range, gran);
```
GeorgeJahad: Is there ever a time when "rollupType" is ENUM and you don't want to call getEnumMetricDataForRange()?  If not, why are you testing "type" for null?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
```diff
@@ -430,6 +453,56 @@ private MetricData getStringMetricDataForRange(Locator locator, Range range, Gra
...
+    public MetricData getEnumMetricDataForRange(final Locator locator, final Range range, Granularity gran) {
+        Map<Locator, MetricData> metricDataMap = getEnumMetricDataForRangeForLocatorList(new ArrayList<Locator>(){{add (locator); }}, range, gran);
+        return metricDataMap.get(locator);
+    }
+
+    public  Map<Locator, MetricData> getEnumMetricDataForRangeForLocatorList(final List<Locator> locator, final Range range, final Granularity gran) {
+        int enumThreadCount = Configuration.getInstance().getIntegerProperty(CoreConfig.ENUM_READ_THREADS);
+        final ColumnFamily<Locator, Long> CF = CassandraModel.getColumnFamily(RollupType.classOf(RollupType.ENUM, gran), gran);
+
+        ExecutorService taskExecutor = new ThreadPoolBuilder().withUnboundedQueue()
```
GeorgeJahad: I'm not sure you want to create a new ThreadPool each time this is called.  Isn't the purpose of a threadpool to have a set of threads warmed up and ready to go?

GeorgeJahad: Also you may want to use an InstrumentedThreadPoolExecutor, so that we get some builtin codahale metrics

blueflood-core/src/main/java/com/rackspacecloud/blueflood/types/BluefloodEnumRollup.java
```diff
@@ -41,6 +41,16 @@ public int getCount() {
         return hashedEnumValues2Count.size();
     }
 
+    public int getNumPoints() {
+        int enumCount = 0;
+        Set keySet = hashedEnumValues2Count.keySet();
+        Iterator itr = keySet.iterator();
+        while(itr.hasNext()){
+            enumCount+= hashedEnumValues2Count.get(itr.next());
+        }
```
GeorgeJahad: wouldn't it be clearer to iterate through the values with the foreach syntax, something like this:
for (Long value: hashedEnumValues2Count.values()):
    enumCount += value;

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
```diff
@@ -559,4 +632,53 @@ public Boolean apply(Row<Locator, Long> row) {
...
+            for (Long hash : enumRollup.getHashedEnumValuesWithCounts().keySet()) {
+                String enumValueString = hash2enumValues.get(hash);
+                enumRollup.getStringEnumValuesWithCounts().put(enumValueString, enumRollup.getHashedEnumValuesWithCounts().get(hash));
+                pointsEnum.add(new Points.Point<T>(timestamp,(T)enumRollup));
+            }
+        }
+        return pointsEnum;
+    }
+
+    private Map<Long, String> getEnumValueFromHash(ColumnList<Long> enumValues) {
```
GeorgeJahad: NIT:  Would it be better to make this the public method and take a list of locators, and call getColumnsFromEnumCF() internally?

No one really wants the ColumnList do they?  They just want the Map<Long, String> ?

blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/io/AstyanaxReaderIntegrationTest.java
```diff
@@ -89,6 +89,111 @@ public void testCanReadExcessEnumMetrics() throws Exception {
...
+                BluefloodEnumRollup er = actualData.get(timestamp).getData();
+                Assert.assertEquals(2, er.getStringEnumValuesWithCounts().size());
+                for (String value : er.getStringEnumValuesWithCounts().keySet()) {
+                    Assert.assertTrue(value.contains("enumValue"));
+                    Assert.assertEquals(1L, (long) er.getStringEnumValuesWithCounts().get(value));
+                }
+            }
+        }
+    }
+
```
GeorgeJahad: NIT: it seems like there is a lot of redundant code  in these 3 methrods that you could factor out into parameterized methods.

blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/inputs/handlers/HttpEnumEndToEndTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+    private URI getEnumQueryURI(String metricName, Map<String, String> parameterMap) throws URISyntaxException {
+        URIBuilder builder = new URIBuilder().setScheme("http").setHost("127.0.0.1")
+                .setPort(queryPort).setPath("/v2.0/" + tenant_id + "/views/"+metricName);
+
+        Set<String> parameters = parameterMap.keySet();
+        Iterator<String> setIterator = parameters.iterator();
+        while (setIterator.hasNext()){
+            String paramName = setIterator.next();
+            builder.setParameter(paramName, parameterMap.get(paramName));
```
GeorgeJahad: i prefer the foreach idiom:
for(String paramName: parameters)
instead of using iterators

blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/inputs/handlers/HttpEnumEndToEndTest.java
```diff
@@ -0,0 +1,103 @@
...
+        parameterMap.put("to", String.valueOf(metric1.getCollectionTime() + 10000));
+        parameterMap.put("points", "200");
+
+        HttpGet get = new HttpGet(getEnumQueryURI("enum_metric2", parameterMap));
+        HttpResponse response = client.execute(get);
+
+        String responseString = EntityUtils.toString(response.getEntity());
+        Assert.assertEquals(200, response.getStatusLine().getStatusCode());
+        Assert.assertTrue(responseString.contains("enumValue"));
+    }
```
GeorgeJahad: I'm not sure if this is the right place, but we need a test that confirms that the data returned contains everything mentioned in the spec here: https://one.rackspace.com/pages/viewpage.action?title=Enum+Metrics&spaceKey=cloudmetrics

For example, the spec talks about the "type" field in the sections on splot and mplot queries.  Is that field being set correctly?

blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/inputs/handlers/HttpHandlerIntegrationTest.java
```diff
@@ -270,16 +270,9 @@ public void testHttpAggregatedMultiIngestionHappyCase() throws Exception {
...
 
         final Locator locator2 = Locator.createLocatorFromPathComponents("5405577", "call_xyz_api");
-        Points<BluefloodEnumRollup> points2 = new FakeEnumAstyanaxReader().getDataToRoll(BluefloodEnumRollup.class,
+        Points<BluefloodEnumRollup> points2 = AstyanaxReader.getInstance().getDataToRoll(BluefloodEnumRollup.class,
                 locator2, new Range(1439231323000L, 1439231325000L), CassandraModel.getColumnFamily(BluefloodEnumRollup.class, Granularity.FULL));
         Assert.assertEquals(1, points2.getPoints().size());
-        for (Long timestamp : points2.getPoints().keySet()) {
-            BluefloodEnumRollup er = points2.getPoints().get(timestamp).getData();
-            for (String value : er.getStringEnumValuesWithCounts().keySet()) {
-                Assert.assertEquals("OK",value);
```
GeorgeJahad: Did these asserts get moved?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
GeorgeJahad: Also can we give it a better name, something like getEnumStringsFromColumnList()?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
tilogaat: good catch @GeorgeJahad Will fix.

blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/inputs/handlers/HttpHandlerIntegrationTest.java
tilogaat: Yes they are now in HttpRollupHandlerWithESIntegrationTest.java in 85db11c

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
tilogaat: Actually if you see the build method, the InstrumentedThreadPoolExecutor.instrument is already invoked in it.

blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/inputs/handlers/HttpEnumEndToEndTest.java
tilogaat: Need to fix. I do not think we are adding it yet.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
```diff
@@ -559,4 +630,53 @@ public Boolean apply(Row<Locator, Long> row) {
...
+            for (Long hash : enumRollup.getHashedEnumValuesWithCounts().keySet()) {
+                String enumValueString = hash2enumValues.get(hash);
+                enumRollup.getStringEnumValuesWithCounts().put(enumValueString, enumRollup.getHashedEnumValuesWithCounts().get(hash));
+                pointsEnum.add(new Points.Point<T>(timestamp,(T)enumRollup));
+            }
+        }
+        return pointsEnum;
+    }
+
+    private Map<Long, String> getEnumValueFromHashes(ColumnList<Long> enumValues) {
```
VinnyQ: This method is useful and should be made public.  I am going to need it in my PR.  And what happened to the "getColumnsFromEnumCF" method?  I am going to need that also.  Or should I just add that on my own?

VinnyQ: Oh nevermind about the getColumnsFromEnumCF() comment.  I see that you renaamed it to getStringAndHashMappingForEnums() and refactored it to pull for a list of locators for optimization.

I can modify my code to use the new method in my PR once I rebased.

VinnyQ:
:+1:  LGTM!  Great tests!  Good new methods.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
```diff
@@ -58,8 +61,15 @@
...
+    private static int enumThreadCount = Configuration.getInstance().getIntegerProperty(CoreConfig.ENUM_READ_THREADS);
+    private ExecutorService taskExecutor =  null;
 
     public static AstyanaxReader getInstance() {
+        if (INSTANCE.taskExecutor == null || INSTANCE.taskExecutor.isShutdown()) {
+            INSTANCE.taskExecutor = new ThreadPoolBuilder().withUnboundedQueue()
+                    .withCorePoolSize(enumThreadCount)
+                    .withMaxPoolSize(enumThreadCount).withName("Retrieving Enum Values").build();
+        }
         return INSTANCE;
```
VinnyQ: I like how you made the threadpool a class instance here.  We're definite we'll get only on instance of taskExecutor across all AtyanaxReader instances?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
```diff
@@ -559,4 +630,53 @@ public Boolean apply(Row<Locator, Long> row) {
...
+    }
+
+    public Map<Locator, ColumnList<Long>> getStringAndHashMappingForEnums(final List<Locator> locators) {
+        final Map<Locator, ColumnList<Long>> columns = new HashMap<Locator, ColumnList<Long>>();
+
+        try {
+            OperationResult<Rows<Locator, Long>> query = getKeyspace()
+                    .prepareQuery(CassandraModel.CF_METRICS_ENUM)
+                    .getKeySlice(locators)
+                    .execute();
```
VinnyQ: Does the above query to cassandra only return locators (with the .getKeySlice filter) that has enum values in the CF_METRICS_ENUM table?  Or does it retain the original locators list even if no enum values are found for some of the locators?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
```diff
@@ -559,4 +630,53 @@ public Boolean apply(Row<Locator, Long> row) {
...
+        HashMap<Long,String> hash2enumValues = new HashMap<Long, String>();
+
+        for (Column<Long> col: enumValues) {
+            hash2enumValues.put(col.getName(), col.getStringValue());
+        }
+
+        return hash2enumValues;
+    }
+
+    public Map<Locator, ColumnList<Long>> getStringAndHashMappingForEnums(final List<Locator> locators) {
```
VinnyQ: NIT:  Looking at this method, it looks like you're only get querying for HashMapping for the locators passed in.  I am not sure what the "String" part in the method name "getStringAndHashMappingForEnums" refer to.

I would just go with "getEnumHashMappings(final List<Locator> locators)", but no biggie if you disagree.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
tilogaat: hmm. Good point I will add a test for that and see how it behaves.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
tilogaat: Agreed. will fix.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
tilogaat: @VinnyQ do you think it will be more intuitive if it's a static variable? I did not understand why you think it should be a static.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
tilogaat: @VinnyQ Ok i will make it public.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
tilogaat: added in 7f7e003

PR #528, opened 2015-09-29
George/fix dogfooding

GeorgeJahad:
The dogfooded graphs weren't matching the standard Graphite versions because:
1. the full res granularity was different.
2. Graphite interpolates Null data points, (which makes NonNegative Derivative work better.)

This PR addresses both those issues.

However it is still not perfect, and there are still some discrepancies between the two that will require investigating


tilogaat:
+1

PR #529, opened 2015-09-30
Add enums to the finder

GeorgeJahad:
## INTRO:
There are two parts to the finder api:
1. find_metrics() which returns a list of metric names based on a glob pattern
2. fetch_multi() with returns the data for a list of metrics

Both support submetrics_aliases.  Without submetric_aliases configured in, the data returned uses the default submetric, which is usually "average".  If the user wants to use something other than the default submetric, he creates an alias for that submetric in the finder config.  Then he appends the alias to the end of the metric name in the grafana dashboard.  So for example if he wants to use the "max" submetric for metric "x.y", he creates an alias like "_max" in the config and specifies the metric name "x.y._max" in the dashboard.

Note: Currently submetrics have to be entirely enabled or disabled in grafana server.  If they are enabled then every metric in the dashboard needs to be appended with a submetric alias, or they wont be found.

## CHANGES IN THIS PR:
Adding enums effects both find_metrics() and fetch_multi().

In find_metrics(), without submetrics, it has to do two reads.  For example, if the pattern requested is "x.y", then there are two possibilities, either "y" is an enum_value or not.  So find metric has to read both "x" and "x.y".   It does the enum read, "x", in a separate thread in an attempt to reduce the IO time.

In find_metrics(), with submetrics, the assumption is that the metric name of an enum metric will end with a submetric alias for ".enum".  So only one read is necessary.

In fetch_multi(), each enum value is treated as a separate metric, (which allows Grafana to graph them separately.)  So now the code needs to massage the incoming data to merge the values with the metric names, and create separate metrics for each enum_value.

### DIFF
--- a/contrib/graphite/blueflood.py
+++ b/contrib/graphite/blueflood.py
@@ -10,6 +10,10 @@ import traceback
 import os.path
 import fnmatch
 
+import Queue
+import threading
+
+
 try:
     from graphite_api.intervals import Interval, IntervalSet
     from graphite_api.node import LeafNode, BranchNode
@@ -49,10 +53,11 @@ def calc_res(start, stop):
     res = 'MIN1440'
   return res
 
-class TenantBluefloodFinder(object):
+class TenantBluefloodFinder(threading.Thread):
   __fetch_multi__ = 'tenant_blueflood'
   def __init__(self, config=None):
-    print("Blueflood Finder v28")
+    print("Blueflood Finder v31")
+    threading.Thread.__init__(self)
     if os.path.isfile("/root/pdb-flag"):
       import remote_pdb
       remote_pdb.RemotePdb('127.0.0.1', 4444).set_trace()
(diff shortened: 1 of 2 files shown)


VinnyQ:
:+1: lgtm.  Nice tests.  

Maybe you can squash some of the smaller commits together with into several big commits, for the sake of the project history.

$ git rebase -i HEAD~23

Let me know if you would like to do that and unfamiliar with rebase squashing.  It's pretty straight forward once you know the general flow and commands

contrib/graphite/blueflood.py
```diff
@@ -79,24 +84,34 @@ def __init__(self, config=None):
...
+  def complete(self, metric, complete_len):
+    #returns true if metric is a complete metric name wrt the query
     metric_parts = metric.split('.')
-    if self.enable_submetrics:
-      complete_len = query_depth - 1
-    else:
-      complete_len = query_depth 
     return (len(metric_parts) == complete_len)
 
-
```
tilogaat: nit: do we need these blank spaces?

contrib/graphite/blueflood.py
```diff
@@ -107,19 +122,30 @@ def make_request(self, url, payload, headers):
     return r
 
   def find_metrics_endpoint(self, endpoint, tenant):
-    return "%s/v2.0/%s/metrics/search" % (endpoint, tenant)
+    return "%s/v2.0/%s/metrics/search?include_enum_values=true" % (endpoint, tenant)
```
tilogaat: will this endpoint be used to find all metrics? It's a little awkward to invoke include_enum_values for querying all metrics, assuming that all values are enums.

tilogaat:
+1

PR #530, opened 2015-10-01
Check for bad enum metrics during rollup

VinnyQ:
WHAT:  "Check for bad enum metrics" story (https://jira.rax.io/browse/CMD-749)

Create a threadpool of threads, which can be kicked off by RollupService everytime it is activated, that will run in parallel to validate enum metrics and check for bad metrics.  Enum metrics are consider "bad" when the number of unique enum values ingested for the metric exceeded a certain configurable threshold (e.g. 100).  

The thread will also write new enums indexes to elastic search, which will contains all enum values for the metric -- if it is not a bad metric.  This allows users to query blueflood with the include_enum_values param and retrieve metric names along with all enum values in the result.

WHY:  prevent blueflood from ingesting enum metrics with too many unique values, which isn't useful and an incorrect usage of enum metrics.  Also allow users to see a list of all enum values ingested for  metrics by creating the "enums" indexes in elasticsearch.

HOW: 
Add thread EnumValidator, which valid enum metrics and
- mark metrics as bad if exceeds threshold, or
- write elastic search enums index of values for querying.

This is done via a runnable thread EnumValidator that is stared by the RollupService, for any batch of metrics that are involved in the rollup.  The EnumValidator threads process the metrics in parallel to rollup, and does not depend on data from the rollup, except for the locators of the metrics affected.

EnumValidator threads are manage by a threadpool executor enumValidatorExecutor, added to the RollupService.

### DIFF
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
@@ -679,4 +679,32 @@ public class AstyanaxReader extends AstyanaxIO {
         }
         return columns;
     }
+
+    public Map<Locator, ArrayList<String>> getEnumStringMappings(final List<Locator> locators) {
+        final Map<Locator, ArrayList<String>> map = new HashMap<Locator, ArrayList<String>>();
+
+        try {
+            OperationResult<Rows<Locator, Long>> query = getKeyspace()
+                    .prepareQuery(CassandraModel.CF_METRICS_ENUM)
+                    .getKeySlice(locators)
+                    .execute();
+
+            for (Row<Locator, Long> row : query.getResult()) {
+                ColumnList<Long> cols = row.getColumns();
+                ArrayList<String> enumStrings = new ArrayList<String>();
+                for (Column col : cols) {
+                    enumStrings.add(col.getStringValue());
+                }
+                map.put(row.getKey(), enumStrings);
+            }
+        } catch (ConnectionException e) {
+            if (e instanceof NotFoundException) { // TODO: Not really sure what happens when one of the keys is not found.
+                Instrumentation.markNotFound(CassandraModel.CF_METRICS_ENUM);
+            } else {
+                log.warn("Enum String read query failed for column family " + CassandraModel.CF_METRICS_ENUM.getName(), e);
+                Instrumentation.markReadError(e);
+            }
+        }
+        return map;
+    }
 }
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxWriter.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxWriter.java
@@ -209,10 +209,10 @@ public class AstyanaxWriter extends AstyanaxIO {
         }
     }
 
-    public void writeExcessEnumMetric(Locator locator, Long timestamp) throws ConnectionException {
+    public void writeExcessEnumMetric(Locator locator) throws ConnectionException {
         Timer.Context ctx = Instrumentation.getWriteTimerContext(CassandraModel.CF_METRICS_EXCESS_ENUMS);
         try {
-            keyspace.prepareColumnMutation(CassandraModel.CF_METRICS_EXCESS_ENUMS, locator, timestamp)
+            keyspace.prepareColumnMutation(CassandraModel.CF_METRICS_EXCESS_ENUMS, locator, 0L)
                     .putEmptyColumn(null).execute();
         } catch (ConnectionException e) {
             Instrumentation.markWriteError(e);
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/EnumValidator.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/EnumValidator.java
@@ -1,0 +1,167 @@
+/*
+ * Copyright 2015 Rackspace
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+package com.rackspacecloud.blueflood.service;
+
+import com.netflix.astyanax.connectionpool.exceptions.ConnectionException;
+import com.netflix.astyanax.model.ColumnList;
+import com.rackspacecloud.blueflood.concurrent.ThreadPoolBuilder;
+import com.rackspacecloud.blueflood.io.AstyanaxReader;
+import com.rackspacecloud.blueflood.io.AstyanaxWriter;
+import com.rackspacecloud.blueflood.io.DiscoveryIO;
+import com.rackspacecloud.blueflood.io.SearchResult;
+import com.rackspacecloud.blueflood.types.BluefloodEnumRollup;
+import com.rackspacecloud.blueflood.types.IMetric;
+import com.rackspacecloud.blueflood.types.Locator;
+import com.rackspacecloud.blueflood.types.PreaggregatedMetric;
+import com.rackspacecloud.blueflood.utils.ModuleLoader;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.*;
+import java.util.concurrent.*;
+
+/** This class handles collecting enum values for specific metrics, via their locators, and checking whether
+    a uniqueness count of the enum values have reached a certain threshold.  If it has, the class mark the metric
+    as bad by inserting its locator into the metrics_excess_enums table.  If it hasn't reached the threshold, then the class
+    will create/update the elastic search "enums" index for the metric, which will contain a list of all unique
+    enum values that has been ingested for it.
+**/
+public class EnumValidator implements Runnable {
+
+    private static final Logger log = LoggerFactory.getLogger(EnumValidator.class);
+    private static final Configuration config = Configuration.getInstance();
+    private static final DiscoveryIO discoveryIO = (DiscoveryIO) ModuleLoader.getInstance(DiscoveryIO.class, CoreConfig.ENUMS_DISCOVERY_MODULES);
+    private static final int ENUM_UNIQUE_VALUES_THRESHOLD = config.getIntegerProperty(CoreConfig.ENUM_UNIQUE_VALUES_THRESHOLD);
+    private static final int ENUM_VALIDATOR_THREADS = Configuration.getInstance().getIntegerProperty(CoreConfig.ENUM_VALIDATOR_THREADS);
+
+    HashSet<Locator> locators;
+
+    public EnumValidator(Set<Locator> locators) {
+        this.locators = (HashSet<Locator>) locators;
+    }
+
+    @Override
+    public void run() {
+
+        ExecutorService taskExecutor = new ThreadPoolBuilder().withUnboundedQueue()
+                .withCorePoolSize(ENUM_VALIDATOR_THREADS)
+                .withMaxPoolSize(ENUM_VALIDATOR_THREADS)
+                .withName("Enum Validator Task Executor")
+                .build();
+
+        for (final Locator locator : locators) {
+            // start async read for enum values of locator from cassandra
+            Future<ColumnList<Long>> enumHashValuesFuture = taskExecutor.submit(new Callable() {
+                @Override
+                public ColumnList<Long> call() throws Exception {
+                    return AstyanaxReader.getInstance().getColumnsFromEnumCF(locator);
+                }
+
+            });
+
+            // start async read for enum values of locator from elastic search
+            Future<List<SearchResult>> esEnumValuesFuture = taskExecutor.submit(new Callable() {
+                @Override
+                public List<SearchResult> call() throws Exception {
+                    return discoveryIO.search(locator.getTenantId(), locator.getMetricName());
+                }
+            });
+
+
+            Map<Long, String> enumStringValues = null;
+            List<SearchResult> esEnumValues = null;
+            try {
+                // values from cassandra db
+                ColumnList<Long> enumHashValues = enumHashValuesFuture.get();
+                if ((enumHashValues != null) && (enumHashValues.size() > 0)) {
+                    enumStringValues = AstyanaxReader.getInstance().getEnumValueFromHash(enumHashValues);
+                    log.debug(String.format("Enum values for metric %s: %s", locator.toString(), enumStringValues.toString()));
+                }
+                else {
+                    log.debug(String.format("No enum values found for metric %s", locator.toString()));
+                }
+
+                // values from elasticsearch
+                esEnumValues = esEnumValuesFuture.get();
+
+                validateThresholdAndIndex(locator, enumStringValues, esEnumValues);
+
+            } catch (InterruptedException e) {
+                log.error("Interrupted Exception during query of Enum metrics in EnumValidator", e);
+            } catch (ExecutionException e) {
+                log.error("Execution Exception during query of Enum metrics in EnumValidator", e);
+            }
+            finally {
+                taskExecutor.shutdown();
+            }
+        }
+    }
+
+    private void validateThresholdAndIndex(Locator locator, Map<Long, String> enumStringValues, List<SearchResult> esEnumValues) {
+
+        if ((enumStringValues != null) && (enumStringValues.size() > ENUM_UNIQUE_VALUES_THRESHOLD)) {
+            // if values_from_cassandra.count > threshold,
+            // write locator to bad metric table
+            try {
+                AstyanaxWriter.getInstance().writeExcessEnumMetric(locator);
+            } catch (ConnectionException e) {
+                log.error("Failed to write bad metric {} to table; {}", locator.toString(), e.getMessage());
+            }
+        }
+        else {
+            // create or update index of metric with enum values if cassandra and elasticsearch values are different
+            // get arraylist of cassandra enum values
+            ArrayList<String> dbValues = null;
+            if ((enumStringValues != null) && (enumStringValues.size() > 0)) {
+                dbValues = new ArrayList<String>(enumStringValues.values());
+            }
+
+            // get arraylist of elasticsearch enum values
+            ArrayList<String> esValues = null;
+            if ((esEnumValues != null) && (esEnumValues.get(0) != null)) {
+                esValues = esEnumValues.get(0).getEnumValues();
+            }
+
+            // sort the two array lists of enum values
+            if (dbValues != null) Collections.sort(dbValues);
+            if (esValues != null) Collections.sort(esValues);
+
+            // verify equals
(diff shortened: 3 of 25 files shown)


blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
```diff
@@ -47,8 +48,8 @@
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
-import java.sql.Connection;
 import java.util.*;
+import java.util.concurrent.*;
 
 public class AstyanaxReader extends AstyanaxIO {
```
VinnyQ: Note that changes for this class are copied from the "Adding queries for enums" PR 527:  https://github.com/rackerlabs/blueflood/pull/527, because I need the functionalities added in that story for my own use (i.e. the getColumnsFromEnumCF() and getEnumValueFromHash() methods).  
Once that PR is merged in, I'll rebase and use that PR copy of AstyanaxReader instead.

VinnyQ:
More tests are needed for this PR before it can be merged:  unit tests for EnumValidator, and integration end2end test which ingest enums metrics, run rollup and validator, and query for the metrics with include_enum_values.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxWriter.java
```diff
@@ -209,10 +209,10 @@ public void writeMetadataValue(Locator locator, String metaKey, String metaValue
         }
     }
 
-    public void writeExcessEnumMetric(Locator locator, Long timestamp) throws ConnectionException {
+    public void writeExcessEnumMetric(Locator locator) throws ConnectionException {
```
VinnyQ: George and me decided that a timestamp value is not needed for writing to the "bad" enum metrics cassandra CF (metrics_excess_enums).  So I removed the param here.

VinnyQ:
@GeorgeJahad @tilogaat @goru97 please start reviewing this PR when you have a chance.  Thanks!

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
GeorgeJahad: So it sounds like I can ignore all of the changes in this file, is that correct?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/EnumValidator.java
```diff
@@ -0,0 +1,167 @@
...
+
+    @Override
+    public void run() {
+
+        ExecutorService taskExecutor = new ThreadPoolBuilder().withUnboundedQueue()
+                .withCorePoolSize(ENUM_VALIDATOR_THREADS)
+                .withMaxPoolSize(ENUM_VALIDATOR_THREADS)
+                .withName("Enum Validator Task Executor")
+                .build();
+
```
GeorgeJahad: I think it may be very expensive to create a new threadpool each time this is run.  Besides the thread that is running is already using a threadpool, (the one created in rollupservice, right?)  Could we reuse that one somehow?  For example, in the constructor above you pass in the locators.  you could also pass in the same threadpool from the rollupservice and use that one here.  What do you think?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/EnumValidator.java
```diff
@@ -0,0 +1,167 @@
...
+            Future<ColumnList<Long>> enumHashValuesFuture = taskExecutor.submit(new Callable() {
+                @Override
+                public ColumnList<Long> call() throws Exception {
+                    return AstyanaxReader.getInstance().getColumnsFromEnumCF(locator);
+                }
+
+            });
+
+            // start async read for enum values of locator from elastic search
+            Future<List<SearchResult>> esEnumValuesFuture = taskExecutor.submit(new Callable() {
```
GeorgeJahad: Since the current thread is already async, (running in a thread pool,)  both of these doesn't need to be async.  One can just run in the current thread, can't it?  I would make the ES call a future and just run the cassandra one in this thread

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/EnumValidator.java
```diff
@@ -0,0 +1,167 @@
...
+                // values from elasticsearch
+                esEnumValues = esEnumValuesFuture.get();
+
+                validateThresholdAndIndex(locator, enumStringValues, esEnumValues);
+
+            } catch (InterruptedException e) {
+                log.error("Interrupted Exception during query of Enum metrics in EnumValidator", e);
+            } catch (ExecutionException e) {
+                log.error("Execution Exception during query of Enum metrics in EnumValidator", e);
+            }
```
GeorgeJahad: Do we want to catch all exceptions here?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/LocatorFetchRunnable.java
```diff
@@ -81,11 +89,21 @@ public void run() {
...
         } catch (RuntimeException e) {
             executionContext.markUnsuccessful(e);
             log.error("Failed reading locators for slot: " + parentSlot, e);
         }
+
+        // start a thread with EnumValidator runnable to validate enum values for this set of locators
+        try {
+            enumValidatorExecutor.execute(new EnumValidator(locators));
+        } catch (Throwable any) {
+            executionContext.markUnsuccessful(any);
```
GeorgeJahad: I don't think we want to mark the executionContext here.  it is just for rollups.  I would remove that line.

And I don't love that we are catching all Throwable's there, (I know the rollup code does, and I'm not sure it is right there either.)  I would just catch all exceptions instead of all throwables.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/EnumValidator.java
```diff
@@ -0,0 +1,167 @@
...
+    public void run() {
+
+        ExecutorService taskExecutor = new ThreadPoolBuilder().withUnboundedQueue()
+                .withCorePoolSize(ENUM_VALIDATOR_THREADS)
+                .withMaxPoolSize(ENUM_VALIDATOR_THREADS)
+                .withName("Enum Validator Task Executor")
+                .build();
+
+        for (final Locator locator : locators) {
+            // start async read for enum values of locator from cassandra
```
GeorgeJahad: Right at the beginning here we need to check if this is an enum and continue if it is not.  We don't want to do any more processing than we have to.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/LocatorFetchRunnable.java
VinnyQ: We'll want to limit this to only the 5m granularity.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/EnumValidator.java
VinnyQ: Good idea, let me try that.  As long as we don't mind they'll share the same thread name.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/EnumValidator.java
VinnyQ: Great idea, let's do this.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
VinnyQ: That is correct, ignore all changes in this AstanaxReader class in this PR.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/RollupService.java
```diff
@@ -179,6 +181,15 @@ protected void afterExecute(Runnable r, Throwable t) {
...
+
+        // create executor threadpool for EnumValidator
+        int enumValidatorThreadCount = Configuration.getInstance().getIntegerProperty(CoreConfig.ENUM_VALIDATOR_THREADS);
+        this.enumValidatorExecutor = new ThreadPoolBuilder()
+                .withName("Validating Enum Metrics")
+                .withCorePoolSize(enumValidatorThreadCount)
+                .withMaxPoolSize(enumValidatorThreadCount)
+                .withUnboundedQueue()
+                .build();
     }
```
GeorgeJahad: It would be good to instrument this threadpool, like on line 182

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/EnumValidator.java
GeorgeJahad: Also, we need to somehow make sure this only runs for the 5 minute rollups. Its a waste to run it for all the others.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/LocatorFetchRunnable.java
```diff
@@ -81,11 +89,21 @@ public void run() {
...
 
         try {
+            // get a list of all locators to rollup for a shard
             locators.addAll(AstyanaxReader.getInstance().getLocatorsToRollup(shard));
         } catch (RuntimeException e) {
             executionContext.markUnsuccessful(e);
             log.error("Failed reading locators for slot: " + parentSlot, e);
         }
+
+        // start a thread with EnumValidator runnable to validate enum values for this set of locators
```
GeorgeJahad: you probably want to check the granularity here and only run if it is 5 minutes

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/EnumValidator.java
VinnyQ: We decided to just make sync read for both the cassandra and ES retrievals, eliminating the taskExecutor threadpool entirely here.  And only read from ES after we verified that the metric is good, and use the results to check whether to update it with the cassandra definitive values.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/EnumValidator.java
VinnyQ: Moved and fixed.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/EnumValidator.java
VinnyQ: Removed executor and thread completely and just do sync reads for these data.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/RollupService.java
VinnyQ: Instrumented!

blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/inputs/handlers/HttpHandlerIntegrationTest.java
```diff
@@ -496,7 +496,7 @@ public static void shutdown() {
             return pointsEnum;
         }
 
-        private Map<Long, String> getEnumValueFromHash(ColumnList<Long> enumValues) {
+        public Map<Long, String> getEnumValueFromHash(ColumnList<Long> enumValues) {
```
tilogaat: I think I got rid of them in this test class

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/EnumValidator.java
```diff
@@ -0,0 +1,146 @@
...
+            }
+
+            // sort the two array lists of enum values
+            if (currentEnumValues != null) Collections.sort(currentEnumValues);
+            if (elasticsearchEnumValues != null) Collections.sort(elasticsearchEnumValues);
+
+            // compare two list of enum values
+            if (((currentEnumValues != null) && (!currentEnumValues.equals(elasticsearchEnumValues))) ||
+                ((currentEnumValues == null) && (elasticsearchEnumValues != null) && (elasticsearchEnumValues.size() > 0)))
+            {
```
tilogaat: At this point, we do not want to create a union of the ```elasticsearchEnumValues``` with the ```currentEnumValues```, do we? We assume that currentEnumValues is the definitive source.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/EnumValidator.java
```diff
@@ -0,0 +1,146 @@
...
+
+    public EnumValidator(Set<Locator> locators) {
+        this.locators = locators;
+    }
+
+    @Override
+    public void run() {
+
+        if (locators == null) return;
+        Map<Locator, ColumnList<Long>> cfMetrics = AstyanaxReader.getInstance().getStringAndHashMappingForEnums(new ArrayList(locators));
```
tilogaat: @VinnyQ would it help if this method directly returns the HashMap with the mapping of Locator to List<Strings>

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/EnumValidator.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    private BluefloodEnumRollup createRollupWithEnumValues(ArrayList<String> enumValues) {
+        BluefloodEnumRollup rollup = new BluefloodEnumRollup();
+        for (String val : enumValues) {
+            rollup = rollup.withEnumValue(val);
+        }
+        return rollup;
+    }
+}
```
tilogaat: @VinnyQ Have we already worked on the feature to reject enums that are in the bad metrics table? We should probably make it a story if not already there.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/EnumValidator.java
```diff
@@ -0,0 +1,146 @@
...
+
+                // convert enum values to array list
+                ArrayList<String> currentEnumValues = null;
+                if ((enumStringValuesFromCF != null) && (enumStringValuesFromCF.size() > 0)) {
+                    currentEnumValues = new ArrayList<String>(enumStringValuesFromCF.values());
+                }
+
+                // validate enum values count and write to index or bad metric
+                validateThresholdAndWrite(locator, currentEnumValues);
+
```
tilogaat: Would it be more efficient to create a list of locators that need to be written and write them all to ES in one batch? Not sure what the performance implications are of doing a separate write for each metric. Do you know?

blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/inputs/handlers/HttpHandlerIntegrationTest.java
VinnyQ: I'll using your version of this file when merging.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/EnumValidator.java
VinnyQ: that is correct, just the currentEnumValues.  Not the union.  Note that the above is not a union, just a test for differences.  If there are differences, it's writing the currrentEnumValues (from Cassandra) to elasticsearch, replacing any old values.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/EnumValidator.java
VinnyQ: It would definitely help.  I can probably add a new method in AstyanaxReader that does so once I have rebased your PR into my branch.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/EnumValidator.java
VinnyQ: Yes, George has already done that in a previous story when he added the metric_excess_enums table, which is already merged in.  I believe it's this PR here:  https://github.com/rackerlabs/blueflood/pull/526.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/EnumValidator.java
VinnyQ: That's a good point.  I think the DiscoveryWriter was created for that purpose.  I think I can refactor the code to use it for this purpose once I figure out how it works.

VinnyQ: Although, from what I can tell, the writes to ES are extremely fast with the routing enabled.  And we'll only be writing if there are new enum values ingested for a metric, which probably won't happen very often.  If DiscoveryWriter makes the code more complex to maintain, I might recommend that we just keep it simple as-is for now and see what performance looks like, then optimize it if we see a need.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
```diff
@@ -679,4 +679,32 @@ public Boolean apply(Row<Locator, Long> row) {
         }
         return columns;
     }
+
+    public Map<Locator, ArrayList<String>> getEnumStringMappings(final List<Locator> locators) {
```
VinnyQ: @tilogaat I added this method which will query return locator to ArrayList of enum string values mapping, please review.

blueflood-elasticsearch/src/test/java/com/rackspacecloud/blueflood/io/EnumElasticIOTest.java
```diff
@@ -70,6 +78,16 @@ public void setup() throws IOException {
...
+
+        // setup mock for ElasticSearch Client
+        when(bulkRequestBuilderMock.execute()).thenReturn(listenableActionFutureMock);
+        when(esClientMock.prepareBulk()).thenReturn(bulkRequestBuilderMock);
+        when(esClientMock.prepareIndex(EnumElasticIO.ENUMS_INDEX_NAME_WRITE, EnumElasticIO.ENUMS_DOCUMENT_TYPE))
+                .thenReturn(indexRequestBuilderMock);
+        when(indexRequestBuilderMock.setId(anyString())).thenReturn(indexRequestBuilderMock);
+        when(indexRequestBuilderMock.setSource(any(XContentBuilder.class))).thenReturn(indexRequestBuilderMock);
+        when(indexRequestBuilderMock.setCreate(anyBoolean())).thenReturn(indexRequestBuilderMock);
+        when(indexRequestBuilderMock.setRouting(anyString())).thenReturn(indexRequestBuilderMock);
```
VinnyQ: Just some refactoring here, moving mock setup to setup() so it's reusable in multiple tests.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/EnumValidator.java
```diff
@@ -0,0 +1,146 @@
...
+                try {
+                    discoveryIO.insertDiscovery(enumMetric);
+                }
+                catch (Exception e) {
+                    log.error(String.format("Exception writing enums index to elasticsearch for %s: %s", locator.toString(), e.getMessage()), e);
+                }
+            }
+        }
+    }
+
```
tilogaat: nit: nested ifs. is it easy and more readable to extract some of them into their own methods?

tilogaat:
+1! <3 the tests so much

PR #532, opened 2015-10-05
Ingestion server shutdown

izrik:
This PR adds a `shutdownService` method to the `HttpMetricsIngestionServer` class. This would allow us to shutdown the ingestion service during tests, so that we can run several tests separately, instead of trying to cram everything into one test class.

### DIFF
--- a/blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionServerShutdownIntegrationTest.java
+++ b/blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionServerShutdownIntegrationTest.java
@@ -41,6 +41,7 @@ import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
 
+import java.net.ConnectException;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.Calendar;
@@ -88,9 +89,6 @@ public class HttpMetricsIngestionServerShutdownIntegrationTest {
         client = vendor.getClient();
     }
 
-    @Rule
-    public final ExpectedException exception = ExpectedException.none();
-
     @Test
     public void testHttpIngestionHappyCase() throws Exception {
 
@@ -110,8 +108,12 @@ public class HttpMetricsIngestionServerShutdownIntegrationTest {
         server.shutdownServer();
 
         // then
-        exception.expect(java.net.ConnectException.class);
-        HttpResponse response2 = client.execute(post2);
+        try {
+            HttpResponse response2 = client.execute(post2);
+            Assert.fail("We should have received a Connect exception");
+        } catch (ConnectException ex) {
+            Assert.assertEquals("Connection refused", ex.getMessage());
+        }
     }
 
 
--- a/blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionServerShutdownIntegrationTest.java
+++ b/blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionServerShutdownIntegrationTest.java
@@ -1,0 +1,127 @@
+/*
+ * Copyright 2015 Rackspace
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


izrik:
Failures: I think `HttpMetricsIngestionServerShutdownIntegrationTest` and `HttpHandlerIntegrationTest` are interfering with each other. I'll investigate.

izrik:
Ok, tests pass now.

blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionServerShutdownIntegrationTest.java
```diff
@@ -0,0 +1,129 @@
...
+                ContentType.APPLICATION_JSON);
+        post2.setEntity(entity2);
+
+        // when
+        server.shutdownServer();
+
+        // then
+        try {
+            HttpResponse response2 = client.execute(post2);
+            Assert.fail("We should have received a Connect exception");
```
tilogaat: Can I suggest using the ExpectedException from Junit instead of using Assert.fail

```
@Test(expected=ConnectException.class)
```
```

File: blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionServer.java [hunk 2]
```diff
@@ -75,6 +72,10 @@
     private TimeValue timeout;
     private static int MAX_CONTENT_LENGTH = 1048576; // 1 MB
 
+    private ServerBootstrap bootstrap;
+    private ChannelFactory channelFactory;
+    private ChannelGroup allOpenChannels = new DefaultChannelGroup("allOpenChannels");
```
tilogaat: As far as I can see, the field ```bootstrap``` is never used. We just assign server to it. Do we need it? 
Also possibly we do not need ```channelFactory``` as an instance variable, do we? We need only ```allOpenChannels``` to be the instance variable since it's used to shutdown the server.

blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionServerShutdownIntegrationTest.java
tilogaat: Good test, I like the fact that it demonstrates we can separate out ingestion related tests into separate files. Our next step should be to do it.

izrik: I originally did something like that, using jUnit's [`ExpectedException`](http://junit.org/javadoc/latest/org/junit/rules/ExpectedException.html) rule. However, `ConnectException` can indicate a number of failure conditions, including timeout on connection.

For example, if the service did not shut down correctly, and the socket is still open, but it's timing out while opening a connection (for whatever reason). That would give us a false pass, because a timeout on a socket connection would yield a `ConnectException`. If our test code is only looking for the exception, and not considering the underlying cause, it would be incorrect. Therefore, we have to look at the exception message to make sure that it actually represents a refused connection; and in order to do that, we need the exception object to assert against, which prevents the use of the `ExpectedException` rule.

Would it help if I added a comment explaining the `try/catch` to future readers?

FWIW, Groovy actually has something that does this, built into the language. You could type:

```
when:
// do something

then:
def ex = notThrown(ConnectException)
ex.message == "Connection refused"
```

It looks much nicer than explicitly catching the exception, can allows you to make assertions on it.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionServer.java
izrik: True. I had thought about adding a stage to the pipeline (in `MetricsHttpServerPipelineFactory`'s `getPipeline` method) to collect all channels that get created into `allOpenChannels`. Then when we shut it down, we could make sure all channels get closed, and call `releaseExternalResources` on the factory as well. However, I'm not entirely sure about that right now. Keeping track of all channels that get created, over an extended period of time, could consume a lot of memory needlessly. It would depend on whether `ChannelGroup` is smart enough to *remove* closed channels from itself automatically, which I don't know yet. I'll remove the extraneous fields for now.

blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionServerShutdownIntegrationTest.java
tilogaat: Yea this explanation in a comment will be very useful.

tilogaat:
+1

PR #533, opened 2015-10-06
Explanatory note on the try/catch in the test.

izrik:
A follow-on to PR https://github.com/rackerlabs/blueflood/pull/532. No code changes, just a comment to explain why we need to put a try/catch in the test, which is normally unorthodox.


tilogaat:
+1

PR #534, opened 2015-10-13
WIP: Rollup integration test

tilogaat:
What:
Integration test that tests number of rollup points at each granularity for enum rollups. 

How: 
Generates a new point for an enum metric once every minute, writes it to Cassandra, generates rollups at each granularity for it and then finally queries the rolled up points from Cassandra
Refactored some logic for code reuse. 

Why:
The rollup work was already coded. Adding a test to verify that all the wiring works. 

Todo:
Rollup generation is terribly time consuming. Trying to speed it up by using more threads.


tilogaat:
Accidentally opened it against master. Should've opened it against enum-feature

PR #535, opened 2015-10-13
RFR: Rollup integration test

tilogaat:
What:
Integration test that tests number of rollup points at each granularity for enum rollups. 

How: 
Generates a new point for an enum metric once every minute, writes it to Cassandra, generates rollups at each granularity for it and then finally queries the rolled up points from Cassandra
Refactored some logic for code reuse. 
The rollup generation process is really time consuming. Therefore, combined all tests into one test. 

Why:
The rollup work was already coded. Adding a test to verify that all the wiring works. 

~~Todo:~~
~~Rollup generation is terribly time consuming. Trying to speed it up by using more threads.~~

### DIFF
--- a/blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/outputs/handlers/HttpEnumRollupIntegrationTest.java
+++ b/blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/outputs/handlers/HttpEnumRollupIntegrationTest.java
@@ -1,0 +1,131 @@
+/*
+ * Copyright 2013 Rackspace
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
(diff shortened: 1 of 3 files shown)


blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/outputs/handlers/HttpEnumRollupIntegrationTest.java
```diff
@@ -0,0 +1,131 @@
...
+            MetadataCache.getInstance().put(enumMetric.getLocator(), MetricMetadata.ROLLUP_TYPE.name().toLowerCase(), RollupType.ENUM.toString());
+            writer.insertMetrics(metrics, CassandraModel.CF_METRICS_PREAGGREGATED_FULL);
+        }
+
+        httpHandler = new HttpRollupsQueryHandler();
+
+        // generate every level of rollup for the raw data
+        Granularity g = Granularity.FULL;
+        while (g != Granularity.MIN_1440) {
+            g = g.coarser();
```
VinnyQ: This looks like we will never run generteEnumRollups against the granularity of Granularity.FULL.  Is this a correct thing to do?  If it's not, move this line below the enumLocators for loop.

tilogaat: @VinnyQ Yes that's the correct thing to do, because we are specifying the destination granularity. The full resolution data already exists.

VinnyQ: ah cool, that makes sense.

VinnyQ:
:+1: lgtm

PR #537, opened 2015-10-16
add ingest and query requests and tests for enum metrics

VinnyQ:
WHAT:  Add tests for enum metrics to the grinder

HOW:  add ingestion and query classes for enum metrics requests to POST and GET metrics with enum values.


contrib/grinder/scripts/ingestenum.py
```diff
@@ -0,0 +1,86 @@
...
+from utils import *
+from net.grinder.script import Test
+from net.grinder.plugin.http import HTTPRequest
+
+class EnumIngestThread(AbstractThread):
+  # The list of metric numbers for all threads in this worker
+  metrics = []
+
+  # Grinder test reporting infrastructure
+  test1 = Test(3, "Enum Ingest test")
```
tilogaat: @VinnyQ I think this test number is not unique, which is why it was not running. You need to change the test numbers for all the query tests too.

contrib/grinder/scripts/query.py
```diff
@@ -97,11 +97,27 @@ def generate(self, time, logger):
 #    logger(result.getText())
     return result
 
+class EnumSearchQuery(AbstractQuery):
+  query_interval_name = 'enum_search_queries_per_interval'
+  query_name = "EnumSearchQuery"
+  test_number = 6
```
tilogaat: @VinnyQ The single plot query is 3 in this file, which is probably clashing with your test number.

tilogaat:
@VinnyQ 

In the file grinder.py add this line to make the script work, 

```
import ingestenum
```

contrib/grinder/properties/grinder.properties
```diff
@@ -1,16 +1,19 @@
...
 grinder.bf.name_fmt="t4.int.abcdefg.hijklmnop.qrstuvw.xyz.ABCDEFG.HIJKLMNOP.QRSTUVW.XYZ.abcdefg.hijklmnop.qrstuvw.xyz.met.%d"
 grinder.bf.report_interval=10000
 grinder.bf.annotations_num_tenants=4
 grinder.bf.num_tenants=4
+grinder.bf.enum_num_tenants=4
 grinder.bf.metrics_per_tenant=15
+grinder.bf.enum_metrics_per_tenant=5
 grinder.bf.batch_size=5
 grinder.bf.ingest_concurrency=15
+grinder.bf.enum_ingest_concurrency=15
```
tilogaat: @VinnyQ grinder.properties.local needs this config too.

tilogaat:
@VinnyQ Some observations that I made after running the test:
```
http://localhost:19000/v2.0/tenantId/ingest/aggregated 
```
is returning a 400. 
```
http://localhost:20000/v2.0/3/metrics/search?query=enum_grinder_t4.int.abcdefg.hijklmnop.qrstuvw.xyz.ABCDEFG.HIJKLMNOP.QRSTUVW.XYZ.abcdefg.hijklmnop.qrstuvw.xyz.met.*&include_enum_values=true 
```

is returning a 404, after running it against this blueflood checkout. Do you have the required BF enum changes of enums in this checkout?

VinnyQ:
@tilogaat Fixed the 400 errors, it was because the grinder was submitting array of payloads to the single aggregated ingestion endpoint.  Modifying it to the aggregated/multi endpoint fixes it!

Please review/test when you have a chance and merge if looks good!

contrib/grinder/properties/grinder.properties
VinnyQ: Added!

contrib/grinder/scripts/tests.py
```diff
@@ -138,16 +148,20 @@ def test_init_process(self):
     single_plot_queries_agent0 = int(math.ceil(query.default_config['singleplot_per_interval']/num_query_nodes))
     multi_plot_queries_agent0 = int(math.ceil(query.default_config['multiplot_per_interval']/num_query_nodes))
     search_queries_agent0 = int(math.ceil(query.default_config['search_queries_per_interval']/num_query_nodes))
+    enum_search_queries_agent0 = int(math.ceil(query.default_config['enum_search_queries_per_interval']/num_query_nodes))
     annotation_queries_agent0 = int(math.ceil(query.default_config['annotations_queries_per_interval']/num_query_nodes))
```
tilogaat: @VinnyQ Can we add some tests for slices for the enums ingestion thread? See 
https://github.com/VinnyQ/blueflood/blob/8488324484900a92a53f48191cef50f18112966f/contrib/grinder/scripts/tests.py#L124

contrib/grinder/scripts/tests.py
```diff
@@ -176,19 +190,23 @@ def test_init_process(self):
...
     self.assertEqual(thread.slice, [query.SearchQuery] * 2)
 
+    thread = query.QueryThread(6)
+    self.assertEqual(thread.slice, [query.EnumSearchQuery] * 2)
+
     thread = query.QueryThread(9)
-    self.assertEqual(thread.slice, [query.AnnotationsQuery])
+    self.assertEqual(thread.slice, [query.AnnotationsQuery] * 2)
 
   def test_generate_payload(self):
```
tilogaat: @VinnyQ Do you think we should add a test here for generate payload?

contrib/grinder/scripts/tests.py
VinnyQ: ok i'll add.

contrib/grinder/scripts/tests.py
VinnyQ: good call i'll add

tilogaat:
+1 after adding all tests

tilogaat:
+1 lgtm

PR #538, opened 2015-10-16
Adding some tracking for delayed metrics using the JMX request tracker

tilogaat:
What:
Using the Tracker class to track tenants that may be sending us delayed metrics. 

Why:
Currently, the re-roll metric tracks only the number of re-rolls. We need to add some on demand logging using JMX for tracking the tenants that may be sending us delayed metrics. 

How: 
Used @VinnyQ's request tracker and added another method to track delayed metrics if a flag is set.

### DIFF
--- a/blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/PreaggregateConversions.java
+++ b/blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/PreaggregateConversions.java
@@ -36,7 +36,6 @@ public class PreaggregateConversions {
     // todo: punt on TTL
     private static final TimeValue DEFAULT_TTL = new TimeValue(48, TimeUnit.HOURS);
     private static final String NAME_DELIMITER = "//.";
-    private static final long millisIn10Minutes = 10 * 60 * 1000;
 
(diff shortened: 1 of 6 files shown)


blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/PreaggregateConversions.java
```diff
@@ -43,6 +46,10 @@
     
     public static Collection<IMetric> buildMetricsCollection(AggregatedPayload payload) {
         Collection<IMetric> metrics = new ArrayList<IMetric>();
+        long nowMillis = new DateTime().getMillis();
+        if (nowMillis - payload.getTimestamp() > millisIn10Minutes) {
```
VinnyQ: if the rollup is 10 minutes or older we are considering it "delayed"?

We might want to make this bench mark configurable instead of hard coding it, in case we want to tweek the value.

tilogaat: Fixed in e1cb1eb

VinnyQ:
:+1:

PR #541, opened 2015-10-21
Add tracking of delayed metric names

tilogaat:
*What:*
After speaking to @drenalin23, we decided that tracking names of the metrics would also be useful along with the tenant id.

### DIFF
--- a/blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionHandler.java
+++ b/blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionHandler.java
@@ -124,7 +124,7 @@ public class HttpMetricsIngestionHandler implements HttpRequestHandler {
                 containerMetrics = jsonMetricsContainer.toMetrics();
                 forceTTLsIfConfigured(containerMetrics);
 
-                if (!jsonMetricsContainer.areDelayedMetricsPresent()) {
+                if (jsonMetricsContainer.areDelayedMetricsPresent()) {
                     Tracker.trackDelayedMetricsTenant(tenantId);
                 }
             } catch (InvalidDataException ex) {
--- a/blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionHandler.java
+++ b/blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionHandler.java
@@ -125,7 +125,7 @@ public class HttpMetricsIngestionHandler implements HttpRequestHandler {
                 forceTTLsIfConfigured(containerMetrics);
 
                 if (jsonMetricsContainer.areDelayedMetricsPresent()) {
-                    Tracker.trackDelayedMetricsTenant(tenantId);
+                    Tracker.trackDelayedMetricsTenant(tenantId, jsonMetricsContainer.getDelayedMetricNames());
                 }
             } catch (InvalidDataException ex) {
                 // todo: we should measure these. if they spike, we track down the bad client.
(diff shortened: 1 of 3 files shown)


blueflood-http/src/main/java/com/rackspacecloud/blueflood/tracker/Tracker.java
```diff
@@ -123,9 +123,14 @@ public static void track(HttpRequest request) {
...
     }
 
-    public static void trackDelayedMetricsTenant(String tenantid) {
+    public static void trackDelayedMetricsTenant(String tenantid, List<String> metricNames) {
         if (isTrackingDelayedMetrics) {
-            String logMessage = String.format("[TRACKER][DELAYED METRIC] Tenant sending delayed metric is %s",tenantid);
+            StringBuilder sb = new StringBuilder();
+            for(String name : metricNames) {
+                sb.append(name);
+                sb.append(";");
```
VinnyQ: are you sure you want a ';' to separate metric names, and not a comma?  it'll read better in the log/kibana with commas

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionHandler.java
```diff
@@ -124,8 +124,8 @@ public void handle(ChannelHandlerContext ctx, HttpRequest request) {
                 containerMetrics = jsonMetricsContainer.toMetrics();
                 forceTTLsIfConfigured(containerMetrics);
 
-                if (!jsonMetricsContainer.areDelayedMetricsPresent()) {
-                    Tracker.trackDelayedMetricsTenant(tenantId);
+                if (jsonMetricsContainer.areDelayedMetricsPresent()) {
```
VinnyQ: why is the logic swapped around from not to is present?  did we get it wrong initially?

blueflood-http/src/main/java/com/rackspacecloud/blueflood/tracker/Tracker.java
```diff
@@ -123,9 +123,14 @@ public static void track(HttpRequest request) {
...
+            }
+            String logMessage = String.format("[TRACKER][DELAYED METRIC] Tenant sending delayed metric is %s with the delayed metrics -- %s",tenantid,sb.toString());
```
VinnyQ: I wonder if there's a way for us to track how delayed these metrics coming in are.  Can we log the timestamp of ingestion vs metric's timestamp somehow and log it?

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionHandler.java
tilogaat: yes @VinnyQ i had got it wrong. We want to track if there are delayed metrics present.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/tracker/Tracker.java
tilogaat: I can make it a comma.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/tracker/Tracker.java
VinnyQ: this is hard to do during rollup.  How about we add another tracker interface for the Requests handler, that will allow us to turn on tracking for tenantID AND metric name?  Once we have a few of these tenant ids + metricnames coming in that are flagged as delayed metrics, we can turn on the tracking at the request end and see the exact ingestion time and entire payload.

VinnyQ: @tilogaat  let's make this a new story, I am going to +1 this PR once you answered the previous questions

VinnyQ:
:+1: lgtm.  Merge when checks are completed.

tilogaat:
Checks complete merging.

PR #542, opened 2015-10-21
Metric name tracking is causing the log message to be humongous. Not …

tilogaat:
…logging metric names for now.


VinnyQ:
:+1: 

Not sure how complicated you want to make this, but if we really want to log metric names maybe we can do some sort of range filter.  So instead of logging all metric names, you can specify a range of names e.g. from 0-10, 10-20, etc.

PR #548, opened 2015-10-29
Make the HttpEnumRollupIntegrationTest less fragile

tilogaat:
*What:*
Make the HttpEnumRollupIntegrationTest less fragile

*How:*
Sometimes the points rolled up do not exactly correspond to the points expected (because 1500 points roll up to 288 or 289 depending on when the timestamp falls) and the test fails. We want the test to be less fragile and a number in the ballpark of +-5 is ok with us. So I took the difference between the expected and the given, and check to see if the absolute value of the difference is less than 5.

### DIFF
--- a/blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/outputs/handlers/HttpRollupHandlerIntegrationTest.java
+++ b/blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/outputs/handlers/HttpRollupHandlerIntegrationTest.java
@@ -154,7 +154,7 @@ public class HttpRollupHandlerIntegrationTest extends IntegrationTestBase {
                         from,
                         to,
                         points.get(g2));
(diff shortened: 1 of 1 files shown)


blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/outputs/handlers/HttpRollupHandlerIntegrationTest.java
```diff
@@ -154,7 +154,7 @@ public static void testHTTPRollupHandlerGetByPoints(Map<Locator, Map<Granularity
                         from,
                         to,
                         points.get(g2));
-                Assert.assertEquals((int) answers.get(locator).get(g2), data.getData().getPoints().size());
+                Assert.assertTrue(Math.abs((int) answers.get(locator).get(g2)-data.getData().getPoints().size()) <= 5);
```
VinnyQ: good use of abs, +1

VinnyQ:
:+1:  lgtm, thanks for fixing @tilogaat

PR #549, opened 2015-11-05
script for creating column family in db to support enum metrics

VinnyQ:
WHAT:  need to create two new tables to support enum metrics:  "metrics_enum" and "metrics_excess_enums"

HOW:  manually execute the script in this pr (01_add_enums_cf.cql) to create the tables in the respective (stage/prod) environment like so:

```
$ cqlsh 127.0.0.1 -f load.cdl

-- syntax if username/password require
$ cqlsh 127.0.0.1 -f load.cdl -u username -p password
```

Note:  see https://github.com/rackerlabs/blueflood-chef/pull/702 for the cql in blueflood-chef to use for our own environment

### DIFF
--- a/src/cassandra/cli/load.cdl
+++ b/src/cassandra/cli/load.cdl
@@ -1,0 +1,180 @@
+-- this cdl schema create the main tables for blueflood in cassandra using the cqlsh client
+-- usage:
+--   $ cqlsh hostname -f load.cdl -u username -p password
+-- example:
+--   $ cqlsh 127.0.0.1 -f ./blueflood/src/cassandra/cli/load.cdl
+
+CREATE KEYSPACE IF NOT EXISTS "DATA"
+    WITH REPLICATION = { 'class' : 'SimpleStrategy', 'replication_factor' : 1 };
+
+
+CREATE TABLE IF NOT EXISTS "DATA".metrics_locator (
+    key bigint,
+    column1 text,
+    value text,
+    PRIMARY KEY (key, column1)
+) WITH COMPACT STORAGE AND speculative_retry = 'NONE';
+
+CREATE TABLE IF NOT EXISTS "DATA".metrics_discovery (
+    key text,
+    column1 text,
+    value text,
+    PRIMARY KEY (key, column1)
+) WITH COMPACT STORAGE AND speculative_retry = 'NONE';
+
+CREATE TABLE IF NOT EXISTS "DATA".metrics_state (
+    key bigint,
+    column1 text,
+    value bigint,
+    PRIMARY KEY (key, column1)
+) WITH COMPACT STORAGE AND speculative_retry = 'NONE';
+
+CREATE TABLE IF NOT EXISTS "DATA".metrics_metadata (
+    key text,
+    column1 text,
+    value blob,
+    PRIMARY KEY (key, column1)
+) WITH COMPACT STORAGE AND speculative_retry = 'NONE';
+
+CREATE TABLE IF NOT EXISTS "DATA".metrics_string (
+    key text,
+    column1 bigint,
+    value text,
+    PRIMARY KEY (key, column1)
+) WITH COMPACT STORAGE AND speculative_retry = 'NONE';
+
+CREATE TABLE IF NOT EXISTS "DATA".metrics_enum (
+    key text,
+    column1 bigint,
+    value text,
+    PRIMARY KEY (key, column1)
+) WITH COMPACT STORAGE AND speculative_retry = 'NONE';
+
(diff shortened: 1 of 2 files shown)


VinnyQ:
@tilogaat @drenalin23 @goru97 please review.

I am considering setting up a main 00_blueflood_cf.cql that have the CQL syntax for loading up the original blueflood cf metrics tables using the CREATE TABLE IF NOT EXISTS, and deprecating/removing the current load.script file.

This way, if we want to, we can add to the chef cassandra recipe to execute these .cql scripts in order after installing/updating/validating cassandra, in the order of the filename with the two numeric values in front (e.g. 00_, 01_...), which create the tables IF NOT EXISTS for all the existing tables.

Let me know if you wish for me to do so (converting load.script which uses the deprecated cassandra-cli to execute, to 00_blueflood_cf.cql which we can use cqlsh to execute).

drenalin23:
Vinny that sounds like a good idea.

drenalin23:
@VinnyQ did you also see the schema file in blueflood-chef contrib/cassandra - we should update the file too or have them all in one place/process if you don't think there should be two files/methods floating around.

VinnyQ:
@drenalin23 good catch.  Let's stick with the blueflood-chef/contrib/cassandra version as our definitive loading script for *our* stage/prod environment.  I have made the changes and added a new .cql script to be ran with cqlsh here:  https://github.com/rackerlabs/blueflood-chef/pull/702

The one in blueflood we will leave a general script for the opensource community when setting up this project, and they can modify it for their own environment.  I have updated the script here to just have general create table commands for loading up the project.

@drenalin23 @tilogaat @goru97 please review when you have a chance.

tilogaat:
+1

src/cassandra/cli/load.cdl
```diff
@@ -0,0 +1,180 @@
...
+CREATE KEYSPACE IF NOT EXISTS "DATA"
+    WITH REPLICATION = { 'class' : 'SimpleStrategy', 'replication_factor' : 1 };
+
+
+CREATE TABLE IF NOT EXISTS "DATA".metrics_locator (
+    key bigint,
+    column1 text,
+    value text,
+    PRIMARY KEY (key, column1)
+) WITH COMPACT STORAGE AND speculative_retry = 'NONE';
```
VinnyQ: Maybe i should completely eliminate these other two settings two for CREATE TABLE?

They are explicitly different from the default that looks like it might be better for local (single node i.e. your computer) testing.

tilogaat: Will they cause a problem in local testing?

VinnyQ: no if they're different, but the two settings above "WITH COMPACT STORAGE AND speculative_retry = 'NONE';" just make sure the database file is small and no unnecessary cpu cycle attempting to retry db transactions on failure which is only useful in actual multi-nodes cluster (and not single node cluster).

VinnyQ: read about COMPACT STORAGE here:  http://blog.librato.com/posts/cassandra-compact-storage

but a quick quote/snippet from it:
```
mainly targeted towards backward compatibility for definitions created before CQL3... provides a slightly more compact layout of data on disk but at the price of diminished flexibility and extensibility... is not recommended outside of the backward compatibility reason evoked above.

Prior to CQL when the Thrift interface was the only API, all tables were built with compact storage. As the documentation states though, it is a legacy option not intended for anything besides backwards compatibility.
...
In the article, they mentioned they saw a 30x storage size overhead when not using compact storage on a table.
...
Our primary use case for Cassandra is large time-series data set storage. We must optimize for maximum on-disk efficiency since our storage load and hence costs are directionally proportional to the usage of the service. An extra 35% overhead on the storage footprint is significant.
...
If storage efficiency isn’t your primary driving factor, then the schema flexibility and transactional updates that a non-compact table provides can make application development easier.
```

Read about speculative_retry here:  http://www.datastax.com/dev/blog/rapid-read-protection-in-cassandra-2-0-2

but basically it does this:
```
Rapid read protection can be configured to do retry after a fixed period of milliseconds or after a percentile of the typical read latency (tracked per table).
```

TLDR:  both options are really good to have tables created in a local cassandra cluster for testing.  The COMPACT STORAGE option specifically should be apply to any cassandra tables created for blueflood, since it fits the bill for what we're doing (time series data with very little

VinnyQ: OK I am making a decision to leave the two settings for these tables.  Going to merge.

tilogaat: Great research @VinnyQ

PR #550, opened 2015-11-06
Added deploy condition on jdk type

goru97:


VinnyQ:
@goru97 why is this needed?

tilogaat:
@goru97 @VinnyQ i think this is a part of the project to automatically push out builds. It only creates one build for Oracle java and not openjdk

tilogaat:
@goru97 Can you confirm that's what it does?

goru97:
@VinnyQ @tilogaat Previously after adding my changes to travis.yml, travis was creating uber jar whenever there was a push to rax-release branch and tried to deploy it in releases. Because there were two build processes running under different java environment (oracle jdk & open jdk), the first one to complete the build process was able to deploy it and second one failed because there was already a build deployed with matching tag. So to take care of this problem I added this condition as @tilogaat explained: it will only deploy for environment with oraclejdk.

tilogaat:
+1

PR #551, opened 2015-11-10
Default blueflood.conf and blueflood-log4j.properties files for sample run

usnavi:
Contains necessary files to run blueflood locally at ```demo/local/config```.  Assumes the following
- Cassandra is accessible at ```127.0.0.1:9160```
- Elastic search is accessible at ```localhost:9300``` (not required to have blueflood accept requests)

Once this is accepted, I'll update the wiki with these files.


VinnyQ:
@usnavi going to rerun the travis build to see if it's still failing before I +1.  It could just be a hiccup with latency.

VinnyQ:
:+1:

tilogaat:
+1

PR #552, opened 2015-11-10
WIP: Added more fine grained metric name jmx tracking

tilogaat:
*What:*
Added more finer grained control over what we log using JMX. 

*Why:*
Currently we only have the capability of tracking all payloads for a particular tenant, which can be too noisy. This PR adds capability of tracking only certain metrics

*How:*
Added a hashset to hold all the metric names that need to be tracked. If the http response contains even a single one of these metric names, it is logged. 
The capability to add metric name tracking is optional. Usage for our tenant tracker script would be like this. 

```
./blueflood_tenant_tracker.sh add 814831 prod_all
./blueflood_tenant_tracker.sh addMetric example.metric.one 
```


tilogaat:
Superseded by PR #590

PR #553, opened 2015-11-11
remove unneeded assert in test that breaks when run under jdk 1.8

shintasmith:
JDK 1.8 hashes differently than JDK 1.7. We are doing a string compare on a JSON payload (containing hashed keys) before and after deserialization/serialization. This is unneeded Assert -- we are not testing any of our code, since we are just calling methods on Jackson's ObjectMapper.

I propose we remove this Assert since it's breaking with JDK 1.8

### DIFF
--- a/blueflood-core/src/test/java/com/rackspacecloud/blueflood/io/serializers/IMetricSerializerTest.java
+++ b/blueflood-core/src/test/java/com/rackspacecloud/blueflood/io/serializers/IMetricSerializerTest.java
@@ -74,8 +74,6 @@ public class IMetricSerializerTest {
         BluefloodSetRollup setDeserialized = mapper.readValue(setValue, BluefloodSetRollup.class);
         String setSerialized = mapper.writeValueAsString(setDeserialized);
 
-        //Assert.assertEquals(setValue, setSerialized);
-
         BluefloodSetRollup setReserialized = mapper.readValue(setSerialized, BluefloodSetRollup.class);
         Assert.assertEquals(setDeserialized, setReserialized);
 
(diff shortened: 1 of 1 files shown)


shintasmith:
@VinnyQ, @tilogaat, please review.

cc: @usnavi, @chandraaddala

ChandraAddala:
Looks good to me.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/io/serializers/IMetricSerializerTest.java
```diff
@@ -73,7 +73,6 @@ public void testSetSerialization() throws IOException {
 
         BluefloodSetRollup setDeserialized = mapper.readValue(setValue, BluefloodSetRollup.class);
         String setSerialized = mapper.writeValueAsString(setDeserialized);
-        Assert.assertEquals(setValue, setSerialized);
```
VinnyQ: What are we going to be testing here if this is removed?  Why is jdk 1.8 complaing about this assert?

VinnyQ: Nevermind you explained it in the PR description.  Thanks for catching @shintasmith !

tilogaat: I think the way a set is serialized in Java 8, is a little bit different than in JAva 7. I am ok with removing this assert since we are comparing the 2 objects ultimately. However, this type of test is all over in Blueflood and we may run into similar issues later too.

tilogaat:
+1

PR #554, opened 2015-11-11
Removing metrics_mapping.json

usnavi:
Removig metrics_mapping.json since metrics_mapping_v1.json is now used.
Updating tests to use metrics_mapping_v1.json


usnavi:
@VinnyQ  and @tilogaat would you please review

tilogaat:
+1

VinnyQ:
:+1: 
nice, i guess it was only used in one place.  Glad the tests are still working after the changes.

PR #555, opened 2015-11-11
fix condition for writing enum values to elasticsearch enums index

VinnyQ:
WHAT:  fix bug with EnumValidator attempting to write to elasticsearch with non-enum metrics if ElasticSearch cluster is down/unreachable.

WHY:  the result from cassandra retrieving enum values return in a non-null object with size 0, and current code only tested for NULL result.

HOW:  modify the condition to check for both non-null and size > 0 for enum values query of a metric from cassandra, if neither is true then it's a no-op for write to ElasticSearch.


VinnyQ:
@tilogaat @goru97 please review when you have a chance

tilogaat:
+1

PR #556, opened 2015-11-12
Created startup script

ChandraAddala:
I had created a startup script to run blueflood. Looks less verbose compared to the long java command. I had referenced the config files in demo/local/config.

If this looks like a good idea, merge it and I will update the instructions accordingly on the wiki. If not, I dont feel very strongly about this and we leave it the way it is now.


tilogaat:
Excellent idea! +1

shintasmith:
:+1: I had done the same thing, I created a local shell script. I think it's great to have one in the git repo. But I would probably name it blueflood-start.sh to be specific. But that's just me. :-)

PR #557, opened 2015-11-12
fix tests with metrics_mapping.json

VinnyQ:
WHAT:  fix failing tests from master changes of metrics_mapping.json (removal of metrics_mapping_v1.json).

HOW:  modify tests to use metris_mapping.json and change json content to use latest mappings on the elasticsearch search nodes (with analyzer and tokenizer).

NOTE: the failing tests are from *other* tests looking for metrics_mapping.json that was removed in master.  but since I renamed metrics_mapping_v1.json back to metrics_mapping.json, I don't have to modify those broken tests and just have to fix any tests still referencing metrics_mapping_v1.json.

### DIFF
--- a/blueflood-elasticsearch/src/main/resources/metrics_mapping.json
+++ b/blueflood-elasticsearch/src/main/resources/metrics_mapping.json
@@ -13,8 +13,9 @@
                 "index": "not_analyzed"
             },
             "metric_name": {
-                "type": "string",
-                "index": "not_analyzed"
+                "index_analyzer": "prefix-test-analyzer",
+                "search_analyzer": "keyword",
+                "type": "string"
             }
         }
     }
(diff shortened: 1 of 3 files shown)


VinnyQ:
@tilogaat @usnavi @shintasmith @goru97 @ChandraAddala 

Please review.  Have to fix tests in enums-feature from us removing metrics_mapping.json in master.  I took the opportunity to just clean it up and renamed it back to metrics_mappings.json so that we don't just have a random _V1 file.  I've also decided that we should just go with the latest mappings setup in our env because we should be testing against those settings and mappings in our development.

Note that I have removed the step in the https://github.com/rackerlabs/blueflood/wiki/Setup-ElasticSearch wiki that relies on this mapping, so users can just setup ES and use the defaults (no schema) setup initially, which still works.  Or they can grab these test resources files and modify them to their own liking.

Not sure if I should add a "simple" version of the mapping (without the analyzer and tokenizer), to resources just so other users have a simple mapping to use.  At the moment I am leaning toward no.

tilogaat:
+1

shintasmith:
lgtm, :+1:

PR #559, opened 2015-11-19
Make the enums rollup integration test more realistic.

tilogaat:
*What:*
Added a fix for the enums test so that only one of 5 values will be picked for an enum value instead of a new one every single time

*Why:*
Previously, the test was picking a new enum value every single time for each of it's posts. It causes the enum value list to have 1400 unique values (which is not representative of a real enum) causing the enums rollup to go on for a long time. 

*How:*
Create a static list of enum values. ```pickAnEnumValue``` will pick one of the 5 values in that list.

### DIFF
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/IntegrationTestBase.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/IntegrationTestBase.java
@@ -157,7 +157,7 @@ public class IntegrationTestBase {
 
     protected PreaggregatedMetric getEnumMetric(String name, String tenantid, long timestamp) {
         final Locator locator = Locator.createLocatorFromPathComponents(tenantid, name);
-        BluefloodEnumRollup rollup = new BluefloodEnumRollup().withEnumValue(pickAnEnumValue()).withEnumValue(pickAnEnumValue(),1L);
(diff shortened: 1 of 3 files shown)


blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/IntegrationTestBase.java
```diff
@@ -86,6 +86,7 @@ public MutationBatch createMutationBatch() {
     private static final char[] STRING_SEEDS = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ01234567890_".toCharArray();
     private static final Random rand = new Random(System.currentTimeMillis());
     protected static final ConcurrentHashMap<Locator, String> locatorToUnitMap = new ConcurrentHashMap<Locator, String>();
+    private static final List<String> enumValueList = new ArrayList<String>() {{ add("A");  add("B");  add("C"); add("D"); add("E"); }};
```
VinnyQ: You can also write this as:

```
private static final List<String> enumValueList = Arrays.asList("A", "B", "C", "D", "E");
```

VinnyQ:
:+1: 
besides one minor NIT, looks good tilo!  good work!

PR #560, opened 2015-11-20
Fixing a bug with the collectionTime of the enum payload

tilogaat:
*What:*
It's not required to multiply the collectionTime with 1000, because it's already converted into milliseconds before passing it onto the individual threads. Corrected tests to reflect this bug.


GeorgeJahad:
+1

PR #561, opened 2015-11-24
Spinning up Enum validator threads only for Enum metrics

tilogaat:
*What*:
The enum validation code is performed only for enum metrics and not for all others. 

*Why*:
After deploying enums-feature to staging, we found that the heap consumption of rollups is through the roof, as is the rollup wait time. We discovered that a new runnable is created for every metric, which may be the cause of it. 
 
*How*:
We check metadata cache to see if the metric that is being rolled up has the rollupType Enum. Only then we create an EnumValidator runnable and go through the whole process of getting enum values from ElasticSearch.


GeorgeJahad:
+1

PR #563, opened 2015-11-30
WIP: Instrumenting enum rollups

tilogaat:
*What:*
Adding a meter to mark the number of enum rollups being scheduled per minute

*How:*
We already know which locators from a given batch of to-be-rolled-up locators are enums, because we only validate Enums. Utilizing that hashset of enum locators to determine how many locators in the current to-be-rolled batch are enums by incrementing a meter. I used SingleRollupReadContext because it is already being used to track the ```*.iad.blueflood.*.com.rackspacecloud.blueflood.service.RollupService.Rollup-Execution-Timer.count``` (Rollups_per_minute count in the Grafana dashboard)

TODO:
Moar testing!

### DIFF
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
@@ -708,7 +708,7 @@ public class AstyanaxReader extends AstyanaxIO {
         return map;
     }
 
-    public Set<Locator> getEnumLocatorsFromLocatorSet(Set<Locator> locators) throws Exception{
+    public HashSet<Locator> getEnumLocatorsFromLocatorSet(Set<Locator> locators) throws Exception{
         HashSet<Locator> locatorHashSet = new HashSet<Locator>();
 
         for (Locator loc : locators) {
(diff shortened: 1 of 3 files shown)


blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
```diff
@@ -708,7 +708,7 @@ public Boolean apply(Row<Locator, Long> row) {
         return map;
     }
 
-    public Set<Locator> getEnumLocatorsFromLocatorSet(Set<Locator> locators) throws Exception{
+    public HashSet<Locator> getEnumLocatorsFromLocatorSet(Set<Locator> locators) throws Exception{
         HashSet<Locator> locatorHashSet = new HashSet<Locator>();
```
GeorgeJahad: NIT: This change isn't necessary is it?  In Java, my understanding is it is preferable to return the interface, (i.e. Set), and leave the implementation, (HashSet,) as an implementation detail in the method.

GeorgeJahad:
+1

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AstyanaxReader.java
usnavi: :+1:

tilogaat:
Closing this in favor of PR #569

PR #564, opened 2015-12-01
[CMD-714] Python script to clear errant enums.

ChandraAddala:


usnavi:
lgtm - but i've never worked with python.  Two questions:
1) I assume the blank ``contrib/enum_metrics/enum_metrics/__init__.py`` is required?
2) Is there any automated mechanism which runs the unit tests, say in the PR-builder or elsewhere?

ChandraAddala:
@usnavi 
1) Thats correct. Having __init__.py makes that a package. 
2) No. I have to look into that.

contrib/enum_metrics/enum_metrics/clear_errant_enums.py
```diff
@@ -0,0 +1,128 @@
+import sys
+import argparse
+import json
+import dbclient as db
+import esclient as es
```
GeorgeJahad: i don't see the esclient anywhere.  Am I missing something?

contrib/enum_metrics/enum_metrics/dbclient.py
```diff
@@ -0,0 +1,223 @@
...
+        database results.
+
+        :param results:
+        :return:
+        """
+
+        return [(x.key, x.column1, x.value) for x in results]
+
+
+    def get_metrics_excess_enums(self, metric_name):
```
GeorgeJahad: it looks like the get and delete methods below are all identical except for the column family.  Wouldn't it be better to add the CF as a parameter and create a single get and delete method? Within reason, I like DRY code.

contrib/enum_metrics/test/test_clear_errant_enums.py
```diff
@@ -0,0 +1,27 @@
+from unittest import TestCase
```
GeorgeJahad: would it be hard to use something like this:
https://pypi.python.org/pypi/testing.cassandra
to create a dummy cassandra server and dummy data, to create some more extensive unit tests?

contrib/enum_metrics/enum_metrics/clear_errant_enums.py
ChandraAddala: Missed adding that file...Its in the PR now..

contrib/enum_metrics/enum_metrics/dbclient.py
ChandraAddala: I thought about it when writing the code. It felt like I am writing dynamic SQL. I thought I would rather have simple select queries for each table. Since we are not doing any complex query, may be I could extract that logic to one method.

contrib/enum_metrics/test/test_clear_errant_enums.py
ChandraAddala: I am exploring that currently. I thought I will create this PR, get everyone's opinion to see if I am going in the right direction. If it works out okay, I will submit it in my subsequent PR. I am also looking at somehting similar for elastic search.

contrib/enum_metrics/enum_metrics/dbclient.py
```diff
@@ -0,0 +1,181 @@
...
+        """
+
+        key = tenant_id + '.' + metric_name
+        return self.__get_metrics(key, "metrics_excess_enums")
+
+    def get_metrics_preaggregated_full(self, key):
+        """
+        Retrieve data from metrics_preaggregated_full
+
+        :param key:
```
GeorgeJahad: I would prefer you either complete the param/return comments, or remove them.  To have them half done like this is distracting.

contrib/enum_metrics/config.ini
```diff
@@ -0,0 +1,19 @@
+[localhost]
+cassandra_nodes = localhost
+es_nodes = localhost
+
```
GeorgeJahad: This file seems rax specific.  We try not to check such files in to the open BF repo.

Instead this may be a separate PR for bf-chef.  Pull all but localhost out for now.

GeorgeJahad:
+1 after removing ip addrs.

nice work!

ChandraAddala:
Removed the IP addresses of other environments.

contrib/enum_metrics/README.md
```diff
@@ -0,0 +1,93 @@
+enum_metrics python utilities
+=============================
+
+This project contains some python utilities to work with enum metrics. 
+
+### Setup
+
+Get the [blue flood](https://github.com/rackerlabs/blueflood) repo from github. Execute the following commands
```
izrik: Typo: should be `blueflood`, one word.

contrib/enum_metrics/enum_metrics/errant_enums.py
```diff
@@ -0,0 +1,191 @@
...
+    delete_parser = subparsers.add_parser('delete', help='Delete errant enum')
+
+    delete_parser.add_argument('--dryrun',
+                               action='store_true', help='Display errant enums related data for the given metric name.')
+    delete_parser.add_argument('-m', '--metricName',
+                               required=True, help='metric name to be deleted')
+    delete_parser.add_argument('-t', '--tenantId',
+                               required=True, help='tenantId corresponding to the metric name to be deleted')
+
+    environments = [LOCALHOST, QE01, QE02, STAGING, PROD]
```
izrik: Only one of these environments is present in the `config.ini` file. It would be better to use the 'SafeConfigParser.sections()` method to get the list of available environments, instead of hard-coding them. Also, these choices are rax-specific.

izrik:
Just out of curiosity, why is this in python, whereas the rest of the project is in java?

contrib/enum_metrics/enum_metrics/errant_enums.py
```diff
@@ -0,0 +1,191 @@
...
+import dbclient as db
+import esclient as es
+import config as cf
+
+LOCALHOST = 'localhost'
+QE01 = 'qe01'
+QE02 = 'qe02'
+STAGING = 'staging'
+PROD = 'prod'
+
```
drenalin23: Staging = stage
Don't forget perf01.

contrib/enum_metrics/enum_metrics/errant_enums.py
ChandraAddala: I addressed this in the PR https://github.com/rackerlabs/blueflood/pull/575

contrib/enum_metrics/enum_metrics/errant_enums.py
ChandraAddala: @drenalin23  I just made the script environment agnostic. It basically works with whatever environment we configure in config.ini.

ChandraAddala:
@izrik I started out as a command line script to clean up some data. Python made lot of sense. This project seems to host all these python scripts under contrib folder. Thats why I placed the script over there.

izrik:
OK. I'm all for python. It's easy to program in. Nevertheless, I find multi-language projects are a little more work overall. Is this script only to be executed on-demand, or does it need to be integrated into our CI/CD system?

ChandraAddala:
It will be executed on-demand.

GeorgeJahad:
FYI @izrik, there are two other critical pieces of the project that are in python, in the blueflood directory,, contrib/grinder, and contrib/graphite

PR #566, opened 2015-12-02
CMD-763 - Create Dashboard specifically for enums

usnavi:
Adding the following metrics:
- enums ingested meter
- enums queried meter
- excess enum metrics table read errors meter
- execess enum metrics table write errors meter

### DIFF
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/types/RollupType.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/types/RollupType.java
@@ -24,7 +24,7 @@ public enum RollupType {
             return RollupType.BF_BASIC;
         }
     }
-    
+
     public static RollupType fromRollup(Rollup value) {
         if (value instanceof BluefloodSetRollup)
             return RollupType.SET;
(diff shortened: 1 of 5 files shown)


blueflood-core/src/main/java/com/rackspacecloud/blueflood/types/RollupType.java
```diff
@@ -47,6 +47,41 @@ else if (value instanceof BluefloodEnumRollup) {
...
+     *  Get RollupType from Rollup class type.
+     *
+     * @param rollupTypeClass
+     * @return RollupType
+     */
+    public static RollupType fromRollupTypeClass( Class<? extends Rollup> rollupTypeClass )  {
+
+        if (rollupTypeClass.equals(SimpleNumber.class)
+                || rollupTypeClass.equals( BasicRollup.class)
+                || rollupTypeClass.equals( String.class )) {
```
usnavi: @tilogaat @GeorgeJahad  - I added the check for String.class due to a error in a https://github.com/rackerlabs/blueflood/blob/enums-feature/blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/outputs/handlers/HttpRollupHandlerIntegrationTest.java#L125  

I'm not sure if this is the right thing to do.

tilogaat: It's a reasonable assumption for now.

tilogaat:
+1

PR #567, opened 2015-12-07
added script to init ES

GeorgeJahad:


ChandraAddala:
Looks good. The file seems to be located under src/main/resources. Dont know if it is the best place to keep a script.

GeorgeJahad:
I wanted to keep it close to the json files that it requires.  I'm happy to move it if you know of a better place.

shintasmith:
Does this script need to be mentioned in the Onboarding wiki?

blueflood-elasticsearch/src/main/resources/init-es.sh
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
+#WARNING this will destroy existing ES data and reset it to the proper init state
+function checkFile
```
usnavi: maybe put a little documentation about the script?  or is this in another location?

ChandraAddala:
I cant think of a better place. We can leave the file at src/main/resources as it wont cause any harm to package this inside the jar file.

GeorgeJahad:
Instead of the onboarding doc, I added mention of the script here:
https://github.com/rackerlabs/blueflood/wiki/Setup-ElasticSearch

PR #568, opened 2015-12-07
George/grinder enum perf01

GeorgeJahad:
updated grinder properties files for the various environments;

the staging props now use the grinder endpoints which bypass the repose endpoints


shintasmith:
+1

usnavi:
+1

PR #569, opened 2015-12-08
Fix metadata cache bottleneck

tilogaat:
*What*:
Moving enums validation logic from LocatorFetchRunnable to RollupRunnable 

*Why*:
After deploying the change eac288add478dc2ae03430db2befd325d8169005 in the enums-feature branch we discovered that the time required for LocatorFetchRunnable to finish had increased from 200k miliseconds to 1.3 million miliseconds and the rollup throughput was poor as a result. 
Turns out, since we were narrowing down the locators to be validated using the Metadata cache. We were querying the metadata cache for rollup type of all locators to be rolled up causing the cache to thrash and retrieve data from Cassandra. 

*How*:
The fix in this PR puts the logic of enumvalidation in RollupRunnable because the metadata about rollup types is already retrieved and an extra operation to do the same is not required.

### DIFF
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/RollupRunnable.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/RollupRunnable.java
@@ -32,7 +32,8 @@ import org.slf4j.LoggerFactory;
 import com.rackspacecloud.blueflood.eventemitter.RollupEventEmitter;
 import com.rackspacecloud.blueflood.eventemitter.RollupEvent;
 
-import java.util.HashMap;
+import java.util.*;
+import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 
 /** rolls up data into one data point, inserts that data point. */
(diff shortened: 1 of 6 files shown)


blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/RollupRunnable.java
```diff
@@ -59,13 +62,14 @@
         }
     }
 
-    public RollupRunnable(RollupExecutionContext executionContext, SingleRollupReadContext singleRollupReadContext, RollupBatchWriter rollupBatchWriter) {
+    public RollupRunnable(RollupExecutionContext executionContext, SingleRollupReadContext singleRollupReadContext, RollupBatchWriter rollupBatchWriter, ThreadPoolExecutor enumValidatorExecutor) {
```
usnavi: I noticed that child classes are calling this constructor with the ``ThreadPoolExecutor enumValidatorExecutor`` parameter taking a ``null``.  I assumed because the executor isn't used as the ``run()`` method is overridden.  Do these classes need to subclass this class?  Would it be clearer to have this and its sub-classes all implement a common interface rather than sub-class ``RollupRunnable``?  Just curious.

tilogaat: Do you mean ```HistogramRollupRunnable``` ? I think that's the only subclass of RollupRunnable that I found. Yes I think we can refactor these 2 classes to conform to a common interface. I hate passing around null params in the constructor too.

usnavi:
+1, I'll merge this into enums-feature.

PR #565, opened 2015-12-02
Goru97/grafana annotation support

goru97:
*Why:*
Support for annotations in Grafana.

PR #570, opened 2015-12-08
Goru97/grafana annotations support1

goru97:
*What:*
Using graphite-api to talk to blueflood for annotations, till the time CORS issue is fixed by Identity.


contrib/graphite/blueflood.py
```diff
@@ -51,6 +51,7 @@ def calc_res(start, stop):
```
GeorgeJahad: it looks like this is the same as the commit in this PR: https://github.com/rackerlabs/blueflood/pull/565

GeorgeJahad:
+1

contrib/graphite/blueflood.py
goru97: I didn't really get that. Those commits are for finder.

PR #571, opened 2015-12-08
Introduced a config value to enable/disable EnumValidation

tilogaat:
*What:*
* Introduced a config value to enable/disable EnumValidation
* Fixed a bug where the enum rollups meter was counting only the locators getting validated, We need for it to track all enum locators getting rolled up.

### DIFF
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/RollupRunnable.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/RollupRunnable.java
@@ -107,10 +107,12 @@ public class RollupRunnable implements Runnable {
             Granularity dstGran = srcGran.coarser();
             ColumnFamily<Locator, Long> dstCF = CassandraModel.getColumnFamily(rollupClass, dstGran);
 
-            //Run the validation for enums every 5 minutes, when data is being rolledup from full to 5m
-            if (rollupType == RollupType.ENUM && dstGran.equals(Granularity.MIN_5)) {
+            if (rollupType == RollupType.ENUM) {
                 singleRollupReadContext.getEnumMetricsMeter().mark();
-                enumValidatorExecutor.execute(new EnumValidator(Sets.newHashSet(rollupLocator)));
+                //Run the validation for enums every 5 minutes, when data is being rolled up from full to 5m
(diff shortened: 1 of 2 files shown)


blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/RollupRunnable.java
```diff
@@ -107,10 +107,12 @@ public void run() {
             Granularity dstGran = srcGran.coarser();
             ColumnFamily<Locator, Long> dstCF = CassandraModel.getColumnFamily(rollupClass, dstGran);
 
-            //Run the validation for enums every 5 minutes, when data is being rolledup from full to 5m
-            if (rollupType == RollupType.ENUM && dstGran.equals(Granularity.MIN_5)) {
+            if (rollupType == RollupType.ENUM) {
                 singleRollupReadContext.getEnumMetricsMeter().mark();
```
GeorgeJahad: did you want to continue sending metrics even when it is configured out?

GeorgeJahad:
+1

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/RollupRunnable.java
tilogaat: Yes, because we want to track all the enums that are being rolled up. I think I should make it better and track enum rollups at each granularity.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/RollupRunnable.java
```diff
@@ -107,10 +107,12 @@ public void run() {
...
-                enumValidatorExecutor.execute(new EnumValidator(Sets.newHashSet(rollupLocator)));
+                //Run the validation for enums every 5 minutes, when data is being rolled up from full to 5m
+                if (dstGran.equals(Granularity.MIN_5) && Configuration.getInstance().getBooleanProperty(CoreConfig.ENUM_VALIDATOR_ENABLED) == true) {
```
shintasmith: This is just for my own understanding.. Why would anyone want to turn off validation for enum?

GeorgeJahad: There is no direct connection between validations and rollups.  We added to rollups out of convenience, but it increases the load on rollups.  If the rollup load gets too great, we may decide to move it elsewhere, and may want to turn it off temporarily here.

PR #573, opened 2015-12-09
Adding meters for enums per granularity

tilogaat:
*What:*
Changed rollup meter of enums to be per destination granularity to provide better insight into enum locators getting rolled up at each granularity.


GeorgeJahad:
+1

PR #575, opened 2015-12-11
Enums feature

ChandraAddala:
[CMD-714] Python script to clear errant enums.
     - Removing pre-configured environments as they are company specific
     - Environments are dynamic based on the ones configured in config.ini


izrik:
lgtm +1

PR #576, opened 2015-12-11
Modify blueflood Tracker to get tenantId via URI

usnavi:
For the Tracker, pull tenant id from UIR rather than a HTTP header.

### DIFF
--- a/blueflood-http/src/main/java/com/rackspacecloud/blueflood/tracker/Tracker.java
+++ b/blueflood-http/src/main/java/com/rackspacecloud/blueflood/tracker/Tracker.java
@@ -31,11 +31,14 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 public class Tracker implements TrackerMBean {
 
     private static final Logger log = LoggerFactory.getLogger(Tracker.class);
     private static final String trackerName = String.format("com.rackspacecloud.blueflood.tracker:type=%s", Tracker.class.getSimpleName());
+    private static final Pattern patternGetTid = Pattern.compile( "/v\\d+\\.\\d+/([^/]+)/.*" );
 
     static Set tenantIds = new HashSet();
     static boolean isTrackingDelayedMetrics = false;
@@ -95,7 +98,8 @@ public class Tracker implements TrackerMBean {
         // check if tenantId is being tracked by JMX TenantTrackerMBean and log the request if it is
         if (request == null) return;
 
-        String tenantId = request.getHeader("tenantId");
+        String tenantId = findTid( request.getUri() );
+
         if (isTracking(tenantId)) {
 
             // get headers
@@ -121,6 +125,16 @@ public class Tracker implements TrackerMBean {
 
             log.info(logMessage);
         }
+    }
+
+    static String findTid( String uri ) {
+
+        Matcher m = patternGetTid.matcher( uri );
+
+        if( m.matches() )
+            return m.group( 1 );
+        else
+            return null;
     }
 
     public static void trackDelayedMetricsTenant(String tenantid) {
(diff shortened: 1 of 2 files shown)


blueflood-http/src/main/java/com/rackspacecloud/blueflood/tracker/Tracker.java
```diff
@@ -123,6 +127,16 @@ public static void track(HttpRequest request) {
...
     }
 
+    static String findTid( String uri ) {
+
+        Matcher m = patternGetTid.matcher( uri );
+
+        if( m.matches() )
+            return m.group( 1 );
+        else
+            return null;
```
ChandraAddala: Unless you have strong opinion about this, can you return empty string instead of null?

blueflood-http/src/test/java/com/rackspacecloud/blueflood/tracker/TrackerTest.java
```diff
@@ -31,35 +35,53 @@ public void testDoesNotAddTenantTwice() {
...
         tracker.removeAllTenants();
-        Assert.assertFalse("tenant " + testTenant1 + " not removed", tracker.isTracking(testTenant1));
-        Assert.assertFalse("tenant " + testTenant2 + " not removed", tracker.isTracking(testTenant2));
+        assertFalse( "tenant " + testTenant1 + " not removed", tracker.isTracking( testTenant1 ) );
+        assertFalse( "tenant " + testTenant2 + " not removed", tracker.isTracking( testTenant2 ) );
 
         Set tenants = tracker.getTenants();
-        Assert.assertEquals("tenants.size not 0", tenants.size(), 0);
+        assertEquals( "tenants.size not 0", tenants.size(), 0 );
+    }
```
ChandraAddala: Can you add a test with "/" at the end?

blueflood-http/src/main/java/com/rackspacecloud/blueflood/tracker/Tracker.java
usnavi: The original code returned null.  Does it matter in this case?

blueflood-http/src/test/java/com/rackspacecloud/blueflood/tracker/TrackerTest.java
usnavi: done!

blueflood-http/src/main/java/com/rackspacecloud/blueflood/tracker/Tracker.java
shintasmith: I think the method isTracking(tenantId) does not check for null. So it's possible you will get NPE.

usnavi: Maps can have null keys, for some goofy reason, so Map.contains( null ) won't throw and exception.

I'm fine changing it if its for the best, just didn't want to change behavior if not necessary.

shintasmith: Oh, I read the code wrong... I thought it was tenantId.someMethod() so if tenantId is null, you get NPE. But it's actually tenantId*s*. Never mind... Then I don't think there's a point of returning empty string.

shintasmith:
I am fine with the changes. :+1: It's the right way to do it in the long run. 

I do have to point out that we don't validate the tenantId in the URI today (which I think we need to fix soon), so we have to trust ppl use the right tenantId.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/tracker/Tracker.java
ChandraAddala: It is more of a best practice. Dont know how relevant it is to this scenario. We can keep the old behavior as it is.

ChandraAddala:
+1

drenalin23:
Shinta -

I think the cookbook I forked from MaaS includes URI validation in their repose setup - so we should get this with the upgrade (at the repose level).
Thx
Scott

PR #577, opened 2015-12-12
George/fix mplot queries

GeorgeJahad:
This PR fixes the exception generated by enum mplot queries:
java.lang.Exception: average not supported for this type: BluefloodEnumRollup

Multiplot enum and gauge queries were not treating default filterStats the same as singlePlot queries.  This PR refactors the code so singlePlot and multiPlot work the same.

Also added another logmessage for rollups.

### DIFF
--- a/blueflood-http/src/main/java/com/rackspacecloud/blueflood/outputs/serializers/FakeMetricDataGenerator.java
+++ b/blueflood-http/src/main/java/com/rackspacecloud/blueflood/outputs/serializers/FakeMetricDataGenerator.java
@@ -96,6 +96,17 @@ public class FakeMetricDataGenerator {
         return points;
     }
     
+    public static Points<BluefloodEnumRollup> generateFakeEnumRollupPoints() {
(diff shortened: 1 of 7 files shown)


blueflood-http/src/main/java/com/rackspacecloud/blueflood/outputs/serializers/FakeMetricDataGenerator.java
```diff
@@ -96,6 +96,17 @@
         return points;
     }
     
+    public static Points<BluefloodEnumRollup> generateFakeEnumRollupPoints() {
+        Points<BluefloodEnumRollup> points = new Points<BluefloodEnumRollup>();
+        long startTime = 1234567L;
+        for (int i = 0; i < 5; i++) {
```
shintasmith: I am a bit confused why we are generating fake enum rollup points. Is this method only being called in Test classes? If so, should we move it to some class in ```src/test/java``` ?

shintasmith:
The rest of the changes look good, to a novice pair of eyes. ;-) +1. Someone else should probably review too.

PR #578, opened 2015-12-14
Izrik/maven versioning

izrik:
This PR modifies pom files and adds a script to build the code with a release-style (not a snapshot) version string. 

The build script could be run from either jenkins or travis-ci. It should work with both. Here's an example of it in action: https://jenkins.bf.k1k.me/jenkins/job/izrik-clone-blueflood-verify-master/15/console

I haven't made any configurations changes to our jenkins jobs. They will still run maven in the usual way and will ignore the new build script. Changing them to use the build script will be a part of a future story. Same goes for `.travis.yml`.

The script has a place to tag the commit and push said tag to the repo. Before the script is run by a CI/CD system, proper credentials for pushing tags to the repo will have to be configured. I didn't have much trouble settings that up in Jenkins in my prototype jobs, but it is a "gotcha" to be aware of. Setting up said credentials will be a part of the future story to turn on the script.

As far as I know, everything in  the build process should stay exactly the same if this PR is merged. Travis-CI, jenkins, and local dev machines will still produce `2.0.0-SNAPSHOT` artifacts.

Because the `<version>` in the pom is now an expression instead of a hard-coded value, we won't be able to use the maven release plugin to produce releases. But that's ok, since we weren't using it anyways.

### DIFF
--- a/build_and_release.rb
+++ b/build_and_release.rb
@@ -9,6 +9,15 @@
 #
 # Running maven from the command line (e.g. `mvn package`), instead of the
 # script, will still produce artifacts with the SNAPSHOT version string.
+#
+# To produce a snapshot build of the software, run the following at the command
+# line:
+#   mvn package
+#
+# To produce a release build of the software, run the following:
+#   ruby build_and_release VERSION_SUFFIX
+# where VERSION_SUFFIX is a number.
+
 
 require 'rexml/document'
 
--- a/build_and_release.rb
+++ b/build_and_release.rb
@@ -46,12 +46,33 @@ def run_command(command)
   end
 end
 
+def run_command_return_lines(command)
+  lines = []
+  ret = IO.popen(command, :err=>[:child, :out]) {|io|
+    io.each {|line|
+      lines << line
+    }
+  }
+  return lines
+end
+
+def pom_evaluate(property_name)
+  lines = run_command_return_lines("mvn org.apache.maven.plugins:maven-help-plugin:evaluate -Dexpression=#{property_name}")
+  return lines.find {|line| not line.start_with?("[")}.chomp
+end
+
+
 
 if ARGV.length < 1
   $stderr.puts "Usage: #{$0} BUILD_NUMBER"
   exit 1
 end
 
+pom_version = pom_evaluate("project.version")
+if not pom_version.end_with?("-SNAPSHOT")
+  puts "The project's effective version doesn't end with \"-SNAPSHOT\""
+  exit 1
+end
 
 # Get the new version, as MAJOR.MINOR.BUILD
 build_number = ARGV.shift
--- a/build_and_release.rb
+++ b/build_and_release.rb
@@ -78,6 +78,14 @@ end
 build_number = ARGV.shift
 new_version = get_version_number(build_number)
 
+# Check the branch that we're on. if not master, add the branch name and treat as snapshot
+branches = run_command_return_lines("git branch --no-color --contains")
+branch = branches.find {|line| line.start_with?("*")}.sub(/^\*\s*/, '').strip()
+if branch != "master"
+  branch.sub!(/_/, '__')
+  branch.sub!(/[\\\/]/,'_')
+  build_number += "-" + branch + "-SNAPSHOT"
+end
 
 
 # build the project
--- a/build_and_release.rb
+++ b/build_and_release.rb
@@ -89,12 +89,14 @@ end
 
 # Get the new version, as MAJOR.MINOR.BUILD
 build_number = options.shift
-new_version = get_version_number(build_number)
+
+is_release = true
 
 # Check the branch that we're on. if not master, add the branch name and treat as snapshot
 branches = run_command_return_lines("git branch --no-color --contains")
 branch = branches.find {|line| line.start_with?("*")}.sub(/^\*\s*/, '').strip()
 if branch != "master"
+  is_release = false
   branch.sub!(/_/, '__')
   branch.sub!(/[\\\/]/,'_')
   build_number += "-" + branch + "-SNAPSHOT"
@@ -106,8 +108,10 @@ run_command("mvn clean")
 run_command("mvn verify -DBUILD_SUFFIX=#{build_number} -P cassandra-2.0 findbugs:findbugs")
 
 
-if tag_on_success
+if tag_on_success and is_release
   # tag the commit so we can associate it with the release
-  run_command("git tag release-#{new_version}")
-  run_command("git push --tags")
+  new_version = get_version_number(build_number)
+  tag_name = "release-#{new_version}"
+  run_command("git tag #{tag_name}")
+  run_command("git push origin #{tag_name}:refs/tags/#{tag_name}")
 end
--- a/build_and_release.rb
+++ b/build_and_release.rb
@@ -1,0 +1,50 @@
+
+require 'rexml/document'
+
+
+def get_version_number(build_number)
+  pom = REXML::Document.new(File.new("pom.xml"))
+  version = pom.elements["project/properties/BLUEFLOOD_VERSION"].text
+
+  if version.end_with? '${VERSION_SUFFIX}'
+    return version.gsub('${VERSION_SUFFIX}', build_number)
+  end
+
+  return version
+end
+
+def run_command(command)
+  puts "Running command: #{command}"
+  ret = IO.popen(command, :err=>[:child, :out]) {|io|
+    io.each {|line|
+      puts ">> #{line}"
+    }
+  }
+  if $?.exitstatus != 0
+    puts "Failed with code #{$?.exitstatus}"
+    exit 1
+  end
+end
+
+
+if ARGV.length < 1
+  $stderr.puts "Usage: #{$0} BUILD_NUMBER"
+  exit 1
+end
+
+
+# Get the new version, as MAJOR.MINOR.BUILD
+build_number = ARGV.shift
+new_version = get_version_number(build_number)
+
+
+
+# build the project
+run_command("mvn clean")
+run_command("mvn verify -DBUILD_SUFFIX=#{build_number} -P cassandra-2.0 findbugs:findbugs")
+
+
+
+# # tag the commit so we can associate it with the release
+# run_command("git tag release-#{new_version}")
+# run_command("git push --tags")
--- a/build_and_release.rb
+++ b/build_and_release.rb
@@ -76,7 +76,7 @@ end
 
 options = parser.parse!
 
-if options < 1
+if options.length < 1
   $stderr.puts parser.help
   exit 1
 end
--- a/build_and_release.rb
+++ b/build_and_release.rb
@@ -20,6 +20,7 @@
 
 
 require 'rexml/document'
+require 'optparse'
 
 
 def get_version_number(build_number)
@@ -63,8 +64,20 @@ end
 
 
 
-if ARGV.length < 1
-  $stderr.puts "Usage: #{$0} BUILD_NUMBER"
+
+tag_on_success = false
+parser = OptionParser.new do |o|
+  o.banner = "Usage: #{$0} [--[no-]tag-on-success] BUILD_NUMBER"
+
+  o.on('--[no-]tag-on-success') do |value|
+    tag_on_success = value
+  end
+end
+
+options = parser.parse!
+
+if options < 1
+  $stderr.puts parser.help
   exit 1
 end
 
@@ -75,7 +88,7 @@ if not pom_version.end_with?("-SNAPSHOT")
 end
 
 # Get the new version, as MAJOR.MINOR.BUILD
-build_number = ARGV.shift
+build_number = options.shift
 new_version = get_version_number(build_number)
 
 # Check the branch that we're on. if not master, add the branch name and treat as snapshot
@@ -93,7 +106,8 @@ run_command("mvn clean")
 run_command("mvn verify -DBUILD_SUFFIX=#{build_number} -P cassandra-2.0 findbugs:findbugs")
 
 
-
-# tag the commit so we can associate it with the release
-run_command("git tag release-#{new_version}")
-run_command("git push --tags")
+if tag_on_success
+  # tag the commit so we can associate it with the release
+  run_command("git tag release-#{new_version}")
+  run_command("git push --tags")
+end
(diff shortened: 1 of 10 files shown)


GeorgeJahad:
+1 lgtm

pom.xml
```diff
@@ -42,6 +42,8 @@
   <url>http://blueflood.io</url>
 
   <properties>
+    <VERSION_SUFFIX>0-SNAPSHOT</VERSION_SUFFIX>
+    <BLUEFLOOD_VERSION>2.0.${VERSION_SUFFIX}</BLUEFLOOD_VERSION>
     <maven.javadoc.skip>true</maven.javadoc.skip>
```
usnavi: Why have a VERSION_SUFFIX and BLUEFLOOD_VERSION?  Why not just have 1 variable for the version?

ChandraAddala:
@izrik Do we plan to use "maven release plugin" in the future? We use this script till then?

shintasmith:
@izrik, I am hesitant to have this BLUEFLOOD_VERSION and the ruby script that swizzles it. It seems to be diverging from the standard Maven-way. In my past experience, once we diverge, its hard to do things the normal way. 

But I understand the need for CI/CD. I am open to not using maven-release-plugin. However, instead of writing custom script, have you check this out?
http://www.mojohaus.org/buildnumber-maven-plugin/

Someone's write up:
http://www.jayway.com/2012/04/07/continuous-deployment-versioning-and-git/

I admit, I haven't spent much time on this other than a few minutes of reading here and there. So I'm curious to hear your take on this.

build_and_release.rb
```diff
@@ -0,0 +1,61 @@
...
+
+# build the project
+run_command("mvn clean")
+run_command("mvn verify -DBUILD_SUFFIX=#{build_number} -P cassandra-2.0 findbugs:findbugs")
+
+
+
+# tag the commit so we can associate it with the release
+run_command("git tag release-#{new_version}")
+run_command("git push --tags")
```
ChandraAddala: In the maven-release-plugin way of doing things, before it cuts a release it verifies some things. I am thinking those still apply even if we do it some other way.

For ex:
Check that there are no SNAPSHOT dependencies

pom.xml
izrik: They serve separate (bu related) purposes. The VERSION_SUFFIX is what gets overridden by the build script. The BLUEFLOOD_VERSION lets us modify the version for all submodules in a single place.

izrik:
@ChandraAddala We can't use the maven-release-plugin for a number of reasons, among them that the plugin adds commits and pushes hem to the repo. See more here: https://one.rackspace.com/display/cloudmetrics/Maven+Versioning#MavenVersioning-HowMavenfitswithCI/CD
We would use the script eventually. Currently, the script isn't run, neither by Jenkins nor by Travis-CI. First we need to harmonize the configuration of our Jenkins jobs. See here: https://jira.rax.io/browse/CMD-847

build_and_release.rb
izrik: I'll look into that. Maybe we can use some of the goals from the release plugin to do the checks. That would be fine.

izrik:
@shintasmith I'll look into that plugin.

izrik:
The main drawback to this approach, that I can think of, is that it only works properly for `master`. I'll have to think about it some more and figure out what to do about branch builds and PR builds.

shintasmith:
@izrik, branch builds and PR builds should not trigger this release-versioning thing, right?

izrik:
@shintasmith In theory, sure. The problem is that sometimes a dev will want to upload an in-flight version of the software to a testing environment, from their feature branch. I had intended that deployments could eventually use the [maven plugin for chef](https://github.com/chef-cookbooks/maven) to download a specific version of the software to the environment (instead of using cloud files). But if only builds on `master` produce a release build that can be deployed, that throws a wrench in the works. I'll have to figure out some other way to get the artifacts built from a branch and uploaded to a testing env.

Probably, the thing to do would be to only allow release builds on staging and production. Other envs would still be able to use snapshot builds. But that will make things a little more complicated.

shintasmith:
Ah I see what you mean. Yes, this will complicate things.

ChandraAddala:
@izrik @shintasmith 

I was researching a little bit on this and I ran into this. Basically creates a separate release branch for each build.

http://stackoverflow.com/questions/18456111/what-is-the-maven-way-for-automatic-project-versions-when-doing-continuous-deliv

1) Jenkins clones the git repo with the source code, and say the source code has version 1.0-SNAPSHOT
2) Jenkins creates a git branch called 1.0.JENKINS-JOB-NUMBER so the snapshot version is turned into a real version 1.0.124
3) Jenkins invokes the maven versions plugin to change the version number in the pom.xml files from 1.0-SNAPSHOT to 1.0.JENKINS-JOB-NUMBER
4) Jenkins invokes mvn install
5) If the mvn install is a success then Jenkins will push the branch 1.0.JENKINS-JOB-NUMBER and a real non-snapshot version is created with a proper tag in git to reproduce later. If the mvn install fails then Jenkins will just delete the newly created branch and fail the build.

build_and_release.rb
izrik: @ChandraAddala We can use the maven-enforcer-plugin to check all dependencies for snapshots: https://maven.apache.org/enforcer/enforcer-rules/requireReleaseDeps.html

build_and_release.rb
```diff
@@ -0,0 +1,61 @@
...
+# script, will still produce artifacts with the SNAPSHOT version string.
+
+require 'rexml/document'
+
+
+def get_version_number(build_number)
+  pom = REXML::Document.new(File.new("pom.xml"))
+  version = pom.elements["project/properties/BLUEFLOOD_VERSION"].text
+
+  if version.end_with? '${VERSION_SUFFIX}'
```
ChandraAddala: Also if the version does not end with SNAPSHOT, you should raise an exception/failure or something.

pom.xml
ChandraAddala: We can avoid having two variables by doing something similar to the command below. We should not checkin the files changed. After we change versions we can run the "mvn package" command and tag the release.

```
mvn build-helper:parse-version versions:set -DnewVersion=${parsedVersion.majorVersion}.\${parsedVersion.minorVersion}.\${parsedVersion.incrementalVersion}-${BUILD_NUMBER}
```

I found this link helpful which is similar to what you are doing. https://www.thoughtworks.com/insights/blog/continuous-delivery-maven-and-go-maven-central

izrik:
@ChandraAddala I'll look into that. It sounds like a stripped down equivalent of what the maven-release plugin does.

build_and_release.rb
izrik: Will do.

pom.xml
izrik: That was option #2 in my [write-up](https://one.rackspace.com/pages/viewpage.action?title=Maven+Versioning&spaceKey=cloudmetrics#MavenVersioning-ProposedSolutions). I decided not to go that route because it requires editing the pom files. Editing them then leaves us with the question of what to do with the edited files: 
* We can't commit them to master, because we're trying to run this off of new commits to master, which would cause an infinite loop
* We could create a commit as part of a release-specific branch, but we don't really support old versions of the software. We don't backport bug fixes or anything. We just generally forget about them.
* We could create a commit and then just tag it. This doesn't really buy us anything either. It just leaves a lot of dangling, tagged commits in the repo.

ChandraAddala: Sorry for the long message. I thought of talking to you but this way, others can weigh in.

I noticed an issue with the current approach. When using a property(BLUEFLOOD_VERSION) for 'version', it doesnt seem to play well with maven. It seems to generate the warning below when I tried to run 'mvn clean install'. I do not know the implications of the warning. I am using maven 3.3.3.

```
[WARNING] Some problems were encountered while building the effective model for com.rackspacecloud:blueflood-core:jar:2.0.0-SNAPSHOT
[WARNING] 'version' contains an expression but should be a constant. @ com.rackspacecloud:blueflood:${BLUEFLOOD_VERSION}, /Users/chan5120/Documents/workspacecf2/blueflood/pom.xml, line 26, column 14
```
Also this kinda version representation seems a little odd for a typical maven project.
```
<VERSION_SUFFIX>0-SNAPSHOT</VERSION_SUFFIX>
<BLUEFLOOD_VERSION>2.0.${VERSION_SUFFIX}</BLUEFLOOD_VERSION>
```

I was trying to avoid both and see if it is still possible to generate a release with a build number and I have the approach as outlined below. 

1) Lets leave the pom.xml's the way they are. This way mvn can be run directly without the build script.
2) We can use ``mvn versions:set`` to change versions of pom.xml in all the modules to have 2.0.BUILD_NUMBER. We will **NOT** commit the changed pom files. 
3) We do the ``mvn package -P all-modules`` to get released version of jar file(2.0.BUILD_NUMBER)
4) tag the release.

For STEP 2, we do have the option of creating a separate release branch with name '2.0.BUILD_NUMBER', commit the changed pom.xml in that branch for future reference, if we desire. Its upto us. 

We do have to change the version of pom.xml's dynamically. In your build script, you are doing it by passing a paramenter BUILD_NUMBER, and in the above approach, I am using ``mvn versions:set`` command to set version with BUILD_NUMBER.

CONs:
When we are ready for the next build. clean the directory to remove the files created by ``mvn versions:set`` from previous run and we repeat the steps. Can be done with something like ``git clean -f -d``. In Jenkins we can check 'Delete workspace before build starts' to address this.

izrik:
@shintasmith I looked into the maven build number plugin. I couldn't get the latest version to work at all. I don't think it's necessary, either, since we already get build numbers from both jenkins and travis. However, it may be useful for building on branches, if I can get it to work.

build_and_release.rb
izrik: @ChandraAddala 
Now that I look at it, I'm not sure what you mean. The default value of `BLUEFLOOD_VERSION` is `2.0.${VERSION_SUFFIX}`. The `VERSION_SUFFIX` part is provided by the build script (which leads to a release build), or is left as the default of `0-SNAPSHOT` (which leads to a snapshot). There are cases when both with and without the `-SNAPSHOT` suffix is acceptable. Are you proposing throwing an exception when `VERSION_SUFFIX` has the snapshot suffix in it? Because that property will already be overwritten just by running the build script.

ChandraAddala: @izrik I was thinking from outside the script. The script takes a maven project with x.x.x-SNAPSHOT as version and produces artifacts with x.x.${BUILD_NUMBER}. What if the version of the project is not SNAPSHOT in the first place? Shouldnt it fail? Its just a precaution when someone manually changes BLUEFLOOD_VERSION to not have -SNAPHOT and tried to run this script. For ex: If someone cuts a release on project using release plugin, I think it would fail if the underlying project does not have SNAPSHOT version.

izrik: @ChandraAddala I think I understand what you mean. You're talking about the case when someone manually changes the value of `BLUEFLOOD_VERSION` in the pom itself, right? So something like this:
```xml
<BLUEFLOOD_VERSION>2.0.1</BLUEFLOOD_VERSION>
```
In that case, it would be good for the build script to check the format of the variable. Is that what you're driving at?

ChandraAddala: Yes.

izrik: Done.

izrik:
@shintasmith Previously, you said:
> I am hesitant to have this BLUEFLOOD_VERSION and the ruby script that swizzles it. It seems to be diverging from the standard Maven-way. In my past experience, once we diverge, its hard to do things the normal way

I've been trying to figure out a more maven-ish way to do this. Unfortunately, I keep coming up empty. I can't find an elegant way to get maven to do the complex logic necessary to do everything that the ruby build script does (get build number from args, check the version in the current pom for "-SNAPSHOT", add the branch name if we're not on master). About the only way I that I can think of to do that woudl be to write a custom maven plugin.

The purpose of the build script is just to build the software in a CI/CD environment, i.e. jenkins or travis. Anyone doing dev work can just use `mvn` or an IDE as they currently are, without any changes to their workflow. I'm trying to avoid such changes when I use the properties with their default values. If we were to import the custom CI/CD logic into the pom file, I can't guarantee that that wouldn't require changes to the devs' usual workflow.

I understand that writing our own build script from scratch is not easily maintainable in the long term. What you think about using a `build.gradle` file instead? I can get gradle to run maven, I think.

usnavi:
I haven't been following this very closely since @izrik started investigating options.  I'm all for looking into gradle (a newer tool might support newer workflows?)  Perhaps do we want to have a quick meeting early next week around this?

izrik:
@usnavi Just to be clear: I'm not suggesting we switch the whole project to gradle. Rather I'm suggesting a gradle build script to invoke maven, instead of the `build_and_release.rb` script I've written so far. This new gradle build script would only be used by Jenkins and/or Travis-CI builds.

shintasmith:
Yeah... In a way, maven is just not conducive to the way we want to do CI/CD. I am also open to moving to gradle, even though I'm not that crazy about gradle as a build system. But we should probably discuss in a larger team setting, as this is the open source component. There might be people outside Rackspace impacted.

If we can't completely switch to gradle, then lets go with the Ruby script until maven gives us enough grief. :-)

usnavi:
@izrik  - I understand now.  No need to use gradle for this little bit.  Man, what a mess.

izrik:
Rebase'd.

izrik:
I'm getting some odd test failure. I'm going to dig into this. I won't merge the PR yet.

pom.xml
ChandraAddala: @izrik Were you able to look into the warning?

```
[WARNING] Some problems were encountered while building the effective model for com.rackspacecloud:blueflood-core:jar:2.0.0-SNAPSHOT
[WARNING] 'version' contains an expression but should be a constant. @ com.rackspacecloud:blueflood:${BLUEFLOOD_VERSION}, /Users/chan5120/Documents/workspacecf2/blueflood/pom.xml, line 26, column 14
```

izrik:
The test failure was on my end. This PR is good to merge.
@GeorgeJahad @shintasmith @ChandraAddala @usnavi Any other concerns before I merge it?

usnavi:
lgtm

pom.xml
ChandraAddala: @izrik ^

PR #581, opened 2015-12-23
add errant_ttl python script

VinnyQ:
to detect and repair db records with ttl = null or greater than x days.

Please review but DO NOT MERGE.  We will recreate a PR and merge to master after enums-feature is merged to master.

Once I received the +1 I'll package and deploy just the python script to prod to run analysis and repairs as needed.

Usage of script:
```
usage: errant_ttl.py [-h] {show,repair} ...

show and repair cassandra records with errant ttl, e.g. null

positional arguments:
  {show,repair}  commands
    show         get all records in a table with errant ttl
    repair       repair all records in a table with errant ttl

usage: errant_ttl.py show [-h] [-e {localhost}] -cf COLUMNFAMILY [-l LIMIT]
                          [-ttl TTLINSECONDS] [-v]

optional arguments:
  -h, --help            show this help message and exit
  -e {localhost}, --env {localhost}
                        Environment we are pointing to
  -cf COLUMNFAMILY, --columnFamily COLUMNFAMILY
                        name of column family (table) to retrieve rows from
  -l LIMIT, --limit LIMIT
                        limit number of records to retrieve at a time for
                        analyzing; default to 500, use 0 for no limit;
  -ttl TTLINSECONDS, --ttlInSeconds TTLINSECONDS
                        the ttl threshold in seconds to compare against,
                        default to 157680000 seconds (5 years)
  -v, --verbose

usage: errant_ttl.py repair [-h] [-e {localhost}] -cf COLUMNFAMILY [-l LIMIT]
                            [-ttl TTLINSECONDS] [-v] [--dryrun]

optional arguments:
  -h, --help            show this help message and exit
  -e {localhost}, --env {localhost}
                        Environment we are pointing to
  -cf COLUMNFAMILY, --columnFamily COLUMNFAMILY
                        name of column family (table) to retrieve rows from
  -l LIMIT, --limit LIMIT
                        limit number of records to retrieve at a time for
                        analyzing; default to 500, use 0 for no limit;
  -ttl TTLINSECONDS, --ttlInSeconds TTLINSECONDS
                        the ttl threshold in seconds to compare against,
                        default to 157680000 seconds (5 years)
  -v, --verbose
  --dryrun

optional arguments:
  -h, --help     show this help message and exit
```

### DIFF
--- a/contrib/enum_metrics/enum_metrics/errant_ttl.py
+++ b/contrib/enum_metrics/enum_metrics/errant_ttl.py
@@ -1,0 +1,152 @@
+import sys
+import argparse
+import json
+import dbclient as db
+import esclient as es
+import config as cf
+from datetime import datetime
+
+LOCALHOST = 'localhost'
+PROGRESS_MARKER_NUM = 1000000
+errant_ttl_in_seconds = 157680000  # 5 years
+is_verbose = False
+
+
+def parse_arguments(args):
+    """Parses the supplied arguments"""
+
+    parser = argparse.ArgumentParser(prog="errant_ttl.py", description='show and repair cassandra records with errant ttl, e.g. null')
+
+    subparsers = parser.add_subparsers(help='commands')
+
+    show_errant_ttl = subparsers.add_parser('show', help='get all records in a table with errant ttl')
+    repair_errant_ttl = subparsers.add_parser('repair', help='repair all records in a table with errant ttl')
+
+    for p in [show_errant_ttl, repair_errant_ttl]:
+        p.add_argument('-e', '--env', choices=cf.Config.get_environments(),
+                           default=LOCALHOST, help='Environment we are pointing to')
+        p.add_argument('-cf', '--columnFamily',
+                           required=True, help='name of column family (table) to retrieve rows from')
+        p.add_argument("-l", "--limit", required=False, default=500,
+                            help='limit number of records to retrieve at a time for analyzing; default to 500, use 0 for no limit;')
+        p.add_argument("-ttl", "--ttlInSeconds", required=False, help='the ttl threshold in seconds to compare against, default to 157680000 seconds (5 years)')
+        p.add_argument("-v", "--verbose", action="store_true")
+
+    repair_errant_ttl.add_argument('--dryrun', action='store_true')
+
+    args = parser.parse_args()
+
+    if args.ttlInSeconds:
+        global errant_ttl_in_seconds
+        errant_ttl_in_seconds = int(args.ttlInSeconds)
+
+    global is_verbose
+    if args.verbose: is_verbose = True
+    if is_verbose: print "args: " + str(args)
+
+    return args
+
+
+def get_rows_with_ttl(db_client, args):
+    """
+    return all records with tll for columnFamily
+    """
+    cqlstr = "SELECT key, column1, value, ttl(value) as ttl_val FROM " + args.columnFamily
+    if int(args.limit) > 0: cqlstr += " limit " + str(args.limit)
+
+    if is_verbose: print "cqlstr: " + cqlstr
+    results = ()
+    try:
+        prepared_stmt = db_client.session.prepare(cqlstr)
+        results = db_client.session.execute(prepared_stmt)
+    except Exception as e:
+        print "Error (" + type(e).__name__ + " Exception): " + e.message
+
+    return results
+
+
+def delete_errant_ttl(db_client, args, record):
+    """
+    delete record where key,column1
+    """
+    key = str(record[0])
+    column1 = str(record[1])
+
+    cqlstr = "DELETE FROM %s where key=%r and column1=%s;" % (args.columnFamily, key, column1)
+    if is_verbose: print "cqlstr: " + cqlstr
+
+    success = False
+    if not args.dryrun:
+        try:
+            prepared_stmt = db_client.session.prepare(cqlstr)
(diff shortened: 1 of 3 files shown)


contrib/enum_metrics/enum_metrics/errant_ttl.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+def update_errant_ttl(db_client, args, record):
+    """
+    update ttl of record by inserting new key,column1,value with new ttl
+    """
+    key = str(record[0])
+    column1 = str(record[1])
+    value = str(record[2])
+    cqlstr = "INSERT INTO %s (key, column1, value) values(%r, %s, %r) using ttl %s;" % (args.columnFamily, key, column1, value, str(ERRANT_TTL_IN_SECONDS))
```
ChandraAddala: I am not too familiar with cassandra but shouldnt this be update?

contrib/enum_metrics/enum_metrics/errant_ttl.py
```diff
@@ -0,0 +1,120 @@
...
+    if args.dryrun:
+        print "update ttl in " + args.columnFamily + " for " + key + ", " + column1 + ", " + value
+    else:
```
ChandraAddala: Running 'update' command in 'dryrun' mode and running the 'show' command seem to do the same thing. May be 'show' command is a bit redundant?

contrib/enum_metrics/enum_metrics/errant_ttl.py
```diff
@@ -0,0 +1,120 @@
...
+        if isVerbose: print "cqlstr: " + cqlstr
+        prepared_stmt = db_client.session.prepare(cqlstr)
+        db_client.session.execute(prepared_stmt)
```
ChandraAddala: Also you can move construction of the cqlstr inside the 'else' block as you only need the sql if it is not a dryrun.

ChandraAddala:
Can we update all rows to TTL of 3 days, let say to full resolution CF? What problems would this cause? If some rows have TTL of 1 day remaining, in worst case they will live for couple more days. This probably is a ridiculous idea, I am just trying to understand at this point.

contrib/enum_metrics/enum_metrics/errant_ttl.py
VinnyQ: erk I tested it with insert and not update :p   But ok let me try it with update if this matters to you

contrib/enum_metrics/enum_metrics/errant_ttl.py
VinnyQ: maybe but having a show seems safer and easier and more logical than having to type out repair with --dryrun just to "see" the rows.

VinnyQ:
@ChandraAddala re: updating the TTLs of all rows of a table... I am not sure why you would want that accomplished instead of just updating the ones that actually need fixing.   What is your reasoning?

contrib/enum_metrics/enum_metrics/errant_ttl.py
VinnyQ: Nevermind, insert is the only way to set the ttl with "using ttl xxx".  That command does not work with UPDATE and everywhere I see documented that you have to use insert if you want to set the ttl for a record, e.g.: https://docs.datastax.com/en/cql/3.1/cql/cql_using/use_ttl_t.html

or do something hackish like this:  http://stackoverflow.com/questions/24732494/how-to-update-time-to-livettl-of-existing-records-in-cassandra-db

ChandraAddala:
If there are too many rows which need an update, it would be easier to update all of them.

VinnyQ:
@ChandraAddala "update all of them" concept does not exists for cassandra >.<

VinnyQ:
neither does delete all of them for that matter.

However, you can delete (truncate) the entire table!  but that's something we do NOT want.

contrib/enum_metrics/enum_metrics/errant_ttl.py
```diff
@@ -0,0 +1,124 @@
...
+    key = str(record[0])
+    column1 = str(record[1])
+
+    cqlstr = "DELETE FROM %s where key=%r and column1=%s;" % (args.columnFamily, key, column1)
+    if isVerbose: print "cqlstr: " + cqlstr
+
+    if not args.dryrun:
+        prepared_stmt = db_client.session.prepare(cqlstr)
+        db_client.session.execute(prepared_stmt)
+
```
GeorgeJahad: can these return errors?  if so, we should check for them.

VinnyQ: good catch, error handling added

contrib/enum_metrics/enum_metrics/errant_ttl.py
```diff
@@ -0,0 +1,136 @@
...
+
+def delete_errant_ttl(db_client, args, record):
+    """
+    delete record where key,column1
+    """
+    key = str(record[0])
+    column1 = str(record[1])
+
+    cqlstr = "DELETE FROM %s where key=%r and column1=%s;" % (args.columnFamily, key, column1)
```
ChandraAddala: Have we considered "BATCH DELETE"?

usnavi:
@VinnyQ  - has all this work been completed?  should we merge?

VinnyQ:
no longer relevant.  this script is timing out in prod db and we found a different solution (directly deleting sstables files older than yymmdd date).

PR #582, opened 2016-01-01
Avoid binding only to loopbackport/localhost

goru97:
*Why:*
To make blueflood work with docker.


stackedsax:
I think it's fine to have this as the default.  Seems to me that the intention of Blueflood is usually to make metrics available to people via an externally accessible API.  Having 0.0.0.0 as the default seems in keeping with that intention.

+1

PR #585, opened 2016-01-03
Moving the media links to the webpage and the wiki

stackedsax:
Just wanting to get the README to speak solely about the code.

I have updated blueflood.io with these links and added a Media page in the wiki for the same stuff.


tilogaat:
+1

PR #586, opened 2016-01-03
Deleting the datastax stub until such a time as we actually do this work

stackedsax:
I don't think we need this here anymore.  It just adds confusion to anyone who comes to look at the code.


tilogaat:
LGTM @goru97 can you take a quick look too?

goru97:
:+1:

PR #589, opened 2016-01-08
adjust integration tests to not do string compare, to make tests work better in JDK 1.8

shintasmith:
Hashing mechanism is different in JDK 1.8. String compare of json output is generally not a good idea. The tests that are failing under JDK 1.8 has been updated to parse the JSON object and compare it with the expected JSON object.


shintasmith:
@usnavi, @ChandraAddala, @VinnyQ, @tilogaat, @GeorgeJahad pls review.

VinnyQ:
:+1: great job finding the bug and fixing the tests Shinta!

usnavi:
lgtm

PR #590, opened 2016-01-11
add tracking ingest requests for specific metric names

shintasmith:
This PR supersedes PR #552. The goal of the PR is to add a way to track only specific metric names that are sent to already tracked tenant.


tilogaat:
+1

usnavi:
+1

PR #591, opened 2016-01-12
for counters, return sampleCount in numPoints attribute

shintasmith:
Currently, numPoints for Counters holds the value of the stats for the counter, which is incorrect. 
```
curl -i -X GET "http://localhost:20000/v2.0/123456/views/my_counter?from=1452614038000&to=1499614038000&points=20"
HTTP/1.1 200 OK
Content-Length: 224

{
  "unit": "unknown",
  "values": [
    {
      "numPoints": 72,
      "timestamp": 1452556800000,
      "sum": 72
    }
  ],
  "metadata": {
    "limit": null,
    "next_href": null,
    "count": 1,
    "marker": null
  }
}
```

This PR changes it so that numPoints holds the value of sampleCount for the Counters. The same curl call would result:
```
curl -i -X GET "http://localhost:20000/v2.0/123456/views/my_counter?from=1452614038000&to=1499614038000&points=20"
HTTP/1.1 200 OK
Content-Length: 224

{
  "unit": "unknown",
  "values": [
    {
      "numPoints": 3,
      "timestamp": 1452556800000,
      "sum": 72
    }
  ],
  "metadata": {
    "limit": null,
    "next_href": null,
    "count": 1,
    "marker": null
  }
}
```

### DIFF
--- a/blueflood-http/src/main/java/com/rackspacecloud/blueflood/outputs/serializers/BasicRollupsOutputSerializer.java
+++ b/blueflood-http/src/main/java/com/rackspacecloud/blueflood/outputs/serializers/BasicRollupsOutputSerializer.java
@@ -106,7 +106,7 @@ public interface BasicRollupsOutputSerializer<T> {
                 else if (rollup instanceof BluefloodTimerRollup)
                     return ((BluefloodTimerRollup) rollup).getCount();
                 else if (rollup instanceof BluefloodCounterRollup)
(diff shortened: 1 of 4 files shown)


shintasmith:
@tilogaat, @GeorgeJahad, @VinnyQ, @usnavi, @ChandraAddala, please review. Thx!

usnavi:
lgtm

ChandraAddala:
+1

tilogaat:
+1

PR #593, opened 2016-01-18
Logstash support

shintasmith:
We currently run blueflood with ```logstash-support-log4j-1.0.1-SNAPSHOT-jar-with-dependencies.jar``` to enable us to send logs to logstash. This jar contains about 10 other 3rd party libraries.

This jar file was previously only manually built and checked into somewhere. This PR provides a way to (re)build the jar. 

I have also updated a couple of the 3rd party libraries to newer versions, as the version originally used in the jar is no longer available.


shintasmith:
@drenalin23 @usnavi @ChandraAddala @VinnyQ , please review. I use this to build the jar with newer version of logit. I have tested this in rollup nodes on qe02, just to make sure things run without exception. I saw some rollup logs made it to logstash/kibana in qe02.

cc: @GeorgeJahad @tilogaat

drenalin23:
LGTM.  This rocks.  We can build an artifact and upload to the artifact store in the staging account.  Long term I want to deprecate or update that setup - but I don't think that is a project for today (or this week).

ChandraAddala:
+1

PR #594, opened 2016-01-18
add profile, add ignore

shintasmith:
I forgot a few things from the previous PR #593  to enable building logstash-support:
* add an entry to git ignore contrib/logstash-support/target
* add a profile to the main pom.xml so I can do a build from Jenkins


usnavi:
lgtm

PR #595, opened 2016-01-18
add the actual jar file

shintasmith:
The previous PR missed the checkin of the actual jar file because it was ignored by .gitignore.


ChandraAddala:
+1

PR #596, opened 2016-01-19
Removing logging from bf-rollups-delay.py

usnavi:
Hi all, I attempted to run bf-rollups-delay.py on the db node and couldn't because @GeorgeJahad already wrote the file and I was locked out of writing to it.

Looked like the logging was pretty basic and thought it might be best just to get rid of it.  Let me know what you all think.

### DIFF
--- a/ops/rackspace-agent-plugins/bf-rollups-delay.py
+++ b/ops/rackspace-agent-plugins/bf-rollups-delay.py
@@ -17,7 +17,7 @@
 import pycassa
 import sys
 import time
-import logging
+
 from collections import defaultdict
 
 SLOTS = 4032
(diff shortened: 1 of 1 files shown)


shintasmith:
Why remove if it's not broken?

usnavi:
@shintasmith - it is broken in that the logging only writes to one location, and if you can't write to it the whole tool fails.  There is other ways of fixing it, but this looked reasonable enough.  I'm fine with updating the logging to be more flexible.

ChandraAddala:
Can we try configuring logging using a separate file? This link has a simple example. https://docs.python.org/2/howto/logging.html#configuring-logging

PR #597, opened 2016-01-20
test and refactor configuration class

izrik:
As a `tester`,

In order to `test the top-level service classes`,

I want to `override configuration values programmatically`.

Explanation
-----------

The `Configuration` class stores all of the configuration for the system. It includes facilities for loading config values from the config file, overriding with system properties, and falling back to hard-coded default values (from e.g. the `CoreConfig` enum).
Many parts of the system, many classes and objects, rely on this class in order to get set up. In order to test those classes and objects, I would need to override configuration values.

The problem with this is the way the `Configuration` class works.
It is not possible to properly test code that relies on it, because its effects can't be isolated.
All fields in the class are static.
Any change to a config value in one part of the system affects all other parts of the system that interact with that config value.
Also, changes persist from one test invocation to another.
This makes testing the top-level services impossible, because tests cannot be independent.

In order to manually set or override the configuration for a single test (e.g. an as-yet-unwritten `BluefloodServiceStartTest`), I need a way to modify a single `Configuration` instance (or load values into it) that doesn't affect any other instance.

Complicating all of this are a few other problems:
 1. The `getStringProperty` method has side effects. Sometimes. In certain circumstances. Which aren't documented anywhere.
 2. The overriding behavior (system property vs. from config file) is unclear, and the code is hard to read.
 3. **Storing values as strings** is a separate concern from **converting strings to other data types**, but they are conflated within the class. In order to provide proper tests of the type conversion methods, one must modify the global config values.
 4. Existing tests are cramped, and try to test multiple pieces of functionality at the same time.
 5. Once a value is overridden, either by loading from a file or from a system property, there is no way to un-set that value. Subsequent test invocations in the same JVM will always see a value for the given key.
 6. `getProperties` returns a map and ignores the `defaultProps` field completely

Proposed Changes
----------------

Before we can even get to the static/instance problem, the code has to be cleaned up.
This PR makes the following changes:
 - Add more unit tests for the `Configuration class`
 - Separate tests so that they only test one piece of functionality at a time (see #4 above)
 - Split out static conversion methods, so they can be tested apart from modifying global config values (see #3 above)
 - Add a `clearProperty` method that removes key/value pairs from the `props` field (see #5 above)
 - Add a `getAllProperties` method that returns a map with all key/value pairs, including from `defaultProps` (see #6 above)

The goal with this PR is to be able to properly test the `Configuration` class, _as it works now_, so that it can safely be refactored it in the future.
Subsequent PR's will (a) remove the side effects from `getStringProperty`, (b) clean up and clarify the overriding behavior, and (c) finally tackle the creation of independent instances of the class.

### DIFF
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/Configuration.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/Configuration.java
@@ -24,10 +24,6 @@ import java.io.InputStream;
 import java.net.URL;
 import java.util.*;
 
-/**
- * java/conf/bf-dev.con has an exhaustive description of each configuration option.
- *
- */
 public class Configuration {
     private static final Properties defaultProps = new Properties();
     private static Properties props;
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/Configuration.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/Configuration.java
@@ -57,13 +57,21 @@ public class Configuration {
         }
     }
 
+    /**
+     * Convert the existing configuration values into a Map. Does not include defaults from the defaultProps object.
+     *
+     * @return an unmodifiable Map
+     */
     public Map<Object,Object> getProperties() {
         return Collections.unmodifiableMap(props);
     }
 
+    /**
+     * Convert the existing configuration values into a Map, including those specified in defaultProps.
+     *
+     * @return an unmodifiable Map
+     */
     public Map<Object, Object> getAllProperties() {
-
-        // includes defaults values, which are not captured by the above getProperties method
 
         Map<Object, Object> map = new HashMap<Object, Object>();
         for (Object key : defaultProps.keySet()) {
@@ -72,6 +80,7 @@ public class Configuration {
         for (Object key : props.keySet()) {
             map.put(key, props.getProperty(key.toString()));
         }
+
         return Collections.unmodifiableMap(map);
     }
 
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/Configuration.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/Configuration.java
@@ -65,6 +65,20 @@ public class Configuration {
         return Collections.unmodifiableMap(props);
     }
 
+    public Map<Object, Object> getAllProperties() {
+
+        // includes defaults values, which are not captured by the above getProperties method
+
+        Map<Object, Object> map = new HashMap<Object, Object>();
+        for (Object key : defaultProps.keySet()) {
+            map.put(key, defaultProps.getProperty(key.toString()));
+        }
+        for (Object key : props.keySet()) {
+            map.put(key, props.getProperty(key.toString()));
+        }
+        return Collections.unmodifiableMap(map);
+    }
+
     public String getStringProperty(Enum<? extends ConfigDefaults> name) {
         return getStringProperty(name.toString());
     }
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/Configuration.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/Configuration.java
@@ -133,4 +133,9 @@ public class Configuration {
     public void setProperty(String name, String val) {
       props.setProperty(name, val);
     }
+
+    @VisibleForTesting
+    public void clearProperty(String name) {
+        props.remove(name);
+    }
 }
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/Configuration.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/Configuration.java
@@ -95,36 +95,50 @@ public class Configuration {
         return getIntegerProperty(name.toString());
     }
     public int getIntegerProperty(String name) {
-        return Integer.parseInt(getStringProperty(name));
+        return intFromString(getStringProperty(name));
+    }
+    public static int intFromString(String value) {
+        return Integer.parseInt(value);
     }
 
     public float getFloatProperty(Enum<? extends ConfigDefaults> name) {
         return getFloatProperty(name.toString());
     }
     public float getFloatProperty(String name) {
-        return Float.parseFloat(getStringProperty(name));
+        return floatFromString(getStringProperty(name));
+    }
+    public static float floatFromString(String value) {
+        return Float.parseFloat(value);
     }
 
     public long getLongProperty(Enum<? extends ConfigDefaults> name) {
         return getLongProperty(name.toString());
     }
     public long getLongProperty(String name) {
-        return Long.parseLong(getStringProperty(name));
+        return longFromString(getStringProperty(name));
+    }
+    public static long longFromString(String value) {
+        return Long.parseLong(value);
     }
 
     public boolean getBooleanProperty(Enum<? extends ConfigDefaults> name) {
         return getBooleanProperty(name.toString());
     }
-
     public boolean getBooleanProperty(String name) {
-        return "true".equalsIgnoreCase(getStringProperty(name));
+        return booleanFromString(getStringProperty(name));
+    }
+    public static boolean booleanFromString(String value) {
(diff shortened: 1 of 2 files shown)


izrik:
@ChandraAddala @usnavi @shintasmith @VinnyQ Please review and comment.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/Configuration.java
```diff
@@ -81,36 +95,50 @@ public int getIntegerProperty(Enum<? extends ConfigDefaults> name) {
         return getIntegerProperty(name.toString());
     }
```
ChandraAddala: Do not see much value in the following new methods that were added. intFromString, longFromString, floatFromString.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/Configuration.java
```diff
@@ -65,6 +65,20 @@ public void init() throws IOException {
         return Collections.unmodifiableMap(props);
     }
 
+    public Map<Object, Object> getAllProperties() {
```
ChandraAddala: I must be missing something. What does it do differently than getProperties? As far as I can see what getProperties returns should have default values in them. (init() method gets called during instantiation which seem to populate with default values first and then override them)

blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/ConfigurationTest.java
```diff
@@ -55,23 +56,170 @@ public void testInitWithBluefloodConfig() throws IOException {
...
+            // preconditions
+            Assert.assertTrue(
+                    "props should already have 'ROLLUP_KEYSPACE', because that's a default",
+                    props.containsKey(keyName));
+            Assert.assertFalse(
+                    "props should not contain 'original.ROLLUP_KEYSPACE' yet",
+                    props.containsKey("original." + keyName));
+
+            // act
+            System.setProperty(keyName, "some value");
```
usnavi: Why are you setting the System property here?

blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/ConfigurationTest.java
```diff
@@ -55,23 +56,170 @@ public void testInitWithBluefloodConfig() throws IOException {
...
+            config.setProperty(keyName, "some other value");
+            String value = config.getStringProperty(CoreConfig.ROLLUP_KEYSPACE);
+            Assert.assertEquals("some value", value);
```
usnavi: I'm confused here. Why are you writing to ```CoreConfig.ROLLUP_KEYSPACE.toString()``` but reading from ```CoreConfig.ROLLUP_KEYSPACE```?

usnavi: Nevermind.  I see what's happening in ```Configuration```.  Good lord.  Maybe add some comments about the screwiness going on here?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/Configuration.java
izrik: `getProperties` does NOT return the default values. See `testGetPropertiesDoesntWorkForDefaults`. What it does is treat the `props` object as a `Map`: all calls to `Map.get` will bypass the `Properties.getProperty` method, which is where default values are handled.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/ConfigurationTest.java
izrik: Because that is a requirement for the creation of `original.*` key/value pairs. The logic is, admittedly, very convoluted. I'm just trying to capture the behavior of the `getStringProperty` method as it currently works, and will try to clean it up in a subsequent PR.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/ConfigurationTest.java
izrik: Done.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/Configuration.java
ChandraAddala: To load default properties into props you have to call loadDefaults() method. Right? Thats how default properties seem to get loaded https://github.com/rackerlabs/blueflood/blob/master/blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/CoreConfig.java#L158

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/Configuration.java
izrik: These methods are responsible for converting values from strings to whatever datatype is desired.
By separating them into their own methods, we can test them without having to manually set system properties.
The ones you mention,( `intFromString`, `longFromString`, and `floatFromString`) are trivially implemented, because they just call the system parse methods.
Testing these methods ensures they function correctly, although since they're trivial enough that that's easy to tell just from looking at them.
However, the tests also serve as a kind of documentation of how the conversion works now.

We may very well decide to change the number format in the future, and these methods would be useful for that.
As an example, the way it works now, the conversion can't handle certain corner-cases, like leading or trailing whitespace.
If for some reason a configuration value had whitespace in it, an exception would be thrown.
To avoid that, we could change how the values were parsed.
Having the conversion in a separate method makes it easy to know where to make the change, and also makes it easy to test the change.

usnavi:
lgtm - oh boy, @izrik, I'm surprised you're still sane after that exercise :-P

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/Configuration.java
izrik: `loadDefaults` gets called by the various `enum` configuration types during static initialization. However, once that's already happened, the map returned from `getProperties` will not have any of those default values in it.

For what it's worth, the `getProperties` and `getAllProperties` methods are only ever called from `ConfigurationTest`. They are a kind of diagnostic, necessary to tell what's going on inside the `Configuration` class, based on how its internals work now. Since I intend to fix the internals of the class, and make them far more sane, they shouldn't be necessary in the future. I expect them to be removed at some point during the course of refactoring.

izrik:
@usnavi Eh, I've seen worse. I've even _written_ worse. ;)

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/Configuration.java
ChandraAddala: Well, all those three methods only have "core java" code in it. If we are testing those methods, we are only testing if java does what it is supposed to do. But if you have plans to add some application logic later on, thats fine. Also they dont have to be static.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/Configuration.java
ChandraAddala: If thats the case, may be we should fix getProperties, instead of creating a new method? I am fine either way I guess but can you atleast add some documentation to highlight the difference between the two?

ChandraAddala:
I made some comments. Other than that it looks good. Thanks for adding these tests.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/Configuration.java
izrik: Done.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/Configuration.java
shintasmith: Sorry I'm late to the review. But I have to agree with @ChandraAddala. I don't see much values in adding these static methods (intFromString(), floatFromString()) that simply call the core Java methods. You are also adding tests for each of these static methods -- these don't test any of our Configuration logic.

PR #598, opened 2016-01-20
New: added requirements.txt for python demo

Dogild:


stackedsax:
Seems totally reasonable.  Thanks for the PR.

It's funny, though, I recently wrote a bunch of documentation that totally uses files from the demo directory but I have never myself run the demo.  Would it be useful for me to write up something for this demo?  Did you find it useful to go through, @Dogild ?

Also, I know we use `requests` in this demo, but do we need `wsgiref`?

usnavi:
Would @tilogaat @GeorgeJahad @drenalin23 mind looking at this and accepting if appropriate?

Dogild:
@stackedsax The demo is useful ! What I would suggest is to make a demo with a UDP module as well, you don't find a lot of informations with the documentation on this point.

Other point, for this kind of project, people like to see some performance results etc etc...Maybe you could include a locust file (http://locust.io) to show how are the performance of your ReST API. Actually everybody do that when trying different technologies.

For the wsgiref, I did a pip freeze, I maybe messed up one of my virtualenv, let me correct that !

stackedsax:
+1

PR #599, opened 2016-01-20
Updating bf-rollups-delay.py

usnavi:
- moving log to ~/bf-rollup.log
- adding --servers argument
- adding comments & description
- now reports values for all rollup queues, includes ones with 0 value
- changed output value from float to uin32

### DIFF
--- a/ops/rackspace-agent-plugins/bf-rollups-delay.py
+++ b/ops/rackspace-agent-plugins/bf-rollups-delay.py
@@ -1,4 +1,6 @@
-#! /usr/bin/python
+#!/usr/bin/env python
+'''Blueflood Rollup Delay'''
+'''For each rollup level, lists the number of slots which need to processed by blueflood.  For the 5m range, one day is 288 slots.'''
 # Licensed to Rackspace under one or more
 # contributor license agreements.  See the NOTICE file distributed with
 # this work for additional information regarding copyright ownership.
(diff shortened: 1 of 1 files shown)


ops/rackspace-agent-plugins/bf-rollups-delay.py
```diff
@@ -1,4 +1,6 @@
-#! /usr/bin/python
+#!/usr/bin/env python
+'''Blueflood Rollup Delay'''
+'''For each rollup level, lists the number of slots which need to processed by blueflood.  One day is approximately 300 slots.'''
```
shintasmith: Do you mean the max for MIN1440 resolution is about 300 slots?

usnavi: A queue of 300 slots to-be-processed is equivalent of 1 day's worth of processing.  I.e., the system is a day behind.  Is that right @GeorgeJahad ?

shintasmith:
I like what you did with the logging :+1:

usnavi:
@shintasmith I wonder where I got the idea ;-)

ops/rackspace-agent-plugins/bf-rollups-delay.py
```diff
@@ -113,27 +120,34 @@ def print_stats_for_metrics_state(metrics_state_for_shards):
...
 
     for resol, delay in output.items():
-        print 'metric %s float %f slots' % ('_'.join([resol, 'delay']), delay)
+        print 'metric %s uint32 %u' % ('_'.join([resol, 'delay']), delay)
+
 
+def main():
+    parser = argparse.ArgumentParser(description='For each rollup level, lists the number of slots which need to '
+                                                 'be processed by blueflood.  One day is approximately 300 slots.')
+    parser.add_argument( '-s', '--servers', help='Cassandra server IP addresses, space separated', required=True, nargs="+")
```
shintasmith: Are these servers of the same cluster? If yes, what would be the benefit of passing in multiple servers? I am not critiquing, just trying to understand usage. :-)

usnavi: The ```pycassa.ConnectionPool``` API takes a list of servers, it picks a random one to connect to (https://pycassa.github.io/pycassa/api/pycassa/pool.html)

I assume this is to balance the load across the Cassandra servers.  The original version of the script allowed you to pass in multiple servers, and I kept that interface.  We won't be using it in our monitoring checks though.  We'll only be running against db00.

shintasmith: Got it. Thanks!

shintasmith:
lgtm :+1:

PR #600, opened 2016-01-22
fix path in start up script

shintasmith:
**Why:**
Yesterday while helping someone in #blueflood IRC, I noticed that the script assumes that it was run from top level directory and all references file paths (the log4j.properties, the *.conf file) is relative to the top level directory. This person was running the script from the bin/ directory and things didn't work.

**How:**
Change the script so that we use absolute paths for the files we referenced. This way, the script can be run from anywhere on the filesystem.

### DIFF
--- a/bin/blueflood-start.sh
+++ b/bin/blueflood-start.sh
@@ -12,6 +12,12 @@ SCRIPTDIR=`pwd`/`dirname $0`
 # top level source directory
 TOPDIR=$SCRIPTDIR/..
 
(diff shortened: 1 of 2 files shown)


shintasmith:
@ChandraAddala @usnavi @VinnyQ @drenalin23 pls review.

usnavi:
lgtm - nice!

ChandraAddala:
looks good.

PR #601, opened 2016-01-22
test and refactor configuration class - part 2

izrik:
In my previous PR, #597 , I had said that I next wanted to remove the side effects of `getStringProperty`. However, it's still a little too early to do that. Lots of the code relies on that method and doesn't yet have sufficient tests. Before I make that change I want to create more tests across the codebase.

This PR adds several tests that completely describe the overriding behavior of `getStringProperty`. These will be useful when we start changing the method's behavior.

This PR also adds some methods for manually reading, setting, and clearing values within the `Configuration` object, without manually dealing with system properties in the code. These will be useful in the next part, when I modify the codebase to use them instead of the sysprops.

Here's my checklist, roughly in order, of things I intend to do next:

- [x] clarify the overriding behavior
- [ ] change existing code and tests to use `setProperty` and similar methods, instead of setting system properties
- [ ] write a ton of tests for large portions of the codebase
- [ ] don't back up values under `original.*` keys
- [ ] don't store the values of system properties, just return them
- [ ] use independent instances of the class, instead of just a singleton

### DIFF
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/Configuration.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/Configuration.java
@@ -96,14 +96,14 @@ public class Configuration {
         return props.getProperty(name);
     }
 
-    public static String getRawStringProperty(Enum<? extends ConfigDefaults> name) {
+    public String getRawStringProperty(Enum<? extends ConfigDefaults> name) {
         return getRawStringProperty(name.toString());
     }
-    public static String getRawStringProperty(String name) {
+    public String getRawStringProperty(String name) {
         return props.getProperty(name);
     }
 
-    public static boolean containsKey(String key) {
+    public boolean containsKey(String key) {
         return props.containsKey(key);
     }
 
@@ -166,10 +166,10 @@ public class Configuration {
         setProperty(name.toString(), value);
     }
 
-    public static void clearProperty(String name) {
+    public void clearProperty(String name) {
         props.remove(name);
     }
-    public static void clearProperty(Enum<? extends ConfigDefaults> name) {
+    public void clearProperty(Enum<? extends ConfigDefaults> name) {
         clearProperty(name.toString());
     }
 }
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/Configuration.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/Configuration.java
@@ -148,13 +148,17 @@ public class Configuration {
         return list;
     }
 
-    @VisibleForTesting
     public void setProperty(String name, String val) {
       props.setProperty(name, val);
     }
+    public void setProperty(Enum<? extends ConfigDefaults> name, String value) {
+        setProperty(name.toString(), value);
+    }
 
-    @VisibleForTesting
     public void clearProperty(String name) {
         props.remove(name);
+    }
+    public void clearProperty(Enum<? extends ConfigDefaults> name) {
+        clearProperty(name.toString());
     }
 }
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/Configuration.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/Configuration.java
@@ -96,6 +96,17 @@ public class Configuration {
         return props.getProperty(name);
     }
 
+    public static String getRawStringProperty(Enum<? extends ConfigDefaults> name) {
+        return getRawStringProperty(name.toString());
+    }
+    public static String getRawStringProperty(String name) {
+        return props.getProperty(name);
+    }
+
+    public static boolean containsKey(String key) {
+        return props.containsKey(key);
+    }
+
     public int getIntegerProperty(Enum<? extends ConfigDefaults> name) {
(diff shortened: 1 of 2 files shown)


izrik:
@usnavi @shintasmith @ChandraAddala @VinnyQ Please review and comment.
CC: @GeorgeJahad @drenalin23

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/Configuration.java
```diff
@@ -96,6 +96,17 @@ public String getStringProperty(String name) {
...
     }
 
+    public static String getRawStringProperty(Enum<? extends ConfigDefaults> name) {
+        return getRawStringProperty(name.toString());
+    }
+    public static String getRawStringProperty(String name) {
+        return props.getProperty(name);
+    }
+
+    public static boolean containsKey(String key) {
```
shintasmith: Do these need to be static?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/Configuration.java
```diff
@@ -148,13 +159,17 @@ public static boolean booleanFromString(String value) {
...
+    public void setProperty(Enum<? extends ConfigDefaults> name, String value) {
+        setProperty(name.toString(), value);
+    }
 
-    @VisibleForTesting
-    public void clearProperty(String name) {
+    public static void clearProperty(String name) {
         props.remove(name);
     }
+    public static void clearProperty(Enum<? extends ConfigDefaults> name) {
```
shintasmith: Do these need to be static?

blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/ConfigurationTest.java
```diff
@@ -212,7 +220,7 @@ public void testSystemPropertiesOverrideConfigurationValues() {
                     config.getStringProperty(CoreConfig.MAX_CASSANDRA_CONNECTIONS));
         } finally {
             System.clearProperty(keyName);
-            config.clearProperty(keyName);
+            Configuration.clearProperty(keyName);
```
shintasmith: Why access this method staticly?

blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/ConfigurationTest.java
```diff
@@ -270,6 +278,337 @@ public void testGettingValuesCreatesOriginals() {
         }
     }
 
+    @Test
+    public void testOriginals1() {
```
shintasmith: Can we use a more meaningful test names other than testOriginal1 and testOriginal2?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/Configuration.java
izrik: Technically, no. They could also be instance methods. If they're static, they allow two modes of access:
1. `Configuration.getInstance().getRawStringProperty(...)`
2. `Configuration.getRawStringProperty(...)`

If they were instead instance methods, only the first way would work. Honestly, it's mostly just a matter of taste at this point. I eventually want to re-work the class so that the only  static methods are factory methods. Before then, these "raw" methods would go away. They are currently there just so I can read from the `props` field without having to make it public, and without invoking the wierd logic in the normal `getStringProperty` method. I want to refactor the `getStringProperty`, but need more tests across the whole code base before I do that.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/Configuration.java
izrik: See above.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/ConfigurationTest.java
izrik: See above. I can change them back, if it seems better.

usnavi:
other than the test method names, lgtm

blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/ConfigurationTest.java
izrik: Done

izrik:
@shintasmith @usnavi  Test methods renamed!

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/Configuration.java
shintasmith: @izrik, isn't static method harder to mock?

izrik: @shintasmith I imagine so, but this class isn't really in a position to be mocked right now. Nevertheless, I'll change it.

izrik: Done.

izrik:
"their" 
Now I feel like an idiot =(

shintasmith:
much better, lgtm. Once the Travis build finishes, feel free to merge in. Thanks!

PR #602, opened 2016-01-26
get rid of using files to serialize objects

shintasmith:
**Why:**
In order to improve test coverage, we need to be testing the serializer code. Our tests currently gated the serializing code with a system property: GENERATE_*_SERIALIZATION. We never run tests with these system properties set. We used a set of files that were generated a long time ago and these files were checked into Git.

**How:**
I am doing this incrementally. This is the first of many PRs to clean up the tests. With this PR, I am removing our use of the files checked in under ```src/test/resources/serializations``` directory. Everytime we run the tests, we serialize the objects into memory, and de-serialize them back into objects.

I am also getting rid of the loop that supposedly test back serialization/deserialization of older versions. We don't have older versions for most objects. We just have version 1 (whose value is 0). When we do have the next version in the future, we should write them as new methods, instead of testing them all in one method (which btw already has the name V1).

### DIFF
--- a/blueflood-core/src/test/java/com/rackspacecloud/blueflood/io/serializers/CounterRollupSerializationTest.java
+++ b/blueflood-core/src/test/java/com/rackspacecloud/blueflood/io/serializers/CounterRollupSerializationTest.java
@@ -13,21 +13,14 @@
  *    See the License for the specific language governing permissions and
  *    limitations under the License.
  */
-
 package com.rackspacecloud.blueflood.io.serializers;
 
-import com.rackspacecloud.blueflood.io.Constants;
 import com.rackspacecloud.blueflood.types.BluefloodCounterRollup;
 import junit.framework.Assert;
 import org.apache.commons.codec.binary.Base64;
 import org.junit.Test;
 
-import java.io.BufferedReader;
-import java.io.File;
-import java.io.FileOutputStream;
-import java.io.FileReader;
-import java.io.IOException;
-import java.io.OutputStream;
+import java.io.*;
 import java.nio.ByteBuffer;
 
 public class CounterRollupSerializationTest {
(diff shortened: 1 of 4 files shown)


shintasmith:

cc: @tilogaat @GeorgeJahad @drenalin23

blueflood-core/src/test/java/com/rackspacecloud/blueflood/io/serializers/HistogramSerializationTest.java
```diff
@@ -13,12 +13,10 @@
  *    See the License for the specific language governing permissions and
```
usnavi: FYI - @tilogaat says that the histogram API is not used by customers and will eventually be removed.  Not saying we shouldn't test it (perhaps this isn't even related to it) but good to be aware of when expanding the test coverage.

usnavi:
lgtm

blueflood-core/src/test/java/com/rackspacecloud/blueflood/io/serializers/HistogramSerializationTest.java
shintasmith: Ah, I didn't know that. Good to know. Since I am planning to get rid of the serialization file this test uses, I went ahead and clean it up.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/io/serializers/CounterRollupSerializationTest.java
```diff
@@ -36,37 +29,25 @@
...
+
+        ByteBuffer bb = ByteBuffer.wrap(Base64.decodeBase64(reader.readLine().getBytes()));
+        BluefloodCounterRollup cc0 = NumericSerializer.serializerFor(BluefloodCounterRollup.class).fromByteBuffer(bb);
+        Assert.assertEquals(c0, cc0);
+
+        bb = ByteBuffer.wrap(Base64.decodeBase64(reader.readLine().getBytes()));
+        BluefloodCounterRollup cc1 = NumericSerializer.serializerFor(BluefloodCounterRollup.class).fromByteBuffer(bb);
+
+        Assert.assertEquals(c1, cc1);
+        Assert.assertFalse(cc0.equals(cc1));
```
ChandraAddala: Can you close these streams like in a try/finally blocks?

shintasmith: Yeah... I was getting lazy. I thought since they are just ByteArrayIn/OutputStream that closing them would not do much.

PR #604, opened 2016-01-28
Adding --info flag to bf-rollup-delay.py for debugging

usnavi:
For a giving resolution:
- prints out unrolled slots & timestamps for all shards
- prints out last rolled up slot for all shards & timestamps for all shards

### DIFF
--- a/ops/rackspace-agent-plugins/bf-rollups-delay.py
+++ b/ops/rackspace-agent-plugins/bf-rollups-delay.py
@@ -89,7 +89,7 @@ def _get_slot_for_time(now_millis, gran):
     return (GRAN_MAPPINGS[gran]['max_slots'] * full_slot) / SLOTS
 
 
-def print_stats_for_metrics_state(metrics_state_for_shards):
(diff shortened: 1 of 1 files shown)


shintasmith:
If it's debugging output, why is it called ```--info``` ? How about ```--printUnrolledSlots``` and ```--printLastRolledSlots``` ?

usnavi:
I was thinking about ```--debugInfo``` but thought that was redundant.  I figure the tool is so narrow in its scope that the addition data merely expands on it that; a more specific flag doesn't really add value.  We'll see what @ChandraAddala or @VinnyQ @drenalin23 say.  ```---moreInfo```???

shintasmith:
Maybe ```--verbose``` ?

usnavi:
i like ```--verbose```

ChandraAddala:
+1

usnavi:
@shintasmith - now i'm verbose, in talk as well.

shintasmith:
lgtm

PR #605, opened 2016-01-29
add unit test coverage to coveralls

shintasmith:
We are only uploading the result of Integration tests to coveralls. This PR adds the result of unit tests to coveralls.


usnavi:
merge it!

shintasmith:

@stackedsax is there any technical or non-technical reason why we should NOT be uploading our unit tests results to coveralls?

PR #606, opened 2016-01-29
upgrade maven-coverall-plugins

shintasmith:
Upgrade to the latest maven-coverall-plugins. In version 4.0.0, they have added a way to combine multiple coverage reports
https://github.com/trautonen/coveralls-maven-plugin/blob/master/CHANGELOG.md#400


usnavi:
+1

PR #609, opened 2016-02-01
Test some service-related code.

izrik:
This PR adds some tests, chiefly for the `BluefloodServiceStarter` class. It also does a lot of cleanup of `ConfigurationTest` and various tests that use `Configuration` and/or system properties.

I had to make a few un-pretty modifications to `BluefloodServiceStarter` and `RollupService` to make them more testable.

`RollupServiceTest` is currently `@Ignore`d because of an initialization problem related to codahale metrics. Some of the code is trying to register a metric by a certain name, and when that code gets run twice, `MetricsRegistry` throws an exception with `""A metric named " + name + " already exists"`. I have to work out how to get around that.

This PR brings total test coverage from about 67% up to about 70%. In particular, it brings `BluefloodServiceStarter` from 0% to 62%. I'm still seeing some inconsistencies in the coveralls reports though, and I'll have to investigate those a little more.


blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/BluefloodServiceStarter.java
```diff
@@ -295,8 +296,10 @@ public void run() {
                     TimeUnit.MINUTES.toMillis(savePeriodMins));
         }
 
-        // has the side-effect of causing static initialization of Metrics, starting instrumentation reporting.
-        new RestartGauge(Metrics.getRegistry(), RollupService.class);
+        if (!SkipInstantiateRestartGauge) {
+            // has the side-effect of causing static initialization of Metrics, starting instrumentation reporting.
+            new RestartGauge(Metrics.getRegistry(), RollupService.class);
```
shintasmith: Yikes! RestartGauge() is our class. Should we have built in some logic to handle this inside the class?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/BluefloodServiceStarter.java
```diff
@@ -313,4 +316,20 @@ public void run() {
         startEventListenerModules();
         log.info("All blueflood services started");
     }
+
+    @VisibleForTesting
+    public static boolean SkipInstantiateRestartGauge = false;
+
+    @VisibleForTesting
+    public static boolean ThrowInsteadOfExit = false;
```
shintasmith: I am not a big fan of test code cluttering main code. Is it possible to just mock the System.exit() code?
http://stackoverflow.com/questions/309396/java-how-to-test-methods-that-call-system-exit

blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/BluefloodServiceStarterTest.java
```diff
@@ -0,0 +1,275 @@
...
+        BluefloodServiceStarter.ThrowInsteadOfExit = true;
+        BluefloodServiceStarter.SkipInstantiateRestartGauge = true;
+    }
+
+    @Test
+    public void testStartup() throws BluefloodServiceStarterException {
+        String[] args = new String[0];
+        BluefloodServiceStarter.main(args);
+
+        assertNotNull(args);
```
shintasmith: What are we really asserting here? What is the test supposed to check? That it did not get exception?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/BluefloodServiceStarter.java
ChandraAddala: I agree with @shintasmith

blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/BluefloodServiceStarterTest.java
```diff
@@ -0,0 +1,275 @@
...
+        // given
+        Configuration config = Configuration.getInstance();
+        config.setProperty(CoreConfig.CASSANDRA_HOSTS, "");
+
+        // when
+        BluefloodServiceStarterException ex = null;
+        try {
+            BluefloodServiceStarter.validateCassandraHosts();
+        } catch (BluefloodServiceStarterException e) {
+            ex = e;
```
shintasmith: You could do the same thing by doing 
```@Test(expected = BluefloodServiceStarterException.class)```
I would say checking for exit code equals -1 is not that important to do.

File: blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/BluefloodServiceStarterTest.java [hunk 5]
```diff
@@ -0,0 +1,275 @@
...
+        }
+
+        // then
+        assertNotNull(ex);
```
shintasmith: Can use ```@Test(expected = ...)```, see above.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/BluefloodServiceStarter.java
```diff
@@ -247,7 +248,7 @@ private static void startEventListenerModules() {
         }
     }
 
-    public static void main(String args[]) {
+    public static void main(String args[]) throws BluefloodServiceStarterException {
```
ChandraAddala: I dont think we should throw exception out of main method. We should catch it in the main method and exit with a proper exit code.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/BluefloodServiceStarterTest.java
```diff
@@ -0,0 +1,275 @@
...
```
shintasmith: Can use ```@Test(expected = ...)```, see above.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/BluefloodServiceStarter.java
```diff
@@ -313,4 +316,20 @@ public void run() {
...
         log.info("All blueflood services started");
     }
+
+    @VisibleForTesting
+    public static boolean SkipInstantiateRestartGauge = false;
+
+    @VisibleForTesting
+    public static boolean ThrowInsteadOfExit = false;
+
+    private static void SystemExit(int status) throws BluefloodServiceStarterException {
```
ChandraAddala: Method should start with small letter. Also does this method have to be static? This method seems a little redundant to me especially if we get rid of this flag ThrowInsteadOfExit. 

I would have taken this approach if we really want to get rid of System.exit calls. We should throw exception everwhere we have System.exit(1) calls. Store the exit code inside the exception. Catch the exception in the main method and exit with proper exit code.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/BluefloodServiceStarterException.java
```diff
@@ -0,0 +1,14 @@
+package com.rackspacecloud.blueflood.service;
+
+public class BluefloodServiceStarterException extends Exception {
```
ChandraAddala: Can you extend RuntimeException so that we dont have to explicitly throw it?

blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/BluefloodServiceStarterTest.java
```diff
@@ -0,0 +1,275 @@
...
+        // when
+        BluefloodServiceStarterException ex = null;
+        try {
+            BluefloodServiceStarter.main(args);
+        } catch (BluefloodServiceStarterException e) {
+            ex = e;
+        }
+
+        // then
+        assertNull(ex);
```
shintasmith: We can simplify this test method this way:
```
@Test
public void testRollupModeEnabledShouldNotFail() throws Exception {
     config.setProperty(CoreConfig.ROLLUP_MODE, "true");
     String[] args = new String[0];
     BluefloodServiceStarter.main(args);   
}
```

blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/BluefloodServiceStarterTest.java
```diff
@@ -0,0 +1,275 @@
...
+        assertEquals(1, DummyEventListenerService.getInstances().size());
+        assertNotNull(DummyEventListenerService.getMostRecentInstance());
+
+        assertTrue(DummyEventListenerService.getMostRecentInstance().getStartServiceCalled());
+    }
+
+    @After
+    public void tearDown() throws IOException {
+        Configuration config = Configuration.getInstance();
+        config.init();
```
shintasmith: Why are we calling init() at tearDown() ?

blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/BluefloodServiceStarterTest.java
```diff
@@ -0,0 +1,275 @@
...
+
+    @Before
+    public void setUp() throws IOException {
+
+        Configuration config = Configuration.getInstance();
+
+        config.init();
+
+        // take from demo/local/config/blueflood.conf
+        config.setProperty(CoreConfig.CASSANDRA_HOSTS, "127.0.0.1:9160");
```
ChandraAddala: Cant we create a test config file and read form that if you dont want to use "demo/local/config/blueflood.conf"? Are these settings specific for this test in any way?

ChandraAddala: I also think we should try to setup/instantiate configuration class, same as the way the main code does so that tests are close to the main code setup.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/BluefloodServiceStarterTest.java
```diff
@@ -0,0 +1,275 @@
...
+        config.setProperty(CoreConfig.CASSANDRA_MAX_RETRIES, "5");
+        config.setProperty("ELASTICSEARCH_HOSTS", "localhost:9300");
+        config.setProperty("ELASTICSEARCH_CLUSTERNAME", "elasticsearch");
+
+        BluefloodServiceStarter.ThrowInsteadOfExit = true;
+        BluefloodServiceStarter.SkipInstantiateRestartGauge = true;
+    }
+
+    @Test
+    public void testStartup() throws BluefloodServiceStarterException {
```
ChandraAddala: What is being tested in this test?

blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/BluefloodServiceStarterTest.java
izrik: To reset the configuration for other tests. Different tests were interacting with the configuration and not cleaning up after themselves, which interfered with other tests. By calling init both before and after, we make sure that the tests are isolated.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/BluefloodServiceStarter.java
izrik: Probably, but I don't know what the necessary logic would be. I don't really understand everything that this class is doing. It's a major smell in my opinion that it's instantiating an object but not storing a reference to it, a "naked `new`". There's a lot of deep magic underneath. At this point I'm just trying to isolate portions of the system to test them.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/BluefloodServiceStarter.java
izrik: Excellent. I didn't find anything in the oracle docs about trapping `System.exit`, but these other things will do. I'll use them. I'm not a big fan of this either.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/BluefloodServiceStarterTest.java
izrik: The aim was to make assertions on the caught exception. If the team is ok with ignoring that part for now, then `expected` will serve the purpose just fine.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/BluefloodServiceStarter.java
izrik: Will do.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/ConfigurationTest.java
```diff
@@ -37,31 +40,38 @@ public void tearDown() throws IOException {
...
+            Map<Object, Object> properties = config.getProperties();
 
-        Assert.assertEquals("127.0.0.1:19180", config.getStringProperty(CoreConfig.CASSANDRA_HOSTS));
-        System.setProperty("CASSANDRA_HOSTS", "127.0.0.2");
-        Assert.assertEquals("127.0.0.2", config.getStringProperty(CoreConfig.CASSANDRA_HOSTS));
+            Assert.assertNotNull(properties);
 
-        Assert.assertEquals(60000, config.getIntegerProperty(CoreConfig.SCHEDULE_POLL_PERIOD));
+            Assert.assertEquals("127.0.0.1:19180", config.getStringProperty(CoreConfig.CASSANDRA_HOSTS));
+            System.setProperty("CASSANDRA_HOSTS", "127.0.0.2");
```
ChandraAddala: Not a big deal but I would create a constant for "127.0.0.2" if its being used multiple times.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/BluefloodServiceStarterTest.java
izrik: The configuration is more-or-less specific to these tests, especially the part about the `INGEST_MODE`, `QUERY_MODE`, etc being false. If the other files change, this test will still work as before. `testStartup` was the first one I wrote in this class, and was just for starters. The other tests actually test variations on the configuration.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/RollupService.java
```diff
@@ -204,8 +204,14 @@ final void poll() {
     }
 
     public void run() {
+        run(null, null);
+    }
+    public void run(Long maxNumIterations, Long maxMillisRunTime) {
```
shintasmith: My principal is not to change too much of the main code/functionality for just tests. I understand sometimes moving code around (like breaking a big method to smaller methods) so we can test properly is sometimes necessary. I am ok with that example of breaking a big method to smaller ones. But at the very least, we should avoid changing the functionality.

This change introduces new functionality of the main code, for the purpose of tests. I am having troubles justifying it.

A quick google found me this. Now... I admit, I haven't tried it. I don't know for sure if it would work. But this should hopefully give you maybe some ideas:

http://stackoverflow.com/questions/17206527/is-it-possible-to-verify-a-mock-method-running-in-different-thread-in-mockito

Should look at JMockit too, to see if it has something like this. I'd rather us use just one mocking tool, but if it's not the end of the world if we use JMockit and Mockito both.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/BluefloodServiceStarter.java
shintasmith: Got it. I am fine with keeping it here.

FYI. The class is just a wrapper class to the CodeHale metrics API.

izrik:
@shintasmith @ChandraAddala How does it look now?

blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/BluefloodServiceStarterTest.java
```diff
@@ -0,0 +1,191 @@
...
+
+        config.init();
+
+        config.setProperty(CoreConfig.ZOOKEEPER_CLUSTER, "NONE");
+
+        config.setProperty(CoreConfig.INGEST_MODE, "false");
+        config.setProperty(CoreConfig.ROLLUP_MODE, "false");
+        config.setProperty(CoreConfig.QUERY_MODE, "false");
+
+        BluefloodServiceStarter.SkipInstantiateRestartGauge = true;
```
shintasmith: How is this going to work if there are multiple test classes using this and they are run in parallel?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/BluefloodServiceStarter.java
```diff
@@ -313,4 +323,7 @@ public void run() {
         startEventListenerModules();
         log.info("All blueflood services started");
     }
+
+    @VisibleForTesting
+    public static boolean SkipInstantiateRestartGauge = false;
```
ChandraAddala: Can you make the variable start with smaller case? Usually for booleans, it is preceded with "is" or something like that but thats upto you.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/BluefloodServiceStarterTest.java
izrik: That's why I turned parallelization off. There were tests already modifying configuration in parallel before I even started this branch. We shouldn't turn parallel test execution back on until we can isolate the components better.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/BluefloodServiceStarter.java
izrik: Done.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/BluefloodServiceStarter.java
```diff
@@ -295,8 +303,10 @@ public void run() {
                     TimeUnit.MINUTES.toMillis(savePeriodMins));
         }
 
-        // has the side-effect of causing static initialization of Metrics, starting instrumentation reporting.
-        new RestartGauge(Metrics.getRegistry(), RollupService.class);
+        if (shouldInstantiateRestartGauge) {
```
ChandraAddala: If this flag is only for tests, I have an idea. We can create a separate method just for this one line "Metrics.getRegistry()" and then mock the return value from tests? You have to try this to see if it works.  Its worth a shot if it means getting rid of a flag.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/BluefloodServiceStarterTest.java
```diff
@@ -0,0 +1,191 @@
...
+    public void testAllModesDisabled() {
+
+        // given
+        Configuration config = Configuration.getInstance();
+        config.setProperty(CoreConfig.INGEST_MODE, "false");
+        config.setProperty(CoreConfig.QUERY_MODE, "false");
+        config.setProperty(CoreConfig.ROLLUP_MODE, "false");
+
+        // when
+        BluefloodServiceStarter.run();
```
ChandraAddala: To test if a method is not throwing any exception, you can do this..
```
   try {
       method();       
   } catch (Exception e) {
      fail( "Exception should not be thrown" );
   }
```

shintasmith: This will eat the exception. I'd rather the exception to be bubbled up to the reports.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/BluefloodServiceStarter.java
shintasmith: I agree with @ChandraAddala. If all you want to do is skip the call of RestartGauge, then mock it.

izrik: Nope. I tried extracting a method called `initializeInstrumentationReporting` that returned the `RestartGuage` object, and then mocking that method with PowerMock, but it's getting really difficult. I think it's ok to leave the flag as-is for now, until we can refactor the class enough that we don't need it anymore.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/BluefloodServiceStarterTest.java
izrik: That is the opposite of [what @shintasmith suggested](https://github.com/rackerlabs/blueflood/pull/609#discussion_r51589607). I think it's ok for now to leave it as is.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/ConfigurationTest.java
izrik: Those are old tests. I will eventually remove them.

izrik:
@shintasmith @ChandraAddala Any more suggestions or questions?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/BluefloodServiceStarter.java
izrik: Sorry, @shintasmith, apparently the page didn't refresh earlier and I missed your comment. I'll try that.

izrik: This is still getting really hairy. Trying to mock the RestartGauge constructor with PowerMock requires changing the junit version, and I'm still getting `java.lang.NoClassDefFoundError` and `"java.lang.LinkageError: loader constraint violation`. I vote we leave the flag in place, and work on the `RestartGauge` in a separate PR.

shintasmith: It's fine. But please create a story to clean this up. I really don't like these test-only flags in the main code.

izrik: I don't like them any more than anyone else, but fixing the instrumentation stuff is way beyond the scope of this PR.

ChandraAddala: @izrik I had created a PR against your branch. https://github.com/rackerlabs/blueflood/pull/611

Pls try this and let me know if it looks okay. I had to settle for a lower version of powermock which is compatible with the version of mockito we have. May be I should have upgraded both. I didnt try that.

While I was looking into it, I felt like there is no need for all the methods in BluefloodServiceStarter to be static. Its a good idea to actually make all methods non-static, instantiate the class from main method and call the methods using that. May be for a different PR.

PR #610, opened 2016-02-02
Finder performance issues

ChandraAddala:
Finder performance issues
     - Adding a new API /next_token to grab tokens of metric name
     - Refactored common code of ElasticIO and EnumElasticIO to AbstractElasticIO
     - Added tests

### DIFF
--- a/blueflood-core/src/test/java/com/rackspacecloud/blueflood/utils/GlobPatternTest.java
+++ b/blueflood-core/src/test/java/com/rackspacecloud/blueflood/utils/GlobPatternTest.java
@@ -9,7 +9,7 @@ import java.util.regex.PatternSyntaxException;
 public class GlobPatternTest {
 
     @Test
-    public void testGlobToRegex1() {
+    public void testGlobMatchingAnyChar() {
 
         String glob = "*";
         String expectedRegex = ".*";
@@ -19,7 +19,7 @@ public class GlobPatternTest {
     }
 
     @Test
-    public void testGlobToRegex2() {
+    public void testEmptyGlob() {
 
         String glob = "";
         String expectedRegex = "";
@@ -29,7 +29,7 @@ public class GlobPatternTest {
     }
 
     @Test
-    public void testGlobToRegex3() {
+    public void testGlobWithWildcards1() {
 
         String glob = "foo.bar$1.(cat).baz|qux.dog+";
         String expectedRegex = "foo\\.bar\\$1\\.\\(cat\\)\\.baz\\|qux\\.dog\\+";
@@ -39,7 +39,7 @@ public class GlobPatternTest {
     }
 
     @Test
-    public void testMetricNameGlob1() {
+    public void testMetricNameGlob() {
 
         String glob = "foo.bar.*";
         String expectedRegex = "foo\\.bar\\..*";
@@ -49,7 +49,7 @@ public class GlobPatternTest {
     }
 
     @Test
-    public void testMetricNameGlob2() {
+    public void testMetricNameGlobWithWildCards() {
 
         String glob = "f*.bar.*";
         String expectedRegex = "f.*\\.bar\\..*";
@@ -59,7 +59,7 @@ public class GlobPatternTest {
     }
 
     @Test
-    public void testMetricNameGlob5() {
+    public void testMetricNameWithMatchingSingleChar() {
 
         String glob = "foo?";
         String expectedRegex = "foo.";
@@ -69,7 +69,7 @@ public class GlobPatternTest {
     }
 
     @Test
-    public void testMetricNameGlob6() {
+    public void testMetricNameWithMatchingAnyChar() {
 
         String glob = "foo*";
         String expectedRegex = "foo.*";
@@ -79,7 +79,7 @@ public class GlobPatternTest {
     }
 
     @Test
-    public void testMetricNameGlob7() {
+    public void testMetricNameWithGlobSyntax() {
 
         String glob = "foo.{*}";
         String expectedRegex = "foo\\.(.*)";
@@ -90,7 +90,7 @@ public class GlobPatternTest {
 
     //I would expect this test to pass but it did not. Anything inside brackets [ ] should be left unescaped.
     @Ignore
-    public void testMetricNameGlob8() {
+    public void testVariousGlobSyntax1() {
 
         String glob = "fo[^ab.co].*]";
         String expectedRegex = "fo[^ab.co]\\..*]";
@@ -100,7 +100,7 @@ public class GlobPatternTest {
     }
 
     @Test
-    public void testMetricNameGlob9() {
+    public void testVariousGlobSyntax2() {
 
         String glob = "[!abc]oo.*]";
         String expectedRegex = "[^abc]oo\\..*]";
@@ -110,7 +110,7 @@ public class GlobPatternTest {
     }
 
     @Test
-    public void testMetricNameGlob3() {
+    public void testMetricNameWithVariousGlobSyntax() {
 
         String glob = "foo.[bz]*.*";
         String expectedRegex = "foo\\.[bz].*\\..*";
@@ -120,7 +120,7 @@ public class GlobPatternTest {
     }
 
     @Test
-    public void testMetricNameGlob4() {
+    public void testMetricNameGlobWithoutWildCard() {
 
         String glob = "foo.bar";
         String expectedRegex = "foo\\.bar";
@@ -130,13 +130,13 @@ public class GlobPatternTest {
     }
 
     @Test(expected = PatternSyntaxException.class)
-    public void invalidGlob1() {
+    public void invalidGlobWithUnclosedBracket() {
         String glob = "foo.[bar.*";
         GlobPattern pattern = new GlobPattern(glob);
     }
 
     @Test(expected = NullPointerException.class)
-    public void invalidGlob2() {
+    public void testNullGlob() {
         String glob = null;
         GlobPattern pattern = new GlobPattern(glob);
     }
--- a/blueflood-core/src/test/java/com/rackspacecloud/blueflood/utils/GlobPatternTest.java
+++ b/blueflood-core/src/test/java/com/rackspacecloud/blueflood/utils/GlobPatternTest.java
@@ -1,0 +1,144 @@
+package com.rackspacecloud.blueflood.utils;
+
+import org.junit.Assert;
+import org.junit.Ignore;
+import org.junit.Test;
+
+import java.util.regex.PatternSyntaxException;
+
+public class GlobPatternTest {
+
+    @Test
+    public void testGlobToRegex1() {
+
+        String glob = "*";
+        String expectedRegex = ".*";
+
+        GlobPattern pattern = new GlobPattern(glob);
+        Assert.assertEquals("Invalid regex", expectedRegex, pattern.compiled().toString());
+    }
+
+    @Test
+    public void testGlobToRegex2() {
+
+        String glob = "";
+        String expectedRegex = "";
+
+        GlobPattern pattern = new GlobPattern(glob);
+        Assert.assertEquals("Invalid regex", expectedRegex, pattern.compiled().toString());
+    }
+
+    @Test
+    public void testGlobToRegex3() {
+
+        String glob = "foo.bar$1.(cat).baz|qux.dog+";
+        String expectedRegex = "foo\\.bar\\$1\\.\\(cat\\)\\.baz\\|qux\\.dog\\+";
+
+        GlobPattern pattern = new GlobPattern(glob);
+        Assert.assertEquals("Invalid regex", expectedRegex, pattern.compiled().toString());
+    }
+
+    @Test
+    public void testMetricNameGlob1() {
+
+        String glob = "foo.bar.*";
+        String expectedRegex = "foo\\.bar\\..*";
+
+        GlobPattern pattern = new GlobPattern(glob);
+        Assert.assertEquals("Invalid regex", expectedRegex, pattern.compiled().toString());
+    }
+
+    @Test
+    public void testMetricNameGlob2() {
+
+        String glob = "f*.bar.*";
+        String expectedRegex = "f.*\\.bar\\..*";
+
+        GlobPattern pattern = new GlobPattern(glob);
+        Assert.assertEquals("Invalid regex", expectedRegex, pattern.compiled().toString());
+    }
+
+    @Test
+    public void testMetricNameGlob5() {
+
+        String glob = "foo?";
+        String expectedRegex = "foo.";
+
+        GlobPattern pattern = new GlobPattern(glob);
+        Assert.assertEquals("Invalid regex", expectedRegex, pattern.compiled().toString());
+    }
+
+    @Test
+    public void testMetricNameGlob6() {
+
+        String glob = "foo*";
+        String expectedRegex = "foo.*";
+
+        GlobPattern pattern = new GlobPattern(glob);
+        Assert.assertEquals("Invalid regex", expectedRegex, pattern.compiled().toString());
+    }
+
+    @Test
+    public void testMetricNameGlob7() {
+
+        String glob = "foo.{*}";
+        String expectedRegex = "foo\\.(.*)";
+
+        GlobPattern pattern = new GlobPattern(glob);
+        Assert.assertEquals("Invalid regex", expectedRegex, pattern.compiled().toString());
+    }
+
+    //I would expect this test to pass but it did not. Anything inside brackets [ ] should be left unescaped.
+    @Ignore
+    public void testMetricNameGlob8() {
+
+        String glob = "fo[^ab.co].*]";
+        String expectedRegex = "fo[^ab.co]\\..*]";
+
+        GlobPattern pattern = new GlobPattern(glob);
+        Assert.assertEquals("Invalid regex", expectedRegex, pattern.compiled().toString());
+    }
+
+    @Test
+    public void testMetricNameGlob9() {
+
+        String glob = "[!abc]oo.*]";
+        String expectedRegex = "[^abc]oo\\..*]";
+
+        GlobPattern pattern = new GlobPattern(glob);
+        Assert.assertEquals("Invalid regex", expectedRegex, pattern.compiled().toString());
+    }
+
+    @Test
+    public void testMetricNameGlob3() {
+
+        String glob = "foo.[bz]*.*";
+        String expectedRegex = "foo\\.[bz].*\\..*";
+
+        GlobPattern pattern = new GlobPattern(glob);
+        Assert.assertEquals("Invalid regex", expectedRegex, pattern.compiled().toString());
+    }
+
+    @Test
+    public void testMetricNameGlob4() {
+
+        String glob = "foo.bar";
+        String expectedRegex = "foo\\.bar";
+
+        GlobPattern pattern = new GlobPattern(glob);
+        Assert.assertEquals("Invalid regex", expectedRegex, pattern.compiled().toString());
+    }
+
+    @Test(expected = PatternSyntaxException.class)
+    public void invalidGlob1() {
+        String glob = "foo.[bar.*";
+        GlobPattern pattern = new GlobPattern(glob);
+    }
+
+    @Test(expected = NullPointerException.class)
+    public void invalidGlob2() {
+        String glob = null;
+        GlobPattern pattern = new GlobPattern(glob);
+    }
+
+}
--- a/blueflood-core/src/test/java/com/rackspacecloud/blueflood/utils/GlobPatternTest.java
+++ b/blueflood-core/src/test/java/com/rackspacecloud/blueflood/utils/GlobPatternTest.java
@@ -15,7 +15,7 @@ public class GlobPatternTest {
         String expectedRegex = ".*";
 
         GlobPattern pattern = new GlobPattern(glob);
-        Assert.assertEquals("Invalid regex", expectedRegex, pattern.compiled().toString());
+        Assert.assertEquals(expectedRegex, pattern.compiled().toString());
     }
 
     @Test
@@ -25,7 +25,7 @@ public class GlobPatternTest {
         String expectedRegex = "";
 
         GlobPattern pattern = new GlobPattern(glob);
-        Assert.assertEquals("Invalid regex", expectedRegex, pattern.compiled().toString());
+        Assert.assertEquals(expectedRegex, pattern.compiled().toString());
     }
 
     @Test
@@ -35,7 +35,7 @@ public class GlobPatternTest {
         String expectedRegex = "foo\\.bar\\$1\\.\\(cat\\)\\.baz\\|qux\\.dog\\+";
 
         GlobPattern pattern = new GlobPattern(glob);
-        Assert.assertEquals("Invalid regex", expectedRegex, pattern.compiled().toString());
+        Assert.assertEquals(expectedRegex, pattern.compiled().toString());
     }
 
     @Test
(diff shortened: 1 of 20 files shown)


blueflood-core/src/test/java/com/rackspacecloud/blueflood/utils/GlobPatternTest.java
```diff
@@ -0,0 +1,144 @@
+package com.rackspacecloud.blueflood.utils;
+
+import org.junit.Assert;
+import org.junit.Ignore;
+import org.junit.Test;
+
+import java.util.regex.PatternSyntaxException;
+
+public class GlobPatternTest {
```
usnavi: :+1 for doing this!  Can we add some more descriptive names than testGlobToRegex1, testGlobToRegex2, etc.  I'd do something along the lines of GlobToRegexPeriodStar, GlobToRegexEmpty, etc.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/AbstractElasticIO.java
```diff
@@ -0,0 +1,444 @@
...
+
+
+        for (Terms.Bucket bucket: aggregateTerms.getBuckets()) {
+            final String key = bucket.getKey();
+            final String[] tokens = key.split(REGEX_TOKEN_DELIMTER);
+
+            /**
+             * Lets say we get these buckets with prefix of 'foo'. foo should have a doc_count that
+             * matches the sum of all its next levels(only one level deep) which is 3. In the example
+             * below 'foo.bar' is a bar by itself and so its count will be higher than its sublevels.
```
usnavi: add example of test data, as we discussed.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/AbstractElasticIO.java
```diff
@@ -0,0 +1,444 @@
...
+     * @param completeMetricNamesByLevel contains complete metric names.
+     */
+    private void processMetricTokensAggregate(final Terms aggregateTerms,
+                                              final int totalTokensInPrefix,
+                                              final Set<String> tokensWithNextLevel,
+                                              final Map<Integer, Set<String>> completeMetricNamesByLevel) {
+
+        //0 indicates ground level. if prefix is foo.bar, 0 is at foo.bar level
+        final Map<String, Long> partialMetrics0LevelMap = new LinkedHashMap<String, Long>();
+        final Map<String, Long> partialMetrics0LevelAggMap = new LinkedHashMap<String, Long>();
```
usnavi: Can you explain what's exactly held in these maps?  I don't understand the difference between *Map & *AggMap

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/AbstractElasticIO.java
```diff
@@ -0,0 +1,444 @@
...
+    /**
+     * If we think of metric name(Ex: foo.bar.baz) as hierarchical tokens separated by delimiter("."),
+     * this API method would return the next level of tokens which follow the given prefix.
+     *
+     * @param tenant
+     * @param prefix is glob representation
+     * @return
+     * @throws Exception
+     */
+    public List<TokenInfo> getNextTokens(final String tenant, final String prefix) throws Exception {
```
usnavi: We'll see what others say, but might make it easier to understand if you broken this method down a bit more.  It would make it easier to glance at and see the different parts of the process.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/AbstractElasticIO.java
ChandraAddala: Its documented below along with test data. I am going to change this documentation as discussed.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/AbstractElasticIO.java
```diff
@@ -0,0 +1,463 @@
...
+
+            searchForEnumValues(tenant, prefix, enumValuesAs1LevelSet, tokensWithEnumsAs2LevelMap);
+        }
+
+        return prepareResults(tokensWithNextLevel,
+                tokensWithEnumsAs2LevelMap, enumValuesAs1LevelSet);
+    }
+
+    private void searchForEnumValues(String tenant, String prefix, Set<String> enumValuesAs1LevelSet, Map<String, Boolean> tokensWithEnumsAs2LevelMap) {
+        List<String> queries = new ArrayList<String>();
```
usnavi: Would you explain why this is required and how it interfaces with ES?

usnavi:
I think the interface with ES is so obtuse that I am unable to follow this.  I wouldn't be able to pick up a follow-up story or bug on this code.  Perhaps needs more background explanation in the comments and break it down into smaller methods with descriptive method names?

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/AbstractElasticIO.java
```diff
@@ -0,0 +1,463 @@
...
+
+        SearchResponse response = client.prepareSearch(indexes)
+                .setRouting(tenant)
+                .setSize(100000)
+                .setVersion(true)
+                .setQuery(bqb)
+                .execute()
+                .actionGet();
+
+        multiSearchCtx.stop();
```
shintasmith: Do we want to put the above in the finally{} block, in case there are exceptions thrown from the search?

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/AbstractElasticIO.java
```diff
@@ -0,0 +1,463 @@
...
+     * (Some exceptions when prefix has only one level due to the nature of underlying data)
+     *
+     * @param prefix
+     * @param level
+     * @return
+     */
+    private String regexforNextNLevels(final String prefix, final int level) {
+
+        //grabs chars until the next "." which is basically a token
+        String regexToGrabSingleToken = "[^.]+";
```
shintasmith: nitpick, any reason for not making this a private static ?

usnavi: Interesting.  I actually don't see the need for it.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/AbstractElasticIO.java
ChandraAddala: This is existing code. I just refactored it. I will add try/finally blocks in the code I added which follow this pattern.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/AbstractElasticIO.java
```diff
@@ -0,0 +1,463 @@
...
+     *      "aggs": {
+     *          "metric_name_tokens": {
+     *              "terms": {
+     *                  "field" : "metric_name",
+     *                  "include": "<regex>",
+     *                  "size": 0
+     *              }
+     *          }
+     *      }
+     *  }
```
shintasmith: I like it that you put sample request here.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/AbstractElasticIO.java
```diff
@@ -0,0 +1,463 @@
...
+             * immediate sublevels, in this case only foo.bar. If you look at indices above these foo
+             * index points to 2 documents.
+             *
+             * Notice how foo.bar has a doc_count which is higher than its immediate sublevels, in this
+             * case it is foo.bar.baz. Its higher because foo.bar is a metric in itself. We can use this
+             * logic to determine of a given index is a complete metric name or not.
+             *
+             * "buckets" : [ {
+             *     "key" : "foo",
+             *     "doc_count" : 2
```
shintasmith: Why in this case the doc_count for foo is 2? Shouldn't it be 3?

ChandraAddala: foo is pointing to two documents: foo.bar.baz, foo.bar

ChandraAddala:
@usnavi @shintasmith Refactored the main getNextTokens method.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/AbstractElasticIO.java
```diff
@@ -0,0 +1,501 @@
...
+     * 1) Next level of tokens for a given prefix which also have subsequent next level (not checking for enums yet)
+     * 2) For the metric names which do not fall under above category, identifies if any of them are complete metric
+     *    names at a given prefix level(LEVEL_0) or at a next level(LEVEL_1).
+     *
+     * @param response ES response
+     * @param prefix
+     * @param tokensWithNextLevel`WithNextLevel  Next level of tokens for a given prefix which also have subsequent next level
+     * @param completeMetricNamesByLevel contains complete metric names.
+     */
+    private void computeNextLevelOfTokens(final SearchResponse response,
```
shintasmith: Maybe name this method: buildMapOfMetricsNameByLevel() or something like that?

ChandraAddala: This method does two things...1) computeNextLevelOfTokens 2) buildMapOfMetricsNameByLevel :)

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/AbstractElasticIO.java
```diff
@@ -0,0 +1,501 @@
...
+             *  partialMetrics0LevelMap     : {"foo" -> "2"} (Contains all indexes at level 0)
+             *  partialMetrics0LevelAggMap  : {"foo" -> "2"} (Contains aggregate of doc_count of immediate sub levels,
+             *                                                in this case foo.bar)
+             *
+             *  partialMetrics1LevelMap     : {"foo.bar" -> "2"} (Contains all indexes at level 1)
+             *  partialMetrics1LevelAggMap  : {"foo.bar" -> "1"} (Contains aggregate of doc_count of immediate sub
+             *                                                    levels, in this case foo.bar.baz)
+             *
+             *  tokensWithNextLevel      : {'bar'}
+             */
```
usnavi: :+1: for the example.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/AbstractElasticIO.java
```diff
@@ -0,0 +1,501 @@
...
+
+        if (prefix != null && !prefix.isEmpty()) {
+
+            //prefix is a glob and in glob's "." is not a special character.
+            totalTokensInPrefix = getTotalTokensInPrefix(prefix);
+        }
+
+        //0 indicates ground level. if prefix is foo.bar, 0 is at foo.bar level
+        final Map<String, Long> partialMetrics0LevelMap = new LinkedHashMap<String, Long>();
+        final Map<String, Long> partialMetrics0LevelAggMap = new LinkedHashMap<String, Long>();
```
usnavi: How about mapLevel0IndexesCount & mapLevel0IndexesNextLevelCount?  My main focus is to describe what's actually stored in the map, not just what's being used as the key.  Same for the Level 1 map, too.

usnavi: Also, would be be cleaner to use one map with a data structure, say Map<String, LevelCounts>?  LevelCounts contains 2 fields, totalCount and subLevelsCount.  This would also allow the producing method to return this map, rather than having to pass it through the params.

blueflood-elasticsearch/src/test/java/com/rackspacecloud/blueflood/io/ElasticIOTest.java
```diff
@@ -296,4 +231,132 @@ public void testDeDupMetrics() throws Exception {
         elasticIO.setINDEX_NAME_READ(ElasticIOConfig.ELASTICSEARCH_INDEX_NAME_READ.getDefaultValue());
         elasticIO.setINDEX_NAME_WRITE(ElasticIOConfig.ELASTICSEARCH_INDEX_NAME_WRITE.getDefaultValue());
     }
+
+
+    @Test
+    public void testGetNextTokenWithoutPrefix1() throws Exception {
```
usnavi: could you improve these names?

blueflood-elasticsearch/src/test/java/com/rackspacecloud/blueflood/io/EnumElasticIOTest.java
```diff
@@ -297,25 +297,409 @@ public void testBatchQueryWithWildCards() throws Exception {
...
+        Set<String> expectedResults = new HashSet<String>() {{
+            add("a|true");
+            add("b|true");
+        }};
+
+        verifyTokenAndNextLevelFlag(results, expectedResults);
+    }
+
+    @Test
+    public void testGetNextTokenForEnumsWithPrefix1() throws Exception {
```
usnavi: improve names.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/HttpMetricNameTokensHandler.java
```diff
@@ -0,0 +1,109 @@
...
+        // get the prefix param
+        List<String> prefixRequestParam = requestWithParams.getQueryParams().get("prefix");
+
+        String prefix = "";
+        if (prefixRequestParam != null) {
+
+            if (prefixRequestParam.size() != 1) {
+                sendResponse(ctx, request, "Invalid request parameter: prefix",
+                        HttpResponseStatus.BAD_REQUEST);
+                return;
```
usnavi: does the timer need to be stopped at this return statement?

blueflood-http/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/HttpMetricNameTokensHandler.java
```diff
@@ -0,0 +1,109 @@
...
+            } else {
+                prefix = prefixRequestParam.get(0);
+            }
+        }
+
+        if (discoveryHandle == null) {
+            sendResponse(ctx, request, null, HttpResponseStatus.NOT_FOUND);
+            return;
```
usnavi: same here.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/AbstractElasticIO.java
ChandraAddala: If I understand correctly, you want me to use something like this Map<String, Tuple>. I dont believe java has a data structure like that. I would have to create another class to hold the two counts. I guess I can create one.

ChandraAddala: Also I am doing major refactoring. I will update my PR once it is ready.

ChandraAddala:
@shintasmith @usnavi I did another level of refactoring, pls take a look.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/MetricIndexData.java
```diff
@@ -0,0 +1,204 @@
...
+ *     "key" : "foo.bar.baz.aux",
+ *     "doc_count" : 1
+ * } ]
+ *
+ * If we feed each metric index to this class by calling the add(metricIndex, docCount) method,
+ * this class analyzes the data and provides methods to retrieve different aspects of this data
+ * with respect to baseLevel(if set to 2 with above data, uses foo.bar as reference).
+ *
+ */
+public class MetricIndexData {
```
usnavi: this class encapsulates the functionality a lot better.  I'm on the cusp of understanding the algorithm!

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/MetricIndexData.java
```diff
@@ -0,0 +1,204 @@
...
+        /**
+         *
+         * For the ES response shown in class description, for a baseLevel of 2,
+         * the data is classified as follows
+         *
+         * tokensWithNextLevelSet    -> {bar}               (tokens after baseLevel which also have a subsequent level)
+         *
+         * Both these maps maintain data in the form {metricIndex, (actualDocCount, childrenTotalDocCount)}
+         *
+         * metricIndexSameLevelMap   -> {foo.bar -> (2, 2)}     (all indexes which are of same length as baseLevel)
```
usnavi: Should the value of this be (3, 2)? 3 for foo.bar, foo.bar.baz & foo.bar.baz.aux & 2 for foo.bar.baz & foo.bar.baz.aux?

ChandraAddala: It only goes one level down. Just the total count of immediate children.

usnavi: cool, can we make it clear that children means immediate children?  That's probably the strict definition, but I assume others might make the same mistake?

blueflood-http/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/HttpMetricNameTokensHandler.java
ChandraAddala: Dont think so. Its a bad request. We dont need metrics for that. Atleast thats how it is handled in some existing handlers.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/MetricIndexData.java
ChandraAddala: It is mentioned near the logic which makes use of it (getCompleteMetricNames). Also if u think a metric name as hierarchy of paths(like a tree) as we are treating them now, if we count children of parents, it is only the immediate children. One level deep, they become grand children.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/AbstractElasticIO.java
```diff
@@ -0,0 +1,321 @@
...
+        } finally {
+            multiSearchCtx.stop();
+        }
+
+
+        for (SearchHit hit : response.getHits().getHits()) {
+            SearchResult result = convertHitToMetricDiscoveryResult(hit);
+            results.add(result);
+        }
+        return dedupResults(results);
```
shintasmith: This is fine, but to me it's strange that you build a result list above and then trim the dups. Why not just build non-duped list in the first place? It's possible I'm missing a glaring fact. :-)

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/MetricIndexData.java
usnavi: OK, i guess with the docCount being a sum of all paths, including grandchildren, that's where I've gotten confused.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/MetricIndexData.java
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    /**
+     * For a given metricIndex and docCount, classifies the data with respect to baseLevel
+     * and stores it accordingly.
+     *
+     * @param metricIndex
+     * @param docCount is document count
+     */
+    public void add(String metricIndex, long docCount) {
```
shintasmith: The name metricIndex confused me briefly, until I see how this method is being called. It's really a metric name. Maybe a simple comment (in the @param metricIndex) would help?

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/MetricIndexData.java
shintasmith: I like it!

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/TokenInfoListBuilder.java
```diff
@@ -0,0 +1,72 @@
...
+     * if foo.bar.baz is an enum metric in itself with enum values of say [something]
+     *      tokensWithEnumsAs2LevelMap  -> {baz, true}
+     *
+     * if there is another regular metric say at level 1, say foo.bar.test (no enum values)
+     *      tokensWithEnumsAs2LevelMap  -> {baz, true}, {test, false}
+     */
+
+    private final Map<String, Boolean> tokensWithNextLevelMap = new LinkedHashMap<String, Boolean>();
+    private final Map<String, Boolean> enumValuesAs1Level1Map = new LinkedHashMap<String, Boolean>();
+    private final Map<String, Boolean> tokensWithEnumsAs2LevelMap = new LinkedHashMap<String, Boolean>();
```
shintasmith: I don't completely understand why we are again storing the tokens with the next level here in a Map. It seems like you could go straight from ```metricIndexData.getTokensWithNextLevel()``` to a list of TokenInfo objects, no?

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/AbstractElasticIO.java
ChandraAddala: This is existing code. I just put that into a new method during refactoring.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/MetricIndexData.java
ChandraAddala: As discussed it is actually metric index.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/TokenInfoListBuilder.java
ChandraAddala: There could be duplicates...lets say we have two metrics, foo.bar.baz.qux1, foo.bar.baz.qux2, if we use prefix as foo.bar, we get both these metric names as indexes and the output TokenInfo should have only one {baz, true}.

usnavi:
+1

VinnyQ:
:+1: 

Looks great.  This is a big PR with good refactoring and tests.  Good job Chandra!

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/TokenInfoListBuilder.java
```diff
@@ -0,0 +1,68 @@
...
+     *      tokensWithEnumsAs2LevelMap  -> {baz, true}
+     *
+     * if there is another regular metric say at level 1, say foo.bar.test (no enum values)
+     *      tokensWithEnumsAs2LevelMap  -> {baz, true}, {test, false}
+     */
+
+    private final Set<TokenInfo> tokensWithNextLevelSet = new LinkedHashSet<TokenInfo>();
+    private final Set<TokenInfo> enumValuesAs1LevelSet = new LinkedHashSet<TokenInfo>();
+    private final Map<String, Boolean> tokensWithEnumsAs2LevelMap = new LinkedHashMap<String, Boolean>();
+
```
shintasmith: :+1: these look much simpler and cleaner. Thanks!

shintasmith:
lgtm, @ChandraAddala . Merge away

PR #611, opened 2016-02-03
Using powermock to mock static methods in BluefloodServiceStarter

ChandraAddala:


usnavi:
yeah, this is definitely the way to go :+1:

shintasmith:
I love it! :+1: 

cc: @izrik

PR #613, opened 2016-02-03
Remove unused, ZooKeeper-related things

izrik:
This PR removes code and configuration items related to ZooKeeper. I gather that the original plan was to use ZK to coordinate some work on shards; however, it looks like it was never used. There aren't any zookeeper instances in staging or prod, that I could find.

The two tests that supposed to test it were explicitly disabled in the pom file, so they never run. I took them out as well.

After removing the unused code, this puts our code coverage at about 73%.


izrik:
@drenalin23 @shintasmith @ChandraAddala @VinnyQ @usnavi Please review and comment.

blueflood-udp/src/main/java/com/rackspacecloud/blueflood/service/udp/MainIngestor.java
```diff
@@ -100,7 +100,7 @@ private static void checkConfiguration() {
 
         log.info("Configuration is good");
 
-        boolean useZookeeper = !"NONE".equals(config.getStringProperty(CoreConfig.ZOOKEEPER_CLUSTER));
+        boolean useZookeeper = false;
         log.info("Using zookeeper? " + useZookeeper);
```
usnavi: can we delete the boolean and this log statement?  I think that's the only use of the variable.

usnavi:
After this is accepted we'll need to create a story to remove any ZK-related stuff out of blueflood-chef.  Looks like there are some minor references with rollup_redundancy which we can remove.

blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/service/ScheduleContextIntegrationTest.java
```diff
@@ -1,76 +0,0 @@
-/*
```
shintasmith: Why is the ScheduleContextIntegrationTest removed?

shintasmith: I think ScheduledContext is a real class that we still use. Maybe we shouldn't be removing the whole file, but only the parts that uses ZooKeeper?

shintasmith: Oh, I see that the 2 tests in ScheduledContext are related to ZooKeeper only. There are no other tests for ScheduledContext. 😱 I am fine adding the real test for ScheduledContext in a different story, if needed.

shintasmith:
After you address @usnavi 's comment, I'm good with you merging this. +1

blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/service/ScheduleContextIntegrationTest.java
izrik: There is also a unit test called `ScheduledContextTest`. This integration test only tests the stuff related to ZK.

blueflood-udp/src/main/java/com/rackspacecloud/blueflood/service/udp/MainIngestor.java
izrik: Can do.

izrik: @usnavi Done.

usnavi:
lgtm

usnavi:
@izrik - the issue with travis is fixed.  Please rebase to add the fix to your branch and cause travis to rerun the build.

izrik:
Done.

PR #614, opened 2016-02-03
fix(blueflood-elasticsearch): Make init-es script take cli args

kaustavha:
This changes the init-es.sh script so it can take command line args in case our elasticsearch server isnt on localhost.
We want to try running blueflood with objectrockets ES offering instead of local ES.


shintasmith:
lgtm :+1:

PR #615, opened 2016-02-04
Updating bf-rollups-delay.py to support new blueflood installations 

usnavi:
```bf-rollups-delay.py``` had an assumption that every slot already had an active & rollup entry in the metrics_state table.  This PR updates the script not to error out when a slot doesn't have an active or rollup entry.


usnavi:
@shintasmith @VinnyQ @ChandraAddala - please review.

shintasmith:
lgtm

shintasmith:
but @ChandraAddala is the official review master

PR #616, opened 2016-02-04
fixing blueflood-http travis integration test run

usnavi:
Setting the max heap and permsize to prevent blueflood-http integration tests from failing during travis build.


usnavi:
@shintasmith @ChandraAddala @VinnyQ @izrik - may i get an amen?

shintasmith:
lgtm

PR #617, opened 2016-02-04
Add username & password & auth & absolute path

kaustavha:
Username, password and auth: Used to supply creds to auth reqs to a SaaS ES like objectRocket. If username and password arent supplied auth defaults to empty and the `-u` header isnt added to the ES curl commands.  

absolutepath allows us to run `init-es.sh` from any directory, without it you need to be in the `blueflood-elasticsearch/src/main/resources` directory before you can run `./init-es.sh`.

### DIFF
--- a/blueflood-elasticsearch/src/main/resources/init-es.sh
+++ b/blueflood-elasticsearch/src/main/resources/init-es.sh
@@ -4,13 +4,25 @@
 #  are required to properly tokenize the incoming data.
 # You can pass it a url if your elastic search doesn't reside on localhost:9092
 #WARNING: this script will destroy existing ES data and reset it to the proper init state
-
+# Example:
+#    Local ES: ./init-es.sh
+#    ES on OR: ./init-es.sh <url for OR:ES> <username> <password>
 ELASTICSEARCH_URL=${1:-"http://localhost:9200"}
+ES_USERNAME=${2:-}
+ES_PASSWD=${3:-}
 
+# Set a auth header for curl if username and passwd is supplied
+# Allows initing ES when ES is supplied by a SaaS provider like ObjectRocket
+if [ $ES_USERNAME ] && [ $ES_PASSWD ]; then
+    AUTH="-u ${ES_USERNAME}:${ES_PASSWD}"
+fi
+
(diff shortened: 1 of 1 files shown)


usnavi:
@kaustavha  - the issue with travis is fixed.  Please rebase to add the fix to your branch and cause travis to rerun the build.

kaustavha:
@usnavi Rebased

blueflood-elasticsearch/src/main/resources/init-es.sh
```diff
@@ -4,13 +4,25 @@
 #  are required to properly tokenize the incoming data.
 # You can pass it a url if your elastic search doesn't reside on localhost:9092
 #WARNING: this script will destroy existing ES data and reset it to the proper init state
-
+# Example:
+#    Local ES: ./init-es.sh
+#    ES on OR: ./init-es.sh <url for OR:ES> <username> <password>
 ELASTICSEARCH_URL=${1:-"http://localhost:9200"}
+ES_USERNAME=${2:-}
+ES_PASSWD=${3:-}
```
usnavi: Having 3 command line params, should we should be using getopts or another command line parser?

http://wiki.bash-hackers.org/howto/getopts_tutorial

kaustavha: This one dude on SO was quite against getopts and linked this:
http://mywiki.wooledge.org/BashFAQ/035#getopts

3 isnt that many imo, and there expected to be empty for 50% of the use-case i.e. the default use case.
But ill play around with getopts a bit and see how it feels in this scenario, might make sense. 
Alternatively it might make sense to have 2 command line args, with the second being <username>:<password>. That also eliminates a step in this script.

kaustavha:
Hey @usnavi I added getopts and a usage output

usnavi:
+1 thanks @kaustavha for the update!

usnavi:
@shintasmith @ChandraAddala - I'd like to merge this in.  Would one of you take a look?

ChandraAddala:
Looks good.

shintasmith:
lgtm, for what it's worth 😆

PR #623, opened 2016-02-07
Loading indices to ES

goru97:
*Why:*
To fix the error given by BF on Querying for the List of metrices.


contrib/blueflood-docker/configs/ES-Setup/index_settings.json
```diff
@@ -0,0 +1,27 @@
+{
```
ChandraAddala: Is there anyway we can avoid having multiple copies of these files?

ChandraAddala: The original one is located at ./blueflood-elasticsearch/src/main/resources/index_settings.json

goru97: @ChandraAddala I know, but I wanted the blueflood-docker repo to be independent of other packages.

ChandraAddala: If the goal to is to keep this independent, I guess it makes sense, but I still feel if these files change, someone has to remember to change them here.

goru97:
@ChandraAddala You are right, but these files rarely change as they define schemas in ES and the other files are configs (blueflood.conf, properties), which people can adjust to taste.

contrib/blueflood-docker/configs/ES-Setup/index_settings.json
usnavi: Does this need to be independent?  Is there a use case for having our docker example load up a different schema than our actually install?

contrib/blueflood-docker/configs/ES-Setup/init-es.sh
```diff
@@ -0,0 +1,79 @@
+#!/bin/bash
```
usnavi: Is this copied from blueflood-elasticsearch/src/main/resource/init-es.sh?  Same copy questions as before.

goru97: Yes! These are copied from blueflood-elasticsearch/src/main/resource/init-es.sh. Schema rarely changes; it's the blueflood.conf which people usually modify.

usnavi:
@goru97 - let's have a quick chat today and see if we can reference some of hte existing scripts.

goru97:
@ChandraAddala @usnavi Can I merge this for now. I will write a new dockerfile, which will be more according to the standards, very configurable and will also use the resources which are already there in ./blueflood-elasticsearch/src/main/resources. But people are using this repo and are saying it's broken!

usnavi:
Sure, you can merge it now....as long as you promise.... :-P

goru97:
I keep my promises :smiley:

PR #624, opened 2016-02-09
Changing timer name

ChandraAddala:


blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/AbstractElasticIO.java
```diff
@@ -27,7 +27,7 @@
 
     // todo: these should be instances per client.
     protected final Timer searchTimer = Metrics.timer(getClass(), "Search Duration");
-    protected final Timer getNextTokensTimer = Metrics.timer(getClass(), "getNextTokens Duration");
+    protected final Timer getNextTokensTimer = Metrics.timer(getClass(), "getNextTokens ES metric index query Duration");
```
VinnyQ: NIT:  that's a crazy name, sure you want it like that?  
Also I think the other metric labels are capitalized each word except for minor words.

VinnyQ: I guess there's a few cases that's a long name with lower casing, so nevermind.

VinnyQ:
:+1:

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/AbstractElasticIO.java
ChandraAddala: How about this? "ES Metric Index Query Duration"

PR #625, opened 2016-02-09
Improving performance of Rollup On Read

usnavi:
Adding metrics to ascertain the performance of rollups-on-reads.
Breaking up the cassandra calls and rollups into workloads of no more than ROLLUP_ON_READ_REPAIR_SIZE_PER_THREAD ranges.
The workloads are then run in parallel.


blueflood-core/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/RollupHandler.java
```diff
@@ -287,24 +305,109 @@ private Boolean repairMetrics (Locator locator, MetricData metricData, final lon
...
+
+        ListenableFuture<List<List<Points.Point>>> aggregateFuture = Futures.allAsList(futures);
+
+        try {
+            for( List<Points.Point> subList : aggregateFuture.get(rollupOnReadTimeout.getValue(), rollupOnReadTimeout.getUnit()) ) {
+
+                repairedPoints.addAll( subList );
+            }
+        } catch (Exception e) {
+            log.warn(String.format("Exception encountered while doing rollups on read, incomplete rollups will be returned. %s", e.getMessage()));
```
shintasmith: Why not log the exception? By doing e.getMessage() we don't get the stack trace.

usnavi:
@shintasmith @ChandraAddala @VinnyQ 
- Updated the exception logging
- Added more comments for the integration tests
- Updated tests to verify the rolled up values as well as the keys

blueflood-core/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/RollupHandler.java
```diff
@@ -129,6 +143,8 @@ private rollupsOnReadTimers (Timer timer) {
         List<SearchResult> units = null;
         List<Locator> locators = new ArrayList<Locator>();
 
+        Timer.Context c = timerRorCalcUnits.time();
```
shintasmith: So there is already a timer for getRollupByGranularity() method: the ```plotTimers.SPLOT_TIMER.timer.time()``` and ```plotTimers.MPLOT_TIMER.timer.time()```. The ```timerRorCalcUnits.time()``` is a subset of that. Essentially, it looks like plotTimers == timerRorCalcUnits + rollupsOnReadTimer

There's only a small piece of code, a for loop, that is not covered by other timers.

Just throwing the question, do you think it's useful to keep those plotTimers?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/RollupHandler.java
```diff
@@ -287,24 +304,109 @@ private Boolean repairMetrics (Locator locator, MetricData metricData, final lon
...
+
+        ListenableFuture<List<List<Points.Point>>> aggregateFuture = Futures.allAsList(futures);
+
+        try {
+            for( List<Points.Point> subList : aggregateFuture.get(rollupOnReadTimeout.getValue(), rollupOnReadTimeout.getUnit()) ) {
+
+                repairedPoints.addAll( subList );
+            }
+        } catch (Exception e) {
+            log.warn("Exception encountered while doing rollups on read, incomplete rollups will be returned.", e);
```
shintasmith: Would caller get any indication that they are getting an incomplete rollup?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/RollupHandler.java
usnavi: I added this to ascertain the exact time spent accessing elastic search as it can take up to 100-200 ms in some cases.

I felt more comfortable being able to reference an exact metric rather than calculating the differences of other ones.

shintasmith:
Other than the 2 questions above, I don't have any more comments. Lgtm +1

blueflood-core/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/RollupHandler.java
usnavi: I don't think so, they just get less data.  The same holds for our current prod settings where MPlot rollups-on-read are not calculated, and a subset of data is returned.

Since this is an error case, I'm up for altering the behavior, but I'd say lets do that in a different story.

shintasmith: I am fine with that. I was just curious what's the behavior.

shintasmith:
@usnavi :+1:

blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/outputs/RollupHandlerIntegrationTest.java
```diff
@@ -25,129 +27,278 @@
...
+        return sum / count;
+    }
+
+    private double meanOfPointCollectionRoll( Collection<Points.Point<BasicRollup>> fullPoints ) {
+        long sum = 0;
+        int count = 0;
+        for( Points.Point<BasicRollup> p : fullPoints ) {
+
+            sum += p.getData().getAverage().toLong();
+            count++;
```
ChandraAddala: Why do you need count? Cant you use fullPoints.size?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/RollupHandler.java
ChandraAddala: What kinda exception are you expecting here? Just curious.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/RollupHandler.java
```diff
@@ -287,24 +304,109 @@ private Boolean repairMetrics (Locator locator, MetricData metricData, final lon
...
+     * @return A list of iterables of ranges, each iterable list is no bigger than ROLLUP_ON_READ_REPAIR_SIZE_PER_THREAD
+     */
+    private List<Iterable<Range>> divideRangesByGroup( Granularity g, long from, long to ) {
+        List<Iterable<Range>> listRange = new ArrayList<Iterable<Range>>();
+
+        long rangeSize = ROLLUP_ON_READ_REPAIR_SIZE_PER_THREAD * g.milliseconds();
+
+        for ( long start = g.snapMillis( from ); start < to ; ) {
+
+            long end = (start + rangeSize) < to ? (start + rangeSize ) : to;
```
ChandraAddala: If I am understanding this correctly, If you convert both "from" and "to" their corresponding snapMillis, you dont have to do this logic right? You dont have to do it. Just wondering.

usnavi: This method takes a ```from``` & ```to`` time and returns a list of ranges which encompass this time.  No range is bigger than ``ROLLUP_ON_READ_REPAIR_SIZE_PER_THREAD * g.milliseconds()```  Each range will later be processed in its own thread.

Does that answer your question?

ChandraAddala: My question was more about how its done. Why was ``g.snapMillis( from )`` done only for ``from`` but not for ``to``. I see now that it is being done inside ``rangesForInterval`` method. I guess you dont have to do it for ``from`` either, but thats fine.

usnavi: i understand you know.  i'll remove the ```g.snapMillis( from )``` as its unneeded.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/RollupHandler.java
usnavi: ```aggregateFuture.get()``` throws InterruptedException, ExecutionException, TimeoutException

usnavi:
@ChandraAddala - updated the PR as per your comments.

ChandraAddala:
+1

PR #626, opened 2016-02-11
make integration tests work within intellij

shintasmith:
Before we start adding more integration tests, we thought it would be nice to have a mechanism to run these tests from IntelliJ.

The main changes to make that possible are:
* adding ```src/test/resources``` directory to Maven's test resources. This will cause files in that directory to be copied to ```target/test-classes```, which then we can load the files from classpath, instead of relative path. The relative path breaks because you have to manually set the Working Directory setting of each test that you are running from IntelliJ to the submodule directory. 
* leverage pluginManagement and dependencyManagement so that all integration tests across multiple submodules share the same versions of plugins/dependency.


shintasmith:
@usnavi @ChandraAddala @VinnyQ @izrik please review.

cc: @drenalin23

usnavi:
+1 thanks for getting to the bottom of this.

ChandraAddala:
Looks good.

PR #627, opened 2016-02-12
fix jacoco-it javaagent JVM parameter getting overwritten

shintasmith:
**Why:**
I noticed that we are no longer generating the file ```jacoco-it.exec```, which is the raw coverage output of jacoco-maven-plugin. Our coveralls numbers are also lower, from 70% to now 56%.

**How:**
Fix the maven failsafe plugin configuration so that all JVM arguments is in a single &lt;argLine&gt; parameter. The plugin does not take multiple &lt;argLine&gt; configuration. It just takes the last one. So the last parameter overwrites the jacoco.agent JVM argument.


usnavi:
+1

PR #629, opened 2016-02-13
Test ScheduleContext

izrik:
This PR adds tests for `ScheduleContext` class. It increases the code coverage for that class from 70% tp 79%. It also adds several tests that demonstrate the purpose and effects of methods of `ScheduleContext`.


izrik:
@shintasmith @usnavi @VinnyQ @ChandraAddala Please review. In particular, I'm concerned about  duplication of setup code. I'm thinking of extracting methods out to separate test classes/fixtures. For example, the `testGetNextScheduled*` methods would be moved to a new `ScheduleContextGetNextScheduledTest` class, with the common setup code refactored into a `@Begin setup()` method. Same for `testPushBackToScheduled*`, etc. Would anyone have an objection to that?

usnavi:
+.5, @shintasmith and @ChandraAddala are the true +1's.

shintasmith:
I think it would be good for us to learn what ```ScheduleContext``` is for, or at least what your understanding of what it does. Is there any write ups? I think you mentioned you are working on one such write up?

blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/ScheduleContextTest.java
```diff
@@ -385,4 +386,697 @@ public void testRecentlyScheduledShards() {
...
+        long now = 1234000L;
+        long updateTime = now - 2;
+        int shard = ringShards.get(0);
+        Granularity gran = Granularity.MIN_5;
+        int slot = gran.slot(updateTime);
+        Granularity coarserGran = null;
+        try {
+            coarserGran = gran.coarser();
+        } catch (GranularityException e) {
+            Assert.fail("Couldn't get the next coarser granularity");
```
shintasmith: Should we bubble up the exception so we have exception stack trace?

shintasmith:
Other than the exception question, everything else looks good. I am also ok if you want to split the tests to 2 Test classes.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/ScheduleContextTest.java
```diff
@@ -385,4 +386,697 @@ public void testRecentlyScheduledShards() {
...
+
+        // when
+        ctx.update(now, ringShards.get(0));
+
+        // then
+        UpdateStamp stamp;
+
+        stamp = mgr.getUpdateStamp(slotkey5);
+        Assert.assertNotNull(stamp);
+        Assert.assertEquals(UpdateStamp.State.Active, stamp.getState());
```
ChandraAddala: Can we create a method for this? It seems to be used in a lot of places.

```
Assert.assertNotNull(stamp);
Assert.assertEquals(UpdateStamp.State.Active, stamp.getState());
Assert.assertEquals(now, stamp.getTimestamp());
Assert.assertTrue(stamp.isDirty());
```

blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/ScheduleContextTest.java
izrik: Sure. I think that would look cleaner, too.

izrik:
@shintasmith in re what `ScheduleContext` is for: Yes, I am going to produce a write-up and some diagrams. The class's chief purpose is to coordinate access to the list of slots that need to be re-rolled between `RollupService`, `IngestionService`, and Cassandra. There's several other classes involved, but that's the gist of it. Also, since we run rollup and ingestion on separate nodes these days, mediating between them is not as important. It mostly goes like this:

1. Ingestion node marks a slot as dirty
2. Ingestion node persists the dirty flag to Cassandra
3. Rollup node pulls the change to the dirty flag into memory
4. Rollup node does the rollup thing and mark the slot not dirty anymore
5. Rollup node persists the not-dirty flag to Cassandra

blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/ScheduleContextTest.java
izrik: @ChandraAddala Ok, I was able to extract a method. I don't really like it. It's not very pretty or elegant, in my opinion. What do you think.

izrik:
@shintasmith @ChandraAddala @usnavi I think that's it. Any other suggestions?

blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/ScheduleContextTest.java
ChandraAddala: Its no big deal if you leave it the way it is.

ChandraAddala:
Looks good.

shintasmith:
lgtm, thanks for the explanation of ScheduledContext.

izrik:
@shintasmith @ChandraAddala @usnavi @VinnyQ 
Documentation commits!

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/UpdateStamp.java
```diff
@@ -49,16 +61,27 @@ public boolean equals(Object o) {
...
+     * not finished and should, therefore, not be considered rolled.
+     */
     public enum State {
-        // in the database, there are only two states we care about: Active and Rolled.  Running is a ephemeral state
-        // during runtime.  It degrades to Active during a save restore (indicating it is not finished and should,
-        // therefore, not be considered rolled.
-        Active("A"), Running("A"), Rolled("X");
+
+        Active("A"),
+        Running("A"),
```
shintasmith: Are you sure Running is also "A"? I thought we do  have "R" state

shintasmith:
Other than clarification for the state, it looks good! Thank you @izrik for documenting this.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/ScheduleContext.java
```diff
@@ -44,13 +44,63 @@
...
+ * {@link #getSlotStamps(Granularity, int)},
+ * {@link #getRecentlyScheduledShards()},
+ * {@link #getMetricsState(int, String, int)}) which either aren't used, are
+ * only used for testing, or are used to provide information to external system
+ * via JMX or yammer metrics.
+ * <p>
+ *
+ *
+ *
+ * keeps track of dirty slots in memory. Operations must be threadsafe.
```
usnavi: Can we capitalize the K?  I'm crazy, but while reading, since it was a sentence fragment, I thought the front half the sentence got cut off.  Confused me for a bit, especially since complete sentences are used above it.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/UpdateStamp.java
izrik: Yes, I left that value exactly as it originally appeared. I didn't write the enum or the comment, I just re-formated them. My impression is that the state gets serialized to the value when persisted to the database, and then when it gets pulled out of the database it gets deserialized to `Active`. I think that's what it means when it says "degrades to Active during a save restore".

shintasmith: Ok sounds good.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/ScheduleContext.java
izrik: oops. Saw this after I merged it. I'll do that and open a second PR.

usnavi:
other than my crazy K, lgtm

PR #630, opened 2016-02-18
Update grinder configs with proportional load

usnavi:
Updating the grinder configs for perf & staging to proportionally match the production load.  Did not modify qe01/qe02 as they have different node types.

based on load per production db node.

This is using the load/db node numbers of
* 200K metrics/minute ingest
* 7/minute queries


usnavi:
FYI - I didn't calibrate the properities for qe01 & qe02 as their nodes are a different size than staging/perf/prod.

contrib/grinder/properties/grinder.properties.qe01
```diff
@@ -17,7 +17,7 @@ grinder.bf.metrics_per_tenant=210
 grinder.bf.batch_size=150
 grinder.bf.ingest_concurrency=25
 grinder.bf.annotations_concurrency=5
-grinder.bf.num_nodes=1
+grinder.bf.num_nodes=2
```
drenalin23: What does this number of nodes refer to?  We only have one shim00 server in qe01/qe02 environments.

drenalin23:
Other than my inquiry about what that nodes attribute means - LGTM.

usnavi:
@drenalin23 - staging & perf01 both had 2 nodes, so i thought that was the number across all envs.  I've updated the PR as per your comments.

drenalin23:
+1

PR #631, opened 2016-02-18
fixing ROR when TURN_OFF_RR_MPLOT is true

usnavi:
The ```createRepairPointsExecutor``` is used when ```TURN_OFF_RR_MPLOT``` is ```true``` or ```false```.  Ensuring that this object is always created.

Unfortunately, there is no way to verify this in our unit tests since the configuration is pulled from the ```Configuration``` singleton.  We could pull the singleton references out and pass in the constructor, but I think that is more than we want to do now as the ```RollupHandler``` is sub-classed a few times.

Verified this on perf01 with ```TURN_OFF_RR_MPLOT``` as ```true```.

### DIFF
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/RollupHandler.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/RollupHandler.java
@@ -102,13 +102,11 @@ public class RollupHandler {
             rollupsOnReadExecutor = MoreExecutors.listeningDecorator(rollupsOnReadExecutors);
         }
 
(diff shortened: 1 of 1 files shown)


ChandraAddala:
@usnavi Does that mean you will be creating those threads based on  ROLLUP_ON_READ_REPAIR_SIZE_PER_THREAD even if it is not ROR?

usnavi:
ROR always runs for SPlot, its only ROR for MPlot which is turned on and off.

SPlot can get the same speed up as MPlot, but since there aren't the performance issues, its not as critical.

ChandraAddala:
But doing all those threads for Splot is not necessary right, if splot does not have perf issues?

ChandraAddala:
I understand it now. Looks good to me.

PR #632, opened 2016-02-23
Updating staging & perf01 grinder numbers to match new query rates in prod

usnavi:
Now that lbaas has stopped their queries we are now running 1 mplot/minute per db node.


drenalin23:
+1

PR #634, opened 2016-02-28
BFF code cleanup.

ChandraAddala:
Inorder for me to make changes to talk to the new API, I felt the existing code/setup needs some cleanup. In that process I had made the following changes.

1) Created vagrant configuration so that we can run grafana with blueflood finder code pointing to local blueflood run.
2) Added logging capability. I cant deal with print statements.  Also figured out a way to enable these logs in staging/prod servers. 
3) Fixed the current enum bug that exists in prod while populating drop down for enum metrics.
4) Organized dependencies(setup.py, test_requirements.txt).
5) Wrote brand new tests to capture the behavior of finder for various scenarios. 

I learned a little bit about the grafana server setup that happens with the heat template. I had documented all that info. I would appreciate if you can review the wiki as well along with this.

[https://one.rackspace.com/display/cloudmetrics/Blueflood+Finder+grafana+server+internal+details](Grafana server setup with blueflood finder)


contrib/graphite/tests/test_blueflood.py
```diff
@@ -0,0 +1,254 @@
+from unittest import TestCase, skip
```
usnavi: Unless the tests already existed as python, I would have suggested using groovy since the team is more familiar with that, but this is fine.

usnavi: nm, i see that you are testing python functions here, I get it.

contrib/graphite/tests/test_blueflood.py
```diff
@@ -0,0 +1,254 @@
...
+                                 'one.two.three00.fourD',
+                                 'one.two.three00.fourE'}  # fourE has enum values which makes it a branch node
+        expected_leaf_nodes = {'one.two.three00.fourA',
+                               'one.two.three00.fourD',
+                               'one.two.three00.fourB'}  # metric one.two.three00 which has fourA, fourB as enum values
+
+        self.verify_find_nodes_call(prefix, expected_branch_nodes, expected_leaf_nodes,
+                                    self.bff, bf_mock_response, bf_enum_mock_response)
+
+    # @skip("Does not work as expected")
```
usnavi: is this test valid?

ChandraAddala: it is. I removed that line.

contrib/graphite/blueflood.py
```diff
@@ -56,8 +62,9 @@ def calc_res(start, stop):
 class TenantBluefloodFinder(threading.Thread):
   __fetch_multi__ = 'tenant_blueflood'
   __fetch_events__ = 'tenant_blueflood'
+
   def __init__(self, config=None):
-    print("Blueflood Finder v31")
+    logger.debug("Blueflood Finder v31")
```
shintasmith: what is the significance of the version 31? The setup.py says it's version 0.0.1

contrib/graphite/blueflood.py
```diff
@@ -268,15 +289,17 @@ def find_nodes(self, query):
...
 
     except Exception as e:
-     print "Exception in Blueflood find_nodes: "
-     print e
+     logger.exception("Exception in Blueflood find_nodes: ")
      exc_info = sys.exc_info()
      tb = traceback.format_exception(*exc_info)
      for line in tb:
-       print(line)
+       logger.debug(line)
```
shintasmith: Nit-pick, I'm a noob on this. But reading the python doc, 
```

Logger.exception(msg, *args, **kwargs)

    Logs a message with level ERROR on this logger. The arguments are interpreted as for debug(), except that any passed exc_info is not inspected. Exception info is always added to the logging message. This method should only be called from an exception handler.
```
It seems like exception info is always added? So do we still need to loop over the lines and print them as logger.debug() ?

contrib/graphite/README.md
```diff
@@ -0,0 +1,46 @@
...
+    source bf-finder/bin/activate
+    pip install .
+ 
+## Development environment on vagrant
+
+We can setup grafana server on a vagrant vm which visualizes local blueflood using BF finder. To spin up 
+vagrant vm use the below commands.
+
+    vagrant box add ubuntu/trusty64
+    vagrant up
```
usnavi: does this have to be executed from ```contrib/graphite```?

usnavi:
Other than clarifying my vagrant question, lgtm.

shintasmith:
me too... looks good other than the logging clarifications.

contrib/graphite/README.md
ChandraAddala: Yes. Ideally this contrib/graphite should be in its own project. We should talk about it and do it as part of the next story which is about creating a pypi package.

contrib/graphite/blueflood.py
ChandraAddala: I have no idea..its existing code.

contrib/graphite/blueflood.py
ChandraAddala: You are right. I didnt pay much attention. I was replacing print's with logger statements. I will change that code.

PR #640, opened 2016-03-03
externalize the Rollup On Read timeout

shintasmith:
**Why**
Travis running integration tests sometimes failed on ```HttpMultiRollupsQueryHandlerIntegrationTest.testMultiplotQuery()```. System.out.println debug statements showed the test was getting TimeoutException

**How**
Externalize the hard coded timeout to Config so we can set it to something other than 10 seconds. We keep the default as 10 seconds.


shintasmith:

usnavi:
lgtm - i don't think we need to update blueflood-chef with this parameter.  What do you say?

shintasmith:
@usnavi, I agree. I wasn't planning on updating it. It should just stay the default 10 seconds.

ChandraAddala:
looks good.

PR #641, opened 2016-03-03
add oraclejdk18 to Travis

shintasmith:
**What**
Add oraclejdk1.8 to Travis build so we can catch jdk 1.8 build errors during PR. Openjdk18 is not yet supported in Travis:
https://docs.travis-ci.com/user/languages/java#Testing-Against-Multiple-JDKs

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: java
 jdk:
   - oraclejdk7
   - openjdk7
+  - oraclejdk8
 
 sudo: false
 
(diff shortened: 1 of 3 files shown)


shintasmith:
@usnavi @ChandraAddala @VinnyQ @izrik , I finally got Travis to be happy. I had to move the setting of the ROR timeout to the base IntegrationTest class. Otherwise it doesn't get set properly. I think it's ok for all test classes to use this timeout. It will only affect tests related to Rollups.

usnavi:
lgtm - once this is merged I'll rebase my PR and run against java8 as well.

PR #642, opened 2016-03-03
setting proportional load for 10/minute queries per node

usnavi:
Setting perf01 & staging to the correct numbers.


shintasmith:
lgtm

PR #643, opened 2016-03-04
New api /metric_name/search

ChandraAddala:
I had made changes to API as documented in the wiki https://one.rackspace.com/display/cloudmetrics/Blueflood+finder+performance+issues#Bluefloodfinderperformanceissues-APIdesign(final)

     for metric names: foo.bar.xxx,
                       foo.bar.baz.qux,
                       foo.bar with enum values [one, two]
     
     for query=foo.bar.*, returns the below list of metric names
     
     {"foo.bar.xxx", true}
     {"foo.bar.baz", false} 
     {"foo.bar.one", true}   <- From enum metric foo.bar.baz
     {"foo.bar.two", true}   <- From enum metric foo.bar.baz

To produce these results, for query = foo.bar.\*, we query ES to get the below metric indexes.
1) Previous level(foo.bar): If there are complete metrics at this level, ".\*" in the query could be for an enum value.
2) Current level(foo.bar.\*): If there are complete metrics at this level, is_leaf would be false if it is an enum metric.
3) Next level(foo.bar.\*.\*). To get the next level of token and also to figure out if it has subsequent next level.


blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/MetricToken.java
```diff
@@ -0,0 +1,69 @@
...
+        if (o == null || getClass() != o.getClass()) return false;
+
+        MetricToken metricToken = (MetricToken) o;
+
+        if (isLeaf != metricToken.isLeaf) return false;
+        return !(path != null ? !path.equals(metricToken.path) : metricToken.path != null);
+    }
+
+    @Override
+    public int hashCode() {
```
usnavi: :+1 my first thought when I saw you overrode ```equals``` was, "Did he override ```hashcode``` too?"

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/AbstractElasticIO.java
```diff
@@ -101,48 +104,56 @@
...
+     * for metric names: foo.bar.xxx,
+     *                   foo.bar.baz.qux,
+     *                   foo.bar with enum values [one, two]
+     *
+     * for query=foo.bar.*, returns the below list of metric tokens
+     *
+     * new MetricToken("foo.bar.xxx", true)   <- From metric foo.bar.xxx
+     * new MetricToken("foo.bar.baz", false)  <- From metric foo.bar.baz.qux
+     * new MetricToken("foo.bar.one", true)   <- From metric foo.bar.baz
+     * new MetricToken("foo.bar.two", true)   <- From metric foo.bar.baz
```
usnavi: The final 2 examples talk the enums from the metric ```foo.bar.baz``` - shouldn't it be ```foo.bar```?

ChandraAddala: Yes. I have to fix that.

usnavi:
lgtm

shintasmith:
lgtm too

PR #645, opened 2016-03-04
configure finder code to be able to package and upload to pypi

VinnyQ:
Add configuration files and modify setup.py to package and upload to https://pypi.python.org so that package can be installed using `pip install blueflood-graphite-finder`.

See feedback/request @ https://feedback.rackspace.com/forums/345663-metrics-as-a-service/suggestions/12353130-get-the-blueflood-python-module-required-by-grap

### DIFF
--- a/contrib/graphite/README.md
+++ b/contrib/graphite/README.md
@@ -5,7 +5,12 @@ The BF finder is the graphite plugin that allows graphite and grafana to use blu
 
 ### Setup
 
-Get the [blueflood](https://github.com/rackerlabs/blueflood) repo from github. Execute the following commands
+To install package from pypi using pip:
+
+    pip install blueflood-graphite-finder
+
+To install manually with code from github repo:
+    Get the [blueflood](https://github.com/rackerlabs/blueflood) repo from github. Execute the following commands
 
     cd $BLUEFLOOD_REPO_LOCATION/contrib/graphite
     virtualenv bf-finder
--- a/contrib/graphite/setup.py
+++ b/contrib/graphite/setup.py
@@ -4,15 +4,14 @@ from setuptools import setup
 version = '0.0.1'
 
 setup(
-  name='blueflood',
+  name='blueflood-graphite-finder',
   version=version,
-  url='https://github.com/rackerlabs/blueflood',
+  url='https://github.com/rackerlabs/blueflood/tree/master/contrib/graphite',
   license='APL2',
-  keywords='blueflood graphite metrics',
-  author='Gary Dusbabek',
-  author_email='gdusbabek@gmail.com',
-  description=('A plugin for using graphite-web with the cassandra-based '
-               'Blueflood storage backend'),
+  keywords='blueflood graphite finder metrics',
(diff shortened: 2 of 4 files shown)


VinnyQ:
@shintasmith @usnavi @ChandraAddala please review.

glyph:
:+1: , ,generally, but a couple of minor points:

1. Keywords aren't super useful, since google searches can find words in your description.  However, trove classifiers have semantic, unambiguous meaning; you can find a list here: https://pypi.python.org/pypi?%3Aaction=list_classifiers
2. It's generally considered poor form to include multiple modules in a PyPI package; if the package is named `blueflood-graphite-finder`, then users will expect it to be importable as `import blueflood_graphite_finder` or perhaps `import blueflood.graphite.finder`; they definitely won't expect `import auth` to come from there.  So `py_modules` should be reorganized at some point to move into a single package.
3. Is there any particular reason the package is not `zip_safe`?  Just as a personal thing I like to see packages use the relevant packaging APIs to allow for zip safety; among other things this makes simplified (self-contained, single-executable) deployments easier.

contrib/graphite/setup.py
```diff
@@ -4,15 +4,15 @@
...
-  keywords='blueflood graphite metrics',
-  author='Gary Dusbabek',
-  author_email='gdusbabek@gmail.com',
-  description=('A plugin for using graphite-web with the cassandra-based '
-               'Blueflood storage backend'),
+  keywords='blueflood graphite finder metrics',
+  author='Rackspace Metrics',
+  author_email='cloudMetrics-dev@lists.rackspace.com',
+  description=('A plugin for using graphite-web with '
+               'the cassandra-based Blueflood storage backend'),
```
ChandraAddala: Its a "blueflood plugin for graphite-api and graphite-web".

contrib/graphite/README.md
```diff
@@ -5,7 +5,12 @@ The BF finder is the graphite plugin that allows graphite and grafana to use blu
 
 ### Setup
 
-Get the [blueflood](https://github.com/rackerlabs/blueflood) repo from github. Execute the following commands
+To install uploaded package to pypi using pip:
```
ChandraAddala: Do you mean install "from" pypi?

ChandraAddala:
@VinnyQ We were thinking of moving blueflood finder code to a separate repo. May be this is a good time to do that.

VinnyQ:
@glyph thanks for the feedback.  I discussed it with the team and we are all for moving these code into a standalone repo.  I'll coordinate with @ChandraAddala to do that since he currently has work in progress.  Will see what i can do about cleaning up the dependency also.

contrib/graphite/setup.py
VinnyQ: Good call.  Changed.

contrib/graphite/README.md
VinnyQ: that is a better wording, fixed.

ChandraAddala:
Looks good to me.

PR #646, opened 2016-03-07
initial clean up of Astyanax ColumnFamily

shintasmith:
**Why**
Eventually, we will be moving to using Datastax driver and removing Astyanax, in preparation to upgrading to Cassandra 2.1.13. 

This PR removes some unnecessary use of Astyanax's ColumnFamily model class. Blueflood already provides a wrapper class called ```MetricColumnFamily```, which extends the Astyanax ColumnFamily class. But for some reasons some code are written to return Astyanax ColumnFamily directly, propagating the use of this class in more places than necessary.

**How**
* Change the CassandraModel.getColumnFamily() method to return MetricColumnFamily. 
* Change all consumer of this method accordingly.

A small other clean up I did was to move the class ```IntegrationBaseTest.java``` from ```src/main``` to ```src/integration-test```. To make this move successful, I have to also change ```.travis.yml```. Before the PR, we were calling ```mvn package``` before running the tests. I think that is incorrect. We should package *after* tests are running successfully.

I have run the unit & integration tests and they all have passed.


ChandraAddala:
Tests seem to be failing. Other than that everything looks good.

shintasmith:
Yeah... I'm working on it. One of the base class IntegrationBaseTest.java was placed in src/main. I thought I would move it to the right place, which is src/integration-test. But Travis is not happy yet..

shintasmith:
@ChandraAddala @usnavi @VinnyQ @izrik @drenalin23 pls review

usnavi:
lgtm

ChandraAddala:
Looks good.

shintasmith:
thanks all

izrik:
@shintasmith I've run into this problem before. I found that the `integration-test` phase comes after the `package` phase. Integration tests can't be run before then. More info here: https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html#Lifecycle_Reference

shintasmith:
@izrik, I have fixed this

PR #648, opened 2016-03-16
Test module loader

izrik:
This PR adds tests for the `ModuleLoader` class, in the `blueflood-core` module. There are already tests for it in the `blueflood-http` module, but since that's different than the module that the class is defined in, coverage is being reported as 0% (see [here](https://coveralls.io/builds/5442321/source?filename=blueflood-core%2Fsrc%2Fmain%2Fjava%2Fcom%2Frackspacecloud%2Fblueflood%2Futils%2FModuleLoader.java)). Also, these tests better demonstrate the behavior of the class, and provide pertinent negative test cases.

According to JaCoCo, the test class, by itself, provides coverage for `ModuleLoader` to the tune of:
* 75% for instructions 
* 100% for branches
* 78% for lines
What's not tested is exception handling and the default parameter-less constructor.

On the whole, `ModuleLoader` appears to be a thin, confusing, and broken wrapper around `ClassLoader.loadClass(String)`. I believe it should be refactored away.

### DIFF
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/utils/ModuleLoader.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/utils/ModuleLoader.java
@@ -37,7 +37,7 @@ public class ModuleLoader {
         List<String> modules = Configuration.getInstance().getListProperty(moduleName);
         if (modules.isEmpty())
             return null;
-        if (!modules.isEmpty() && modules.size() != 1) {
+        if (modules.size() != 1) {
             throw new RuntimeException("Cannot load service with more than one "+moduleName+" module");
         }
 
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/utils/ModuleLoader.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/utils/ModuleLoader.java
@@ -16,6 +16,7 @@
 
 package com.rackspacecloud.blueflood.utils;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.rackspacecloud.blueflood.service.Configuration;
 import com.rackspacecloud.blueflood.service.CoreConfig;
 import org.slf4j.Logger;
@@ -63,5 +64,10 @@ public class ModuleLoader {
         }
 
         return moduleInstance;
+    }
+
+    @VisibleForTesting
+    public static void clearCache() {
+        loadedModules.clear();
     }
 }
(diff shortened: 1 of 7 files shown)


izrik:
@VinnyQ @ChandraAddala @shintasmith @usnavi Please review and comment. I'll merge once I get +1 and any concerns that come up are addressed.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/utils/ModuleLoaderTest.java
```diff
@@ -0,0 +1,183 @@
...
+                CoreConfig.DISCOVERY_MODULES,
+                "com.rackspacecloud.blueflood.io.DiscoveryIO");
+
+        // when
+        Object loadedModule = ModuleLoader.getInstance(DiscoveryIO.class, CoreConfig.DISCOVERY_MODULES);
+
+        // then
+        Assert.assertNull(loadedModule);
+    }
+}
```
ChandraAddala: Can you also test a scenario where the class does not have a default(no-arg) constructor?

ChandraAddala:
Looks good.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/utils/ModuleLoader.java
```diff
@@ -36,7 +37,7 @@ public static Object getInstance(Class c, CoreConfig moduleName) {
         List<String> modules = Configuration.getInstance().getListProperty(moduleName);
         if (modules.isEmpty())
             return null;
-        if (!modules.isEmpty() && modules.size() != 1) {
+        if (modules.size() != 1) {
```
ChandraAddala: Good refactoring.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/utils/ModuleLoaderTest.java
ChandraAddala: Also for a class which is package-protected and is from a different package than the test case.

izrik: @ChandraAddala Done and done.

izrik:
@ChandraAddala How does it look now?

blueflood-core/src/test/java/com/rackspacecloud/blueflood/utils/ModuleLoaderTest.java
```diff
@@ -0,0 +1,214 @@
...
+        // when
+        Object loadedModule = ModuleLoader.getInstance(DiscoveryIO.class, CoreConfig.DISCOVERY_MODULES);
+
+        // then
+        Assert.assertNotNull(loadedModule);
+        Assert.assertEquals(DummyDiscoveryIO3.class, loadedModule.getClass());
+    }
+
+    @Test
+    public void singleClassInDifferentPackageYieldsThatClass() {
```
ChandraAddala: I am thinking of a test case where it would throw "IllegalAccessException". I believe for that we have to pick a class from a different package which is at package-protected access level or may be a private class to trigger that exception.

izrik:
@ChandraAddala How about now?

ChandraAddala:
Looks good.

PR #649, opened 2016-03-22
refactor common logic out of astyanax specific classes

shintasmith:
**Why**
To prep work for writing the new Datastax driver, I need to gut out common logic of serializing/deserializing various objects out of Astyanax specific classes to generic SerDes classes. This is so that the Datastax driver classes can re-use the generic SerDes classes -- this is necessary for us to be able to read data that's written to Cassandra by Astyanax, using Datastax driver classes.

**How**
The work on this PR is **purely** moving code around. There is no additional logic added. The following are the gist of the moves:
* all Astyanax specific Serializer classes that are inner classes in ```NumericSerializer``` have been converted just wrappers of SerDes classes under ```io.serializers.metrics``` package. The actual logic of converting objects => bytes, and bytes => objects is moved to these individual SerDes classes for each object. The only exception is the RawSerDes, which is the serializer/deserializer of several primitives objects.
* ```NumericSerializer``` is renamed to ```Serializers```
* all the conversions from class => Constants.* => serialized bytes, via the method NumericSerializer.typeOf(), have been reduced to class => serialized bytes. The methods typeOf() and sizeOf() are removed. The NumericSerializer.Type eventually will be removed too. But right now there is a test that's still referencing it. I will take a look at changing the test code in follow up PRs.
* all of json related serializer/deserializer classes have been moved to ```io.serializers.json``` package for clarity. The code in these classes are not changed.
* all test code have been refactored in similar manner to be consistent with the main code.

No additional test code has been added for the new SerDes classes. I will add a few in subsequent PRs.

Unit and integration tests have been run locally and they are passing.


blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/serializers/Serializers.java
```diff
@@ -0,0 +1,218 @@
...
+import com.netflix.astyanax.serializers.AbstractSerializer;
+import com.rackspacecloud.blueflood.exceptions.SerializationException;
+import com.rackspacecloud.blueflood.io.serializers.astyanax.HistogramSerializer;
+import com.rackspacecloud.blueflood.io.serializers.metrics.*;
+import com.rackspacecloud.blueflood.types.*;
+
+import java.nio.ByteBuffer;
+
+public class Serializers {
+    // NumericSerializer can be used with Rollup and full resolution metrics.
```
usnavi: if you want to get picky you can make a private constructor to make sure an instance is not created, but not required.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/serializers/astyanax/HistogramSerializer.java
```diff
@@ -0,0 +1,42 @@
...
+package com.rackspacecloud.blueflood.io.serializers.astyanax;
+
+import com.netflix.astyanax.serializers.AbstractSerializer;
+import com.rackspacecloud.blueflood.io.serializers.metrics.HistogramSerDes;
+import com.rackspacecloud.blueflood.types.HistogramRollup;
+
+import java.nio.ByteBuffer;
+
+public class HistogramSerializer extends AbstractSerializer<HistogramRollup> {
+    private static final HistogramSerDes deSer = new HistogramSerDes();
```
usnavi: I'd recommend doing a private constructor on this.  Not sure if this kind of stuff was in the scope of the PR, but I'll comment on it.  The private constructor will prevent anyone from creating their own instance rather than using the singleton.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/serializers/metrics/AbstractSerDes.java
```diff
@@ -0,0 +1,76 @@
...
+            protobufOut.writeDoubleNoTag(stat.toDouble());
+        } else {
+            protobufOut.writeRawVarint64(stat.toLong());
+        }
+    }
+
+    // these two sizeOf*() methods are created mainly
+    // for code readability
+    protected int sizeOfSize() {
+        return 1;
```
usnavi: do we want to make these abstract?

usnavi:
I assume moving the other Astyanax-related classes such as AstyanaxIO, AstyanaxMetadataIO, etc will be moved in another PR?

usnavi:
Other than questions about if we want to use private constructors on static classes and abstract methods, lgtm

usnavi:
One more thing: didn't comment on the actual code, I figure you might be updating it in a later PR?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/serializers/astyanax/HistogramSerializer.java
shintasmith: Good idea. I will do that.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/serializers/metrics/AbstractSerDes.java
shintasmith: I actually don't want to make these abstract. I want all SerDes classes to share the logic that the size of the serialized bytes is one byte. Does that make sense?

shintasmith:
@usnavi, answers/replies to your questions:
1) Yes, the Astyanax-related classes such as the AstyanaxIO, AstyanaxMetadataIO, etc will move to a different package in later PR. I wasn't planning to change the code much, though.
2) I will mostly be changing the Serializers.java a lot. Because that is where the switch between Astyanax and Datastax will happen. But the rest of the Astyanax specific classes, I am not planning to do a lot of update.

shintasmith:
@usnavi I added the private constructors.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/serializers/astyanax/LocatorSerializer.java
```diff
@@ -28,7 +28,6 @@
     private static final LocatorSerializer instance = new LocatorSerializer();
```
ChandraAddala: to make it consistent can you name this variable in upper case(INSTANCE).

shintasmith: ok can do.

shintasmith: Done

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/serializers/Serializers.java
usnavi: I think this class and its sub classes should have private constructors.

shintasmith: @usnavi I added private constructors to this class and all of its inner classes.

usnavi:
lgtm

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/serializers/metrics/AbstractSerDes.java
ChandraAddala: May be you can write a comment explaining why it is "1"?

ChandraAddala: Also should it be final if the subclasses dont need to change it?

shintasmith: Good point. I added the final and comments.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/serializers/metrics/AbstractSerDes.java
```diff
@@ -0,0 +1,87 @@
...
+import com.rackspacecloud.blueflood.io.Constants;
+import com.rackspacecloud.blueflood.types.*;
+
+import java.io.IOException;
+
+/**
+ * This is an abstract class that most Serialization/Deserialization class
+ * extends from. It contains some utility methods.
+ */
+public abstract class AbstractSerDes {
```
ChandraAddala: I noticed that this class does not have any abstract methods. I hadnt seen such classes before but I guess if you dont want anyone to instantiate this class, this is how you do it. Also none of its children override anything from this class. So, this could very well be perceived as an utility class. 

What I would like though is,
1) either create an interface which contains methods serialize() and deserialize() 
2) or have abstract methods for serialize() and deserialize() here

I do see that its difficult to structure it that way because of classes like RawSerDes. So in summary I am not quire sure how to go about this but I feel like we arent leveraging polymorphism very well.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/serializers/Serializers.java
```diff
@@ -0,0 +1,254 @@
...
+            return (AbstractSerializer<T>)fullInstance;
+        else if (type.equals(Float.class))
+                return (AbstractSerializer<T>)fullInstance;
+        else if (type.equals(byte[].class))
+            return (AbstractSerializer<T>)fullInstance;
+        else if (type.equals(Object.class))
+            return (AbstractSerializer<T>)fullInstance;
+        else
+            return (AbstractSerializer<T>)fullInstance;   
+    }
```
ChandraAddala: In continuation to my comment for AbstractSerDes class, I think all the serializable types(BasicRollUP, SimpleNumber etc) should share a common interface. That would make code look much clean.

ChandraAddala:
Apart from the comment I made, everything looks good.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/serializers/Serializers.java
usnavi: @ChandraAddala - my assumption is that fixing these issues would be a separate endeavor.  My two cents.

shintasmith: Yes @ChandraAddala, cleaning up this part will need to be done later, after I have the corresponding Datastax SerDes classes. I have gutted the other methods that are using a lot of instanceOf, the typeOf() and sizeOf() methods, tho. :-)

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/serializers/metrics/AbstractSerDes.java
shintasmith: Yeah, you are right it's not so much of an Abstract class. It's more of a Base class. The SerDes subclasses share common logic such as the size of size and type in the serialized buffer. I can change the name of the class from AbstractSerDes to BaseSerDes.

I had in mind of creating an interface. But as you noticed on classes like RawSerDes, it's hard to do it. Also, not all objects to be serialized extends or implements a common class/interface. Those objects such as Average, Min, Max do share a parent class, but the Rollup objects share a different parent class. Changing those objects to have a common inheritance is a massive change, I feel the cost not quite justified yet.

shintasmith: @ChandraAddala and I had a side conversation. He's ok with changing the AbstractSerDes to BaseSerDes, and cleaning up the Serializers.serializerFor() in subsequent PR.

shintasmith: I am going to merge this PR. :-)

PR #650, opened 2016-03-23
Test LocatorFetchRunnable

izrik:
This PR breaks the `run` method up into several. New methods are `getLocators`, `processLocator`, `executeRollupForLocator`, `processHistogramForLocator`, `executeHistogramRollupForLocator`, `drainExecutionContext`, and `finishExecution`. All methods but `run` and `drainExecutionContext` are fully tested. I'd like some suggestions on testing the remaining two.

`drainExecutionContext` is particularly difficult, because it sleeps and expects another class to interrupt it. My current thought is to hand-write a subclass of `LocatorFetchRunnable`, instead of using mockito's `mock` method, and do manual verification of method calls. The reason `mock` won't work is because `LocatorFetchRunnable` instance fields are all final, and must be set in a constructor; mockito doesn't allow passing parameters to the constructor. If I manually create a mock, then I can pass the necessary arguments to the constructor via `super()`. Does that sound reasonable?


izrik:
@usnavi @shintasmith @VinnyQ @ChandraAddala

blueflood-core/src/main/java/com/rackspacecloud/blueflood/threading/ThreadPoolSizedExecutorServiceAdapter.java
```diff
@@ -0,0 +1,104 @@
...
+ * {@link ThreadPoolSizedExecutorServiceAdapter} and
+ * {@link SizedExecutorService} should be removed.
+ */
+public class ThreadPoolSizedExecutorServiceAdapter implements SizedExecutorService {
+
+    private final ThreadPoolExecutor executor;
+
+    public ThreadPoolSizedExecutorServiceAdapter(ThreadPoolExecutor executor) {
+        if (executor == null) throw new NullPointerException();
+        this.executor = executor;
```
usnavi: In the future, my preference is not to have a local variable and field have the same name (e.g., executor).  Others can chime in if they have opinions.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/threading/ThreadPoolSizedExecutorServiceAdapter.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    @Override
+    public int getActiveCount() {
+        return this.executor.getActiveCount();
```
usnavi: can we drop the ```this.``` on these calls?

blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/LocatorFetchRunnableTest.java
```diff
@@ -0,0 +1,366 @@
...
+        final List<Locator> locators = new ArrayList<Locator>() {{
+            add(locator1);
+            add(locator2);
+            add(locator3);
+        }};
+        return locators;
+    }
+
+    @Test
+    public void getLocatorsReturnsLocators() {
```
usnavi: Is this test that compelling?  If I understand it correctly, essentially ```getLocators()``` just calls ```astyanaxReader.getLocatorsToRollup()```.  Is that really worth testing?  Am I missing something?  The ```expected``` and ```actual``` are the same class since we are using mockito, right?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/threading/ThreadPoolSizedExecutorServiceAdapter.java
izrik: It's like that all over the code base, so I just went with it. But I understand the concern.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/LocatorFetchRunnableTest.java
izrik: I'm just trying to be thorough. It would seem a little odd to me to test the failure path (`getLocatorsExceptionYieldsEmptySet`) but not the success path.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/threading/ThreadPoolSizedExecutorServiceAdapter.java
usnavi: Let's not follow this convention.  I don't think the consistency buys us anything but more annoyance :-P

ChandraAddala: I do it this way all the time. Infact if you use intelliJ to write up your constructor, that how it does it. I am quite used to seeing it this way.

izrik: We should decide on a convention and run checkstyle, or something like that. We could run it in travis so that naming conventions are applied consistency and show up in PR's.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/threading/ThreadPoolSizedExecutorServiceAdapter.java
```diff
@@ -0,0 +1,104 @@
...
+ * {@link com.rackspacecloud.blueflood.service.RollupBatchWriter#enqueueRollupForWrite}
+ * makes calls to {@link #getActiveCount()} and {@link #getPoolSize()}, which
+ * means it must be hard-coded to use the {@link ThreadPoolExecutor} class
+ * instead of the {@link ExecutorService} interface. Once that method is fixed
+ * and no longer relies on those calls, it can be changed to use just
+ * {@link ExecutorService}, and both
+ * {@link ThreadPoolSizedExecutorServiceAdapter} and
+ * {@link SizedExecutorService} should be removed.
+ */
+public class ThreadPoolSizedExecutorServiceAdapter implements SizedExecutorService {
```
ChandraAddala: I dont see a compelling reason to create this class. I have a feeling we can get by without this class. I think leaving the original signature of LocatorFetchRunnable to have ThreadPoolExecutor as it was having before, is not such a bad thing if we can avoid creating this class.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/LocatorFetchRunnable.java
```diff
@@ -155,6 +141,14 @@ public void run() {
         if (log.isDebugEnabled())
             log.debug("Finished {} rollups for (gran,slot,shard) {} in {}", new Object[] {rollCount, parentSlotKey, System.currentTimeMillis() - waitStart});
 
+        finishExecution(waitStart, executionContext);
+    }
+
+    public void waitForRollups() throws InterruptedException {
+        Thread.currentThread().sleep(LOCATOR_WAIT_FOR_ALL_SECS * 1000);
+    }
```
ChandraAddala: Someone could totally mess with us by calling this repeatedly. We shouldnt expose these kinda methods. If you want control, have LOCATOR_WAIT_FOR_ALL_SECS in a separate constructor. For tests you can set it to zero if you dont want it to wait.

izrik: But it still needs to be overridden in `StubbedLocatorFetchRunnable` in order to confirm that it was called, and to artificially force it to throw `InterruptedException`. Avoiding the time delay is secondary.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/threading/ThreadPoolSizedExecutorServiceAdapter.java
```diff
@@ -0,0 +1,104 @@
...
+
+    private final ThreadPoolExecutor executor;
+
+    public ThreadPoolSizedExecutorServiceAdapter(ThreadPoolExecutor executor) {
+        if (executor == null) throw new NullPointerException();
```
shintasmith: It's strange to be throwing NullPointerException here. Maybe the better exception is IllegalArgumentException?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/LocatorFetchRunnable.java
ChandraAddala: Yeah There is nothing going on when InterruptedException gets thrown. What are you really trying to test here? Also waitForRollups just has code to sleep a thread.
            try {
                waitForRollups();
            } catch (InterruptedException ex) {
                if (log.isTraceEnabled())
                    log.trace("Woken wile waiting for rollups to coalesce for {} {}", parentSlotKey);
            } finally {
                String verb = executionContext.doneReading() ? "writing" : "reading";
                log.debug("Still waiting for rollups to finish {} for {} {}", new Object[] {verb, parentSlotKey, System.currentTimeMillis() - waitStart });
            }

blueflood-core/src/main/java/com/rackspacecloud/blueflood/threading/ThreadPoolSizedExecutorServiceAdapter.java
izrik: The goal was to be able to use interfaces in the constructors of the classes. If `RollupBatchWriter` were changed to not call `getActiveCount()` and `getPoolSize()`, then we could just use `ExecutorService`. I have removed the class.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/threading/ThreadPoolSizedExecutorServiceAdapter.java
izrik: done

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/LocatorFetchRunnable.java
izrik: I made it `protected`. Nevertheless, these were intended to be temporary refactorings, until we can fix the overall system. No one should be calling this.

ChandraAddala:
Looks good.

izrik:
@usnavi I've got one more method to test. Is there anything else that you feel still needs to be addressed in the code so far?

usnavi:
@izrik - nope, lgtm.  looks like my issue with ```this.``` is just me!

izrik:
@ChandraAddala @usnavi OK, I tested the `run` method. Instead of modifying the hand-written subclass of `LocatorFetchRunnable`, I managed to figure out how to use a regular mockito mock. It's still ugly, but it mostly works.

The tests don't do a fantastic job of testing the "problem domain" behind the class. My opinion is that said problem domain is not adequately modeled by the classes we have, but is instead interwoven across many classes in a very non-OOP fashion. However, these tests should serve as an adequate documentation of the existing behavior of the class. The coverage is certainly enough to facilitate refactoring of rollup functionality. 

These three new test classes by themselves provide about 92% line coverage for `LocatorFetchRunnable`.

ChandraAddala:
Looks good.

usnavi:
lgtm

PR #651, opened 2016-03-28
make Tracker a singleton class and register it for DataQueryServer also.

VinnyQ:
# What

Modify the TrackerMbean class implementation to be a singleton, and register it on both Ingest and Query server startup.

# How

Magic.

(i.e. look at the code).

## How to test

After deployment, verify that you can perform enabling/disabling the tracker (via the blueflood_tenant_tracker.sh script found in blueflood-chef/contrib) on both the ingest nodes and the query nodes.

# Why

So we can log http requests and responses on both the ingest nodes and the query nodes.

### DIFF
--- a/blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionServer.java
+++ b/blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionServer.java
@@ -32,7 +32,6 @@ import com.rackspacecloud.blueflood.io.EventsIO;
 import com.rackspacecloud.blueflood.io.IMetricsWriter;
 import com.rackspacecloud.blueflood.service.*;
 import com.rackspacecloud.blueflood.tracker.Tracker;
-import com.rackspacecloud.blueflood.tracker.TrackerMBean;
 import com.rackspacecloud.blueflood.types.IMetric;
 import com.rackspacecloud.blueflood.types.MetricsCollection;
 import com.rackspacecloud.blueflood.utils.ModuleLoader;
@@ -74,8 +73,6 @@ public class HttpMetricsIngestionServer {
 
     private ChannelGroup allOpenChannels = new DefaultChannelGroup("allOpenChannels");
 
-    public TrackerMBean tracker;
-
     public HttpMetricsIngestionServer(ScheduleContext context, IMetricsWriter writer) {
         this.httpIngestPort = Configuration.getInstance().getIntegerProperty(HttpConfig.HTTP_INGESTION_PORT);
         this.httpIngestHost = Configuration.getInstance().getStringProperty(HttpConfig.HTTP_INGESTION_HOST);
@@ -111,10 +108,10 @@ public class HttpMetricsIngestionServer {
         Channel serverChannel = server.bind(new InetSocketAddress(httpIngestHost, httpIngestPort));
         allOpenChannels.add(serverChannel);
 
-        log.info("Starting tracker service");
-        tracker = new Tracker();
+        //register the tracker MBean for JMX/jolokia
+        log.info("Registering tracker service");
+        Tracker.getInstance().register();
     }
-
 
     private class MetricsHttpServerPipelineFactory implements ChannelPipelineFactory {
         private RouteMatcher router;
--- a/blueflood-http/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/HttpMetricDataQueryServer.java
+++ b/blueflood-http/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/HttpMetricDataQueryServer.java
@@ -24,6 +24,7 @@ import com.rackspacecloud.blueflood.io.EventsIO;
 import com.rackspacecloud.blueflood.service.Configuration;
 import com.rackspacecloud.blueflood.service.CoreConfig;
 import com.rackspacecloud.blueflood.service.HttpConfig;
+import com.rackspacecloud.blueflood.tracker.Tracker;
 import com.rackspacecloud.blueflood.utils.ModuleLoader;
 import org.jboss.netty.bootstrap.ServerBootstrap;
 import org.jboss.netty.channel.ChannelPipeline;
@@ -79,6 +80,10 @@ public class HttpMetricDataQueryServer {
                         Executors.newFixedThreadPool(workerThreads)));
         server.setPipelineFactory(new MetricsHttpServerPipelineFactory(router));
         serverChannel =  (ServerChannel) server.bind(new InetSocketAddress(httpQueryHost, httpQueryPort));
+
+        //register the tracker MBean for JMX/jolokia
+        log.info("Registering tracker service");
+        Tracker.getInstance().register();
     }
 
     private class MetricsHttpServerPipelineFactory implements ChannelPipelineFactory {
--- a/blueflood-http/src/main/java/com/rackspacecloud/blueflood/tracker/Tracker.java
+++ b/blueflood-http/src/main/java/com/rackspacecloud/blueflood/tracker/Tracker.java
@@ -38,14 +38,40 @@ public class Tracker implements TrackerMBean {
 
     private static final Logger log = LoggerFactory.getLogger(Tracker.class);
     private static final String trackerName = String.format("com.rackspacecloud.blueflood.tracker:type=%s", Tracker.class.getSimpleName());
-    private static final Pattern patternGetTid = Pattern.compile( "/v\\d+\\.\\d+/([^/]+)/.*" );
+    private final Pattern patternGetTid = Pattern.compile( "/v\\d+\\.\\d+/([^/]+)/.*" );
 
-    static Set tenantIds = new HashSet();
-    static boolean isTrackingDelayedMetrics = false;
-    static Set<String> metricNames = new HashSet<String>();
+    // Tracker is a singleton
+    private static final Tracker instance = new Tracker();
+    private boolean isRegistered = false;
 
-    public Tracker() {
-        registerMBean();
+    private Set tenantIds = new HashSet();
+    private boolean isTrackingDelayedMetrics = false;
+    private Set<String> metricNames = new HashSet<String>();
+
+    // private constructor for singleton
+    private Tracker(){}
+
+    public static Tracker getInstance() {
+        return instance;
+    }
+
+    public synchronized void register() {
+        if (isRegistered) return;
+
+        try {
+            ObjectName objectName = new ObjectName(trackerName);
+
+            // register TrackerMBean only if not already registered
+            MBeanServer mBeanServer = ManagementFactory.getPlatformMBeanServer();
+            if (!mBeanServer.isRegistered(objectName)) {
+                ManagementFactory.getPlatformMBeanServer().registerMBean(instance, objectName);
+            }
+
+            isRegistered = true;
+
+        } catch (Exception exc) {
+            log.error("Unable to register mbean for " + instance.getClass().getSimpleName(), exc);
+        }
     }
 
     public void addTenant(String tenantId) {
@@ -88,11 +114,11 @@ public class Tracker implements TrackerMBean {
         log.info("[TRACKER] All metric names removed.");
     }
 
-    public static boolean isTracking(String tenantId) {
+    public boolean isTracking(String tenantId) {
         return tenantIds.contains(tenantId);
     }
 
-    public static boolean doesMessageContainMetricNames(String logmessage) {
+    public boolean doesMessageContainMetricNames(String logmessage) {
         boolean toLog = false;
 
         if (metricNames.size() == 0) {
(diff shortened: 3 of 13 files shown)


VinnyQ:
@usnavi @shintasmith @ChandraAddala please review, thanks!

blueflood-http/src/main/java/com/rackspacecloud/blueflood/tracker/Tracker.java
```diff
@@ -40,12 +40,27 @@
     private static final String trackerName = String.format("com.rackspacecloud.blueflood.tracker:type=%s", Tracker.class.getSimpleName());
     private static final Pattern patternGetTid = Pattern.compile( "/v\\d+\\.\\d+/([^/]+)/.*" );
 
+    // Tracker is a singleton
+    private static Tracker instance = new Tracker();
```
ChandraAddala: Can you make this final?

usnavi:
lgtm

blueflood-http/src/main/java/com/rackspacecloud/blueflood/tracker/Tracker.java
```diff
@@ -40,12 +40,27 @@
...
+    // private constructor for singleton
+    private Tracker(){}
+
+    public static Tracker getInstance() {
+        register();
+        return instance;
+    }
+
+    public static void register() {
+        if (!isRegistered) {
```
ChandraAddala: You have to add a synchronized block here. If not, there is a chance it will be registered multiple times and we can get `InstanceAlreadyExistsException`.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/HttpMetricDataQueryServer.java
```diff
@@ -79,6 +80,9 @@ public void startServer() {
                         Executors.newFixedThreadPool(workerThreads)));
         server.setPipelineFactory(new MetricsHttpServerPipelineFactory(router));
         serverChannel =  (ServerChannel) server.bind(new InetSocketAddress(httpQueryHost, httpQueryPort));
+
+        log.info("Starting tracker service");
+        Tracker.register();
```
ChandraAddala: When you call `register()` from both these place, whats happen differently between the calls. Inside the method, both these calls seem to be doing the same thing. If there is no difference, can you call `register()` only in the constructor of Tracker?

blueflood-http/src/main/java/com/rackspacecloud/blueflood/tracker/Tracker.java
```diff
@@ -40,12 +40,27 @@
...
-        registerMBean();
+    // private constructor for singleton
+    private Tracker(){}
+
+    public static Tracker getInstance() {
+        register();
+        return instance;
+    }
+
+    public static void register() {
```
shintasmith: I don't understand this and the getInstance() method. If you want to make this a singleton, then lets make all methods be instance method (i.e: not static). Just make the above ```public void register()``` and make the caller do this:
```
Tracker.getInstance().register()
```
Remember, it is easier to mock instances than static methods.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/tracker/Tracker.java
VinnyQ: I'll add it but I don't think this will ever happen as you described.  Tracker is not ran in multitude threads (only registered by two servers service on startup), and the register() method only register the MBean, not instantiate the object.  The object is only instantiate once on class loading.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/HttpMetricDataQueryServer.java
VinnyQ: I did it this way to avoid those class having to declare and instantiate a Tracker object without having to use it for anything.  The code looks cleaner this way.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/tracker/Tracker.java
VinnyQ: @shintasmith 

I used singleton so that multiple servers classes (HttpMetricsIngestionServer & HttpMetricDataQueryServer) can use the same global object without having to create their own.
And I used static methods so that we don't have to pass the objects around to all the individual handlers.  I don't think they have to be mutually exclusive, and I think in this scenario it makes the code a lot simpler to read.  i.e. If I make these methods non-static, I'll have to grab the instance (with Tracker.getInstance()) for every single one of the handlers that needs it (HttpAggregatedIngestionHandler, HttpAggregatedMultiIngestionHandler, HttpEventsIngestionHandler, HttpMetricsIngestionHandler, HttpMultitenantMetricsIngestionHandler, HttpEventsQueryHandler, HttpHistogramQueryHandler, HttpMetricsIndexHandler, HttpMetricTokensHandler, HttpMultiRollupsQueryHandler, HttpRollupsQueryHandler), resulting to a lot more line of codes.  Remember this is more like a "logging" class, not a class that we have to make sure data are absolutely correct and atomic (where singletons *should* be retrieve by instance).  I think the tests setup we have for this class is sufficient (in TrackerTest), we don't need to mock a tracker class anywhere, even when testing the handlers code (similar to other singleton utilities such as Date and Loggers). 

I'll do it though if you think that's more correct.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/HttpMetricDataQueryServer.java
VinnyQ: Actually, let me try moving the register() code to a constructor to see if it works without any instantiation.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/tracker/Tracker.java
VinnyQ: Ok I thought it through and I see your point, I'll make it non-static with a more direct Tracker.getInstance.track() call on all the handlers, to avoid using static methods.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/HttpMetricDataQueryServer.java
ChandraAddala: If your class is a singleton, it gets instantiated during application startup or when classes are getting loaded depending upon whether you are doing lazy instantiation. Either way, if you move the register call inside constructor, it should work. No need to call the register() method from anywhere.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionServer.java
```diff
@@ -111,11 +108,10 @@ public void startServer() {
         Channel serverChannel = server.bind(new InetSocketAddress(httpIngestHost, httpIngestPort));
         allOpenChannels.add(serverChannel);
 
-        log.info("Starting tracker service");
-        tracker = new Tracker();
+        log.info("Registering tracker service");
+        Tracker.getInstance();
```
VinnyQ: Note that if I don't have this call here, the track() laters still work (all tests past and all code functionalities within blueflood will work as intended), but we will not be able to see the MBean on service startup (because it hasn't been registered yet) and will not be able to make any of the calls to the MBean (such as addTenant, addMetricName, setIsTrackingDelayedMetrics) until at least one .track() call is made, at which time the registration is called from the private constructor.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionServer.java
```diff
@@ -111,11 +108,11 @@ public void startServer() {
         Channel serverChannel = server.bind(new InetSocketAddress(httpIngestHost, httpIngestPort));
         allOpenChannels.add(serverChannel);
 
-        log.info("Starting tracker service");
-        tracker = new Tracker();
+        //register the tracker MBean for JMX/jolokia
+        log.info("Registering tracker service");
+        Tracker.getInstance();
```
VinnyQ: Note that if I don't have this call here in the ingest/queryServer.startServer() method, the track(), trackResponse(), and trackDelayedMetricsTenant() in the handlers still work (all code functionalities within blueflood will still work as intended), but we will not be able to see the MBean in JMX/jolokia on service startup because it hasn't been registered yet, and will not be able to make any of the calls to the MBean (such as addTenant, addMetricName, setIsTrackingDelayedMetrics) until at least one of the call is made in the ingest or query handlers (at which time the registration is made from the private constructor).

Doing it here in startServer() register the MBean preemptively.

VinnyQ:
@shintasmith @ChandraAddala can you take a look at this again?  I took your advice and made the singleton non-static, which seems safer.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/HttpMetricDataQueryServer.java
```diff
@@ -79,6 +80,10 @@ public void startServer() {
                         Executors.newFixedThreadPool(workerThreads)));
         server.setPipelineFactory(new MetricsHttpServerPipelineFactory(router));
         serverChannel =  (ServerChannel) server.bind(new InetSocketAddress(httpQueryHost, httpQueryPort));
+
+        //register the tracker MBean for JMX/jolokia
+        log.info("Registering tracker service");
+        Tracker.getInstance();
     }
```
ChandraAddala: I dont think you need to call getInstance(). It only returns the instance. register() already happened when the instance is created during class loading time since its static.

VinnyQ: I tested that and it doesn't, not sure why.  It'll though once one of the handler call .track or .trackResponse.

To test:  started BluefloodServiceStarter class, started VisualVM (a JMX tool) and looked for com.rackspacecloud.blueflood.tracker, it's not there.  Then I submitted a request through the ingestion server and got a 200, refresh VisualVM and com.rackspacecloud.blueflood.tracker shows up.

I then added the Tracker.getInstance call to the server and restarted BluefloodServiceStarted (which calls the Server.startServer() methods).  Open up VisualVM and it shows up.

JVM must not call the Tracker private constructor until something reference it in the code, is what I am seeing.

VinnyQ: this was why I added the register() call earlier, because of the behavior I was seeing.  Maybe if I add it back the code will read better.  Unless you know of a way to define Tracker so it does call the private constructor non-lazily?

shintasmith: I prefer an explicit call if you need to make register() happen. Implicit things that happen behind the scene makes code a bit less readable.

ChandraAddala: We can leave it the way it is.

ChandraAddala:
Looks good.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/tracker/Tracker.java
```diff
@@ -37,36 +37,42 @@
...
+                log.error("Unable to register mbean for " + instance.getClass().getSimpleName(), exc);
             }
-        } catch (Exception exc) {
-            log.error("Unable to register mbean for " + this.getClass().getSimpleName(), exc);
+
+            isRegistered = true;
         }
     }
 
     public static Tracker getInstance() {
```
shintasmith: Do we need this method? It seems like people should just call Tracker.register() ?

blueflood-http/src/main/java/com/rackspacecloud/blueflood/tracker/Tracker.java
```diff
@@ -38,14 +38,42 @@
...
-    public Tracker() {
-        registerMBean();
+    private Set tenantIds = new HashSet();
+    private boolean isTrackingDelayedMetrics = false;
+    private Set<String> metricNames = new HashSet<String>();
+
+    // private constructor for singleton
+    private Tracker(){}
+
+    public static synchronized void register() {
```
shintasmith: @VinnyQ , why don't we just make people call it like this?
```
Tracker.getInstance().register()
```

It is more explicit. When I read a code like that, the code tells me two information:
* Tracker class is a singleton
* register() is an instance method that registers Tracker to some registry.

Compare that to call like this:
```
Tracker.register()
```
The instance creation is hidden and implicit. It's less clear of what this Tracker class is doing.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/tracker/Tracker.java
VinnyQ: Then in the handlers code, everywhere would read:

Tracker.register().track(request);
Tracker.register().trackResponse(request, response);

which makes it more confusing.

Our other option is to rename register() to be called registerAndGetInstance()... which look more funky.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/tracker/Tracker.java
VinnyQ: good point, I can do this.

shintasmith:
lgtm +1 :+1:

PR #652, opened 2016-03-28
Removing blueflood-kafka & blueflood-udp projects as they aren't used anymore.

usnavi:
```blueflood-kafka``` and ```blueflood-udp``` aren't used so we are removing them from the codebase so all existing code is relevant.


ChandraAddala:
I dont know what these projects do but blueflood-kafka sounds like something we might look into in the future. Should we keep it?

usnavi:
We'll still have the code in github, but having projects which aren't used and maintained is confusing for people looking at the codebase.

usnavi:
@shintasmith @VinnyQ @ChandraAddala  - may i merge?

ChandraAddala:
Go for it.

PR #653, opened 2016-03-30
refactor ShardStateIO to incorporate Datastax driver

shintasmith:
**Why**:
The main goal of this PR is to incorporate Datastax driver for the ShardStateIO. 

**What**:
Detailed changes:

* removed references to Astyanax ColumnFamily class from generic utility Instrumentation.java. This involves changing all methods in that class to take String column family name, instead of the Column Family class. All of the names of our column family have been added to CassandraModel.java

* upgraded 3rd party libraries to support Datastax driver:
  * com.codahale.metrics is now io.dropwizard.metrics
  * Guava to latest version 19.0
  * powermock and mockito upgraded to latest version
* changed maven-cassandra-plugin to also create Native Transport port 9042 (the binary native port needed by Datastax driver)
* moved all of the io.Astyanax* classes to io.astyanax.Astyanax* 
* made AstyanaxIO a singleton
* created DatastaxIO singleton
* created IOFactories class that knows how to switch between Astyanax and Datastax driver based on a new configuration parameter


shintasmith:
@usnavi , @ChandraAddala , @VinnyQ , @izrik , please review.

Note that I put ```@Ignore``` in IOFactoriesTest.java because I am still having issues running it under maven-surefire. The tests ran fine under IntelliJ. I will work on fixing that. 

But you guys can start reviewing the changes.

blueflood-core/pom.xml
```diff
@@ -214,33 +227,43 @@
...
+      <exclusions>
+        <exclusion>
+          <groupId>com.yammer.metrics</groupId>
+          <artifactId>metrics-core</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>com.codahale.metrics</groupId>
+          <artifactId>metrics-core</artifactId>
+        </exclusion>
+      </exclusions>
```
usnavi: I guess we should have a story to move to the latest version io.dropwizard.metrics, to clean this up one of these days?

shintasmith: I am changing our component to use the latest io.dropwizard.metrics, in this PR. That snippet is an exclusion from our 3rd party dependencies, who might have been compiled with older version of metrics. I don't want to bring older version of metrics into our build.

shintasmith:
@usnavi , @ChandraAddala , @VinnyQ , @izrik , I fixed the PowerMockito + Surefire issue. I removed the ```@Ignore``` from the IOFactories.java.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/IOFactories.java
```diff
@@ -0,0 +1,49 @@
...
+
+    public final static String ASTYANAX_DRIVER = "astyanax";
+
+    public final static String DATASTAX_DRIVER = "datastax";
+
+    public final static IOFactories singleton() {
+        return INSTANCE;
+    }
+
+    public ShardStateIO getShardStateIO() {
```
usnavi: I assume we'll have a method for every class that requires a Astyanax/Datastax implementation?

shintasmith: Yes, you are correct.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/IOConfig.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * Created by shin4590 on 3/24/16.
+ */
+public class IOConfig {
+
+    protected static final Configuration config = Configuration.getInstance();
```
usnavi: Should this be non-static?

shintasmith: Why should it be instance variable? :-)

usnavi: Its only used within the instance methods, right?  I'm sure @ChandraAddala will have an opinion.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/astyanax/AstyanaxShardStateIO.java
```diff
@@ -0,0 +1,82 @@
...
+import com.rackspacecloud.blueflood.service.UpdateStamp;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.util.Collection;
+import java.util.LinkedList;
+import java.util.Map;
+
+public class AstyanaxShardStateIO implements ShardStateIO {
```
ChandraAddala: Should this be a singleton?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxIO.java
```diff
@@ -0,0 +1,147 @@
...
+        monitorConnection();
+    }
+
+    public static DatastaxIO singleton() { return INSTANCE; }
+
+    private DatastaxIO() {
+    }
+
+    private static void connect() {
+        Set<InetSocketAddress> dbHosts = ioconfig.getUniqueBinaryTransportHostsAsInetSocketAddresses();
```
usnavi: Now that's a method name! :-P

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxIO.java
```diff
@@ -0,0 +1,147 @@
...
+        socketOptions.setConnectTimeoutMillis(ioconfig.getRequestTimeout())
+                     .setReadTimeoutMillis(ioconfig.getRequestTimeout());
+        return socketOptions;
+    }
+
+    private static PoolingOptions getPoolingOptions(int numHosts){
+
+        final PoolingOptions poolingOptions = new PoolingOptions();
+        poolingOptions
+                .setCoreConnectionsPerHost(HostDistance.LOCAL, 15)
```
usnavi: Should we put these defaults into Configuration and allow them to be configured, if need be?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/IOConfig.java
```diff
@@ -0,0 +1,62 @@
...
+import com.rackspacecloud.blueflood.service.CoreConfig;
+
+import java.net.InetSocketAddress;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * Created by shin4590 on 3/24/16.
+ */
```
usnavi: maybe a class description?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxIO.java
```diff
@@ -0,0 +1,147 @@
...
+import java.util.Collection;
+import java.util.Set;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+
+
+/**
+ * Created by shin4590 on 3/10/16.
+ */
```
usnavi: class description?

blueflood-core/src/test/java/com/rackspacecloud/blueflood/io/IOFactoriesTest.java
```diff
@@ -0,0 +1,119 @@
...
+    private static Configuration mockConfiguration;
+
+    @BeforeClass
+    public static void setup() throws Exception {
+        // this is how you mock singleton whose instance is declared
+        // as private static final, as the case for Configuration.class
+        PowerMockito.suppress(PowerMockito.constructor(Configuration.class));
+        mockConfiguration = PowerMockito.mock(Configuration.class);
+        PowerMockito.mockStatic(Configuration.class);
+        when(Configuration.getInstance()).thenReturn(mockConfiguration);
```
usnavi: Don't let this POWER go to your head

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxIO.java
```diff
@@ -0,0 +1,147 @@
...
+
+//    private final static Meter hostMeter = utils.Metrics.meter(DatastaxIO.class, "Hosts Connected");
+//    private final static Meter openConnectionsMeter = utils.Metrics.meter(DatastaxIO.class, "Total Open Connections");
+//    private final static Meter inFlightQueriesMeter = utils.Metrics.meter(DatastaxIO.class, "Total InFlight Queries");
+//    private final static Meter trashedConnectionsMeter = utils.Metrics.meter(DatastaxIO.class, "Total Trashed Connections");
+//    private final static Meter maxLoadMeter = utils.Metrics.meter(DatastaxIO.class, "Maximum Load");
+
+    static {
+        connect();
+        monitorConnection();
```
ChandraAddala: I feel like there is too much `static` going on in this class. May be we can structure it differently? Here is what I think are the issues with it. We can discuss tomorrow in detail but this is where I am coming from.

When we try to connect to cassandra during static initialization, if for some reason, there is an exception, the class wont get loaded at all which could lead to NoClassDefFound exception when we are trying to use it?. Thats my theory from my past experiences. I am not 100% sure of this.  

Also in this class, I like some separation from ioconfig. Too much going on during class loading which provides less control to us.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/IOConfig.java
ChandraAddala: Since IOConfig is also a singleton, dont think it matters much whether we leave it as static or non-static. I will say this though. I like less static. Spring has spoiled me.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/IOFactories.java
```diff
@@ -0,0 +1,49 @@
...
+        String driver = configuration.getStringProperty(CoreConfig.CASSANDRA_DRIVER);
+        if (Strings.isNullOrEmpty(driver) || driver.equalsIgnoreCase(ASTYANAX_DRIVER)) {
+            return new AstyanaxShardStateIO();
+        } else if ( driver.equalsIgnoreCase(DATASTAX_DRIVER) ) {
+            return new DatastaxShardStateIO();
+        } else {
+            throw new IllegalArgumentException(
+                            String.format("Invalid Cassandra driver: %s, must be one of %s|%s",
+                                          driver, ASTYANAX_DRIVER, DATASTAX_DRIVER));
+        }
```
ChandraAddala: Instead of throwing the exception here, cant we validate this when we are loading the `configuration` object itself to make sure the property CASSANDRA_DRIVER has valid value? 

Do we even need to validate our property file generated by blueflood-chef?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/IOFactories.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    // prevent ppl from instantiating directly
+    private IOFactories() {
+    }
```
ChandraAddala: During instantiation of this factory itself, can we determine what kinda driver we are going to use so that rest of the methods will have simple 'if' condition? If you want you can throw the exception here and the application wont even load I believe if they configure a wrong value.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxIO.java
shintasmith: I agree. I don't like the static-ness either. I was trying to keep the behavior of how these drivers/connection pools the same. Astyanax connection pools are initialized in similar manner. I first thought about making these connect() & monitorConnection() instance methods but then I think that I don't know where they should be called from -- would all the Service classes call them? Right now, it sort of happens when the AstyanaxIO class is loaded. Let's discuss.

I am spoiled from Spring. I never had to think about this when I used Spring.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxIO.java
shintasmith: :-) Maybe I should name it ioconfig.getUBTHostsISA() :-P

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxIO.java
shintasmith: Oh, I don't get what you mean by separation from ioconfig. Can you elaborate?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/Instrumentation.java
```diff
@@ -70,19 +69,19 @@
 
     private Instrumentation() {/* Used for JMX exposure */}
 
-    public static Timer.Context getReadTimerContext(ColumnFamily queryCF) {
+    public static Timer.Context getReadTimerContext(String queryCF) {
         return readTimers.getTimerContext(queryCF, false);
```
ChandraAddala: Datastax does not have equivalent of ColumnFamily?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/astyanax/AstyanaxShardStateIO.java
shintasmith: I don't see a big reason to make it a singleton. It is slightly harder to mock a singleton class. But I know how to do it now (I did it for the other singleton classes like AstyanaxIO, etc). You think it's better as singleton?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/IOConfig.java
shintasmith: I will make it non-static. :-)

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxIO.java
shintasmith: I can, yes, let me do that. The Remote thing is not quite relevant to us yet. That's for when we are cross DC. I will try to remove the .setRemote() from code.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/io/IOFactoriesTest.java
shintasmith: I LOVE POWA!!!!

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/IOFactories.java
shintasmith: What if someone tampers with the conf file on the environment? 

I am torn whether we should put the logic to check valid configuration values in the Configuration class, or in the classes that know better what the values mean.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/Instrumentation.java
shintasmith: Nope. The API exposes string as the Column Family name. It does have an internal representation of Column Family (which is called Table). The API is very close to JDBC.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/IOFactories.java
ChandraAddala: I say no validation :) I mean, do we validate currently for each of the configs? If not, why do it now?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/IOFactories.java
shintasmith: I can do that.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/IOFactories.java
shintasmith: With or without validation, I need to define the behavior of what driver should be "active", if it's neither "astyanax" or "datastax", yes? Do I just want to assume anything that's not "astyanax" is datastax, or vice versa?

usnavi: I'm for validating, but agree with @shintasmith torn feeling.  Do we want to leave this here, with a comment, and create a story to revisit how we want to validate the Configuration object values later?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxIO.java
ChandraAddala: I thought connect() method would be much readable if it has a signature like this. Basically if I have a set of db hosts, I should be able to use this class to make a connection without having to use ioconfig to instantiate it. 

`connect(Set<InetSocketAddress> dbHosts)`

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/IOFactories.java
ChandraAddala: It is kinda unusual to have validation for property files. You can write test cases in blueflood-chef to make sure we always configure valid property files. Keep a practice to not update property files manually. Always submit with a PR for blueflood-chef.

As per your question @shintasmith , it should use "astyanax" as default unless they specify "datastax" in the property file. That how I would keep the selection process.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/astyanax/AstyanaxShardStateIO.java
shintasmith: On second thought, I think you are right, it's better to be singleton. I will make it so. Thanks!

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxIO.java
ChandraAddala: I am thinking this class could be a singleton and it initializes lazily may be. After the application starts up, the first one to get the instance initializes the connection. We can force this initialization during startup.

shintasmith:
@usnavi @ChandraAddala @VinnyQ @izrik , I have updated the code per PR comments:
* added class/method descriptions, comments, copyright, etc
* instantiated the right ShardStateIO object in the IOFactories.getShardStateIO(). The code looks crazy ugly but I need to be able to accommodate testing the switch of the different drivers :-(
* kept the DatastaxIO.connect() as is after discussion with @ChandraAddala and @usnavi

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/IOFactories.java
```diff
@@ -0,0 +1,76 @@
...
+        synchronized (IOFactories.class) {
+            if ( shouldUseDatastax() ) {
+                if (shardStateIO == null || !(shardStateIO instanceof DatastaxShardStateIO)) {
+                    shardStateIO = new DatastaxShardStateIO();
+                }
+            } else {
+                if ( shardStateIO == null || ! (shardStateIO instanceof AstyanaxShardStateIO) ) {
+                    shardStateIO = new AstyanaxShardStateIO();
+                }
+            }
```
ChandraAddala: Oh no...What did you do?!!!

shintasmith: 😝 You like that, @ChandraAddala ? 
Seriously, I am open to suggestions. I am trying to incorporate @usnavi 's suggestion of having ShardStateIO instance variable in IOFactories. Again, I have the same issue as creating a boolean variable. Once it is set, there is no way to unset it for a different configuration value.

usnavi: Since this code will be short-lived (eventually the astyanax code will be removed when we move to Cassandra 3.0), I don't want to spend too much of my time thinking on this.  In retrospect, I wish we hadn't said anything :-(.

shintasmith:
@ChandraAddala @usnavi @VinnyQ @izrik , I figured out how to do it in not-so-sucky way. Please take a look at IOContainer (used to be the IOFactories). I am basically making it non-singleton. I can test the class with 94% coverage. Let me know what you think.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/IOContainer.java
```diff
@@ -0,0 +1,106 @@
...
+        public static DriverType getDriverType(String driver) {
+            // if driver is explicitly datastax, then return DATASTAX
+            // anything else, return ASTYANAX
+            if ( ! Strings.isNullOrEmpty(driver) && DriverType.valueOf(driver.toUpperCase()) == DATASTAX ) {
+                return DATASTAX;
+            }
+            return ASTYANAX;
+        }
+    }
+}
```
ChandraAddala: This looks good.

ChandraAddala:
Looks good.

shintasmith:
The test that's failing is the MPlot ROR test. @usnavi, wink wink..

PR #654, opened 2016-03-31
Fix Tracker.trackDelayedMetricsTenant

VinnyQ:
# What
Modify tracker so that the trackDelayedMetricsTenant, when enabled, also log the delayed metric names and collection time.

# How
Modify Tracker.trackDelayedMetricsTenant to also log the name and collectionTime of the delayed metrics that triggered the tracking.

# Why
For easier investigation of delayed metrics when we're receiving them.

# How to test
Enable the delayed metrics tracking using 
```
blueflood-chef/contrib/blueflood_tenant_tracker.sh myusername track_delayed_start node.hostname
```
Ingest metrics with collectionTime < ingestionTime by DELAYED_METRICS_MILLIS interval, verify that the delayed metrics names and collectionTimes are logged after line prefixed with "[TRACKER][DELAYED METRIC] Tenant sending delayed metrics <tenantId>".

### DIFF
--- a/blueflood-http/pom.xml
+++ b/blueflood-http/pom.xml
@@ -72,6 +72,7 @@
         <artifactId>maven-surefire-plugin</artifactId>
         <version>2.14.1</version>
         <configuration>
+          <forkMode>always</forkMode>
           <!-- Skips unit tests if the value of skip.unit.tests property is true -->
           <skipTests>${skip.unit.tests}</skipTests>
           <!-- Excludes integration tests when unit tests are run. -->
@@ -96,6 +97,7 @@
               <goal>integration-test</goal>
             </goals>
             <configuration>
+              <forkMode>always</forkMode>
               <systemPropertyVariables>
                 <CASSANDRA_HOSTS>${cassandra.listenAddress}:${cassandra.rpcPort}</CASSANDRA_HOSTS>
                 <DEFAULT_CASSANDRA_PORT>${cassandra.rpcPort}</DEFAULT_CASSANDRA_PORT>
@@ -177,11 +179,11 @@
       <version>4.3.3</version>
     </dependency>
 
-      <dependency>
-          <groupId>org.apache.commons</groupId>
-          <artifactId>commons-io</artifactId>
-          <version>1.3.2</version>
-      </dependency>
+    <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-io</artifactId>
+      <version>1.3.2</version>
+    </dependency>
 
       <!-- testing dependencies -->
     <dependency>
@@ -195,9 +197,21 @@
     <dependency>
       <groupId>org.mockito</groupId>
       <artifactId>mockito-all</artifactId>
-      <version>1.8.4</version>
+      <version>1.10.19</version>
       <scope>test</scope>
       <optional>true</optional>
+    </dependency>
+
+    <dependency>
+      <groupId>org.powermock</groupId>
+      <artifactId>powermock-module-junit4</artifactId>
+      <version>1.6.2</version>
+    </dependency>
+
+    <dependency>
+      <groupId>org.powermock</groupId>
+      <artifactId>powermock-api-mockito</artifactId>
+      <version>1.6.2</version>
     </dependency>
 
     <dependency>
--- a/blueflood-http/pom.xml
+++ b/blueflood-http/pom.xml
@@ -97,7 +97,6 @@
               <goal>integration-test</goal>
             </goals>
             <configuration>
-              <forkMode>always</forkMode>
               <systemPropertyVariables>
                 <CASSANDRA_HOSTS>${cassandra.listenAddress}:${cassandra.rpcPort}</CASSANDRA_HOSTS>
                 <DEFAULT_CASSANDRA_PORT>${cassandra.rpcPort}</DEFAULT_CASSANDRA_PORT>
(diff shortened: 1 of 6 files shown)


VinnyQ:
@izrik @shintasmith @usnavi @ChandraAddala please review, thanks!

blueflood-http/pom.xml
```diff
@@ -72,6 +72,7 @@
         <artifactId>maven-surefire-plugin</artifactId>
         <version>2.14.1</version>
         <configuration>
+          <forkMode>always</forkMode>
```
usnavi: what's this for?

blueflood-http/src/main/java/com/rackspacecloud/blueflood/tracker/Tracker.java
```diff
@@ -114,6 +111,24 @@ public void removeAllMetricNames() {
         log.info("[TRACKER] All metric names removed.");
     }
 
+    public Set<String> getMetricNames() {
+        return this.metricNames;
```
usnavi: I guess I'm out voted on using ```this.``` :-(

shintasmith: This is not the same issue, because there is no parameter with the same name in the method. We should just remove the "this." part.

usnavi: I'm all for that!  

PS - It's the same issue in my book :-P

blueflood-http/src/main/java/com/rackspacecloud/blueflood/tracker/Tracker.java
```diff
@@ -221,6 +215,16 @@ public void trackResponse(HttpRequest request, HttpResponse response) {
 
     }
 
+    public void trackDelayedMetricsTenant(String tenantid, List<Metric> delayedMetrics) {
+        if (isTrackingDelayedMetrics) {
+            String logMessage = String.format("[TRACKER][DELAYED METRIC] Tenant sending delayed metrics %s\n",tenantid);
+            for (Metric metric : delayedMetrics) {
+                logMessage += String.format("%s has collectionTime %d\n", metric.getLocator().toString(), metric.getCollectionTime());
```
ChandraAddala: Do you think its worth printing the time in human readable format? Also the amount of delay.

shintasmith:
Dont forget to update blueflood_tenant_tracker.sh script in blueflood-chef :-)

blueflood-http/src/main/java/com/rackspacecloud/blueflood/tracker/Tracker.java
ChandraAddala: I was going to comment the same. We dont need "this" here.

blueflood-http/pom.xml
VinnyQ: The static mocking using PowerMock doesn't seem to work without it, and anyway it makes each tests cleaner/more separated because every test suite is running in its own space.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/tracker/Tracker.java
VinnyQ: I can convert collectionTime to human readable.

For the amount of delay, that'll takes extra logic because 1) we have to pass in the ingestion time (or add a new timestamp property to each metric, which I was considering), and 2) we have to pull the DELAYED_METRICS_MILLIS from property and calculate the difference for every delayed metrics.  I wasn't sure if I should add that kind of complexity and extra calculation just for this.

VinnyQ:
@shintasmith actually this change doesn't involve making any changes in that blueflood-chef blueflood_tenant_track script.  I was considering adding a flag to turn the extra logging on/off for delayed metrics tracking, but I figure we'll probably need that details all the time anyhow, so we might as well just try logging everything without needing to have extra flags.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/tracker/Tracker.java
VinnyQ: I don't have strong feeling about this either way.  I'll get rid of it as your preference.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/tracker/Tracker.java
VinnyQ: Converted to date and log in yyyy-MM-dd HH:mm:ss format.  Decided that it is too much changes and extra calculation to print the ingest time and interval of delay, which you can easily do by looking at the log timestamp and do the calculation outside the log.

VinnyQ:
@shintasmith @usnavi @ChandraAddala please review again and +1 if we're good.

Thanks!

blueflood-http/src/main/java/com/rackspacecloud/blueflood/tracker/Tracker.java
VinnyQ: Ok thinking it through, I realize it's not hard to just use the current system time and calculate the difference between that and collectionTime and log that also, since that's what we'll be doing anyway using the log timestamp.  I hope this is good enough for now.  

The more *correct* way though is to add a new "creationTime" for each metric (add a new property in the Metric class and timestamp it in the constructor), then subtract that from collectionTime to calculate the actual delays.  But doing so will require us changing all the serialization code, which is too intrusive just for this change.  (If we do have a reason to track the ingestion/class creation time in the fgure, however, then the code can be modified to use that).

usnavi:
lgtm

shintasmith:
lgtm, for now, although we should look into why Powermock doesn't work without forkMode=always. I didn't have to set that in blueflood-core when I used Powermock there.

VinnyQ:
Thanks. 

@shintasmith it doesn't work specifically for static mocking the Logger class via maven surefire-plugin.  It works fine without it when running via IntelliJ.  Without it I get a "null pointer error" for when the tests try to use the mocked logger class.

PR #655, opened 2016-04-01
Removing value checks to ROR tests until we can investigate with CMD-1001

usnavi:


ChandraAddala:
slacker!

shintasmith:
@usnavi another test is failing

usnavi:
@shintasmith - hmmm, this is more concerning.  Even though this is called "MultiRollupsQuery" this isn't verifying the generated values like the tests I added.  This is something that has been around for a while.

usnavi:
@VinnyQ @izrik @ChandraAddala @shintasmith - welp, this ran 3 times on travis without failing.  (Boy that's confident!)

ChandraAddala:
Looks good.

PR #656, opened 2016-04-01
Test rollup event emitter

izrik:
`Emitter<>` serves as a kind of publish/subscribe event notification system. Listeners can be registered to listen for an event of particular name. The `RollupEventEmitter` is a subclass of `EventEmitter<RollupEvent>` that reports events related to rollups. However, the subclass is broken, IMHO.

The main thing is that all events will be silently ignored if there isn't a listener registered for events named `ROLLUP_EVENT_NAME`, which has a value of "rollup". To be clear: even events that have a different name, events that are for some other name, will not be delivered at all, unless a listener is registered for `ROLLUP_EVENT_NAME`.

Additionally, the purpose of the class seems to be to automatically set the unit of any incoming `RollupEvent` and then just pass it on as usual. I honestly don't think that should warrant a separate thread pool. The thread pool business is making it difficult to test.

The specific functionality that `RollupEventEmitter` adds is to check all events to be emitted and make sure that they have the right unit set. However, it is hard-coded to get the unit from the `com.rackspacecloud.blueflood.io.ElasticIO` class, which is in the `blueflood-elasticsearch` module. As such, it is currently infeasible to write unit tests for this functionality. They would have to be integration tests.

This PR adds tests for the `RollupEventEmitter`, `Emitter<T>`, and `RollupEvent` classes. This brings line coverage up to about 65%, 97%, and 100%, respectively. I don't think I can get `RollupEventEmitter` any higher, given the above limitations.

### DIFF
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/eventemitter/RollupEventEmitter.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/eventemitter/RollupEventEmitter.java
@@ -16,6 +16,7 @@
 
 package com.rackspacecloud.blueflood.eventemitter;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Function;
 import com.google.common.collect.Lists;
 import com.rackspacecloud.blueflood.concurrent.ThreadPoolBuilder;
@@ -73,6 +74,7 @@ public class RollupEventEmitter extends Emitter<RollupEvent> {
                 .withUnboundedQueue()
                 .build());
     }
+    @VisibleForTesting
     public RollupEventEmitter(ExecutorService executor) {
         eventExecutors = executor;
     }
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/eventemitter/RollupEventEmitter.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/eventemitter/RollupEventEmitter.java
@@ -37,7 +37,7 @@ public class RollupEventEmitter extends Emitter<RollupEvent> {
     private static ExecutorService eventExecutors;
     private static final RollupEventEmitter instance = new RollupEventEmitter();
 
-    private RollupEventEmitter() {
+    public RollupEventEmitter() {
         eventExecutors = new ThreadPoolBuilder()
                 .withName("RollupEventEmitter ThreadPool")
                 .withCorePoolSize(numberOfWorkers)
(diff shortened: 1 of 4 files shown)


izrik:
Don't merge yet.
@ChandraAddala @usnavi @shintasmith @VinnyQ What do you guys think of this class? I'd like to just fix it, even though that's not technically part of the story. Fixing it would make it much easier to test.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/eventemitter/RollupEventEmitter.java
```diff
@@ -30,20 +30,51 @@
...
+ * {@link com.rackspacecloud.blueflood.types.HistogramRollup}
+ *
+ * @todo Merge {@code RollupEventEmitter} and {@link Emitter<>} into a single
+ * class. {@link Emitter<>} doesn't appear to be used anywhere else.
+ *
+ * @todo Change the {@code eventPayload}/{@code args} parameter of
+ * {@link RollupEventEmitter#emit} from an array to a single item. There don't
+ * appear to be any parts of the codebase that use more than a single event
+ * argument.
+ */
```
usnavi: @izrik - would you also make a story to document this work.  We might forget about it if its not in our backlog.

izrik: Done.

izrik:
@usnavi @ChandraAddala @shintasmith @VinnyQ OK, I think that's the most coverage I can feasibly get significantly without modifying the class. I've created some new stories to that effect. Does anyone have any other suggestions?

blueflood-core/src/test/java/com/rackspacecloud/blueflood/eventemitter/EmitterTest.java
```diff
@@ -0,0 +1,310 @@
...
+        });
+        Future<Object> f2 = executors.submit(new Callable<Object>() {
+            @Override
+            public Object call() throws Exception {
+                startLatch.await();
+                emitter.emit(testEventName, obj2);
+                return null;
+            }
+        });
+        Thread.sleep(1000);
```
usnavi: why sleep here?

izrik: Because the person who wrote that particular test either doesn't know how to test multithreaded code, or were constrained to test it the wrong way due to bad architecture. It looks like he's giving the threads enough time to reach the `startLatch.await()` calls.

usnavi: Cool, can you put a comment here outlining that?

usnavi:
lgtm - please create task to track this in our storyboard.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/eventemitter/EmitterTest.java
izrik: Done.

izrik:
@usnavi Already done. I added a comment to the jira ticket with links to new stories.

usnavi:
@izrik - I mean, add a task for this PR - or did I miss it in the storyboard?

izrik:
@usnavi This is part of CMD-945. Or did you mean sub-task?

PR #657, opened 2016-04-05
Delayed metrics separation

ChandraAddala:
This is the first part of the improvements proposed to handle delayed metrics. What it provides is a separation of rollups of delayed metrics from the on-time metrics. Now we have two config parameters 
ROLLUP_DELAY_MILLIS to control the rollup of on-time metrics and DELAYED_METRICS_ROLLUP_DELAY_MILLIS to control the rollup of delayed metrics.

If we set both these config parameters to same value, then we process rollup of delayed metrics same as on-time ones. More details about the design can be found here.

https://one.rackspace.com/pages/viewpage.action?pageId=162810603#BluefloodShard/SlotManagement-Option1B:Savinglastrolluptime

Followup stories:
* Change blueflood-chef to include the new config parameter
* Analyze the impact of this change on graphs ( queued rollups, scheduled slot check)

### DIFF
--- a/blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/service/ShardStateIntegrationTest.java
+++ b/blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/service/ShardStateIntegrationTest.java
@@ -577,7 +577,9 @@ public class ShardStateIntegrationTest extends IntegrationTestBase {
                 for (Map.Entry<Granularity, Map<Integer, UpdateStamp>> e0 : updates.entrySet()) {
                     for (Map.Entry<Integer, UpdateStamp> e1 : e0.getValue().entrySet()) {
                         SlotState state = new SlotState(e0.getKey(), e1.getKey(), e1.getValue().getState());
-                        state.withTimestamp(e1.getValue().getTimestamp());
+                        state.withTimestamp(e1.getValue().getTimestamp())
+                             .withLastUpdatedTimestamp(state.getLastUpdatedTimestamp());
+
                         states.add(state);
                     }
                 }
--- a/blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/service/ShardStateIntegrationTest.java
+++ b/blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/service/ShardStateIntegrationTest.java
@@ -22,6 +22,7 @@ import com.rackspacecloud.blueflood.io.ShardStateIO;
 import com.rackspacecloud.blueflood.io.datastax.DatastaxShardStateIO;
 import com.rackspacecloud.blueflood.rollup.Granularity;
 import com.rackspacecloud.blueflood.rollup.SlotKey;
+import com.rackspacecloud.blueflood.utils.ClockImpl;
 import com.rackspacecloud.blueflood.utils.Util;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
@@ -43,6 +44,9 @@ public class ShardStateIntegrationTest extends IntegrationTestBase {
     
     private ShardStateIO io;
     private List<Integer> shards;
+
+    private final int DELAYED_METRICS_ROLLUP_DELAY_MILLIS = 300000;
+    private final int ROLLUP_DELAY_MILLIS = 300000;
     
     public ShardStateIntegrationTest(ShardStateIO io, List<Integer> shardsToTest) {
         this.io = io;
@@ -52,7 +56,7 @@ public class ShardStateIntegrationTest extends IntegrationTestBase {
     @Test
     public void testSingleShardManager() {
         long time = 1234000L;
-        ScheduleContext ctx = new ScheduleContext(time, shards);
+        ScheduleContext ctx = new ScheduleContext(time, shards, new ClockImpl());
         ShardStateWorker pull = new ShardStatePuller(shards, ctx.getShardStateManager(), this.io);
         ShardStateWorker push = new ShardStatePusher(shards, ctx.getShardStateManager(), this.io);
         
@@ -89,8 +93,8 @@ public class ShardStateIntegrationTest extends IntegrationTestBase {
     public void testRollupFailureForDelayedMetrics() {
         long time = 1234000L;
         Collection<Integer> managedShards = Lists.newArrayList(0);
-        ScheduleContext ingestionCtx = new ScheduleContext(time, managedShards);
-        ScheduleContext rollupCtx = new ScheduleContext(time, managedShards);
+        ScheduleContext ingestionCtx = new ScheduleContext(time, managedShards, new ClockImpl());
+        ScheduleContext rollupCtx = new ScheduleContext(time, managedShards, new ClockImpl());
         // Shard workers for rollup ctx
         ShardStateWorker rollupPuller = new ShardStatePuller(managedShards, rollupCtx.getShardStateManager(), this.io);
         ShardStateWorker rollupPusher = new ShardStatePusher(managedShards, rollupCtx.getShardStateManager(), this.io);
@@ -104,7 +108,7 @@ public class ShardStateIntegrationTest extends IntegrationTestBase {
 
         rollupPuller.performOperation(); // Shard state is read on rollup host
         rollupCtx.setCurrentTimeMillis(time + 600000);
-        rollupCtx.scheduleEligibleSlots(300000, 7200000);
+        rollupCtx.scheduleEligibleSlots(ROLLUP_DELAY_MILLIS, DELAYED_METRICS_ROLLUP_DELAY_MILLIS);
         Assert.assertEquals(1, rollupCtx.getScheduledCount());
 
         // Simulate the hierarchical scheduling of slots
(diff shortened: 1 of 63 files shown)


blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/CoreConfig.java
```diff
@@ -150,6 +150,7 @@
     
     // how long we typically wait to schedule a rollup.
     ROLLUP_DELAY_MILLIS("300000"),
+    DELAYED_METRICS_ROLLUP_DELAY_MILLIS("300000"),
```
usnavi: how about ```DELAYED_METRICS_ROLLUP_DELAY_MILLIS_DELAY_DELAY```?  :-P

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/ScheduleContext.java
```diff
@@ -198,9 +201,10 @@ public void update(long millis, int shard) {
      * {@code maxAgeMillis}, then that slot will scheduled.
      *
      * @param maxAgeMillis
+     * @param delayedMetricsMaxAgeMillis
      */
     // only one thread should be calling in this puppy.
-    void scheduleSlotsOlderThan(long maxAgeMillis) {
+    void scheduleEligibleSlots(long maxAgeMillis, long delayedMetricsMaxAgeMillis) {
```
usnavi: can you update the comment to reference the new parameter in the desc?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/UpdateStamp.java
```diff
@@ -35,21 +35,32 @@
...
     private boolean dirty;
-    
+
+    private long lastRollupTimestamp;
+
     public UpdateStamp(long timestamp, State state, boolean dirty) {
         setTimestamp(timestamp);
         setState(state);
         setDirty(dirty);
     }
```
usnavi: can this constructor be removed?

usnavi:
other than the minor comments, lgtm.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/UpdateStamp.java
ChandraAddala: It is still being used. This is the case of a finding a slot for the first time during ingest which is not present in metrics_state table yet. This will probably happen only during dev.

https://github.com/rackerlabs/blueflood/blob/f24a8ba3990dd38e26e29da1d3ec234285c478f1/blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/ShardStateManager.java#L270

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/CoreConfig.java
shintasmith: But why do we want to set the default for DELAYED_METRICS_ROLLUP_DELAY_MILLIS be the same as normal?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxShardStateIO.java
```diff
@@ -53,19 +53,24 @@
         try {
             Select.Where statement = QueryBuilder
                     .select()
-                    .all()
+                    .column("key")
+                    .column("column1")
+                    .column("value")
+                    .writeTime("value")
```
shintasmith: wow, nice that it's so easy! Either @usnavi or you will get merge conflict on this file. :-)

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/ScheduleContext.java
```diff
@@ -141,10 +141,13 @@
     /** shard lock manager */
     private final ShardLockManager lockManager;
 
-    public ScheduleContext(long currentTimeMillis, Collection<Integer> managedShards) {
+    private final Clock clock;
+
+    public ScheduleContext(long currentTimeMillis, Collection<Integer> managedShards, Clock clock) {
```
shintasmith: Maybe you should provide a convenience constructor that takes only 2 parameters, and the Clock is defaulted to DefaultClockImpl ? I think that would be the majority of the calls. Only in tests that you would provide a different Clock implementation.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/ScheduleContextClearFromRunningTest.java
```diff
@@ -35,10 +39,15 @@ public void setUp() throws GranularityException {
         coarserGran = gran.coarser();
         coarserSlot = coarserGran.slot(updateTime);
 
-        ctx = new ScheduleContext(now, shards);
+        ctx = new ScheduleContext(now, shards, new Clock() {
+            @Override
+            public Instant now() {
+                return new Instant(lastRollupTime);
```
shintasmith: You do an override here, but mock in later test class. Curious, why two different ways?

blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/ShardStateManagerTest.java
```diff
@@ -0,0 +1,225 @@
...
+        //new incoming slot state
+        slotStateManager.updateSlotOnRead(newUpdateForActiveSlot);
+
+        Map<Integer, UpdateStamp> slotStamps = slotStateManager.getSlotStamps();
+        UpdateStamp updateStamp = slotStamps.get(TEST_SLOT);
+
+        assertEquals("Unexpected state",  UpdateStamp.State.Active, updateStamp.getState());
+        assertEquals("Last ingestion timestamp is incorrect", existingLastIngestionTime, updateStamp.getTimestamp());
+        assertEquals("Last rollup timestamp is incorrect", newRolledSlotLastUpdatedTime, updateStamp.getLastRollupTimestamp());
+    }
```
shintasmith: :+1: nice set of tests!

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/CoreConfig.java
ChandraAddala: If you set it same as normal, there is no behavior change. Also I dont know what default value we want to set this to? May be 20 mins?

blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/ScheduleContextClearFromRunningTest.java
ChandraAddala: Well in there I knew the value before hand. But in other test cases, I had to set value later based on the scenario I was testing.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/ScheduleContext.java
ChandraAddala: @shintasmith Done!

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/ScheduleContext.java
ChandraAddala: Done!

ChandraAddala:
@shintasmith @usnavi @izrik Shall I merge this if you are done with the review?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/UpdateStamp.java
izrik: Would it be feasible to make one constructor call the other, so we don't duplicate anything?

blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/service/ShardStateIntegrationTest.java
```diff
@@ -414,21 +423,21 @@ public void testSlotStateConvergence() throws InterruptedException {
...
-        ScheduleContext ctxIngestor1 = new ScheduleContext(time, shards);
+        ScheduleContext ctxIngestor1 = new ScheduleContext(time, shards, new DefaultClockImpl());
         ShardStatePuller pullerIngestor1 = new ShardStatePuller(shards, ctxIngestor1.getShardStateManager(), this.io);
         ShardStatePusher pusherIngestor1 = new ShardStatePusher(shards, ctxIngestor1.getShardStateManager(), this.io);
         allWorkers.add(pullerIngestor1);
         allWorkers.add(pusherIngestor1);
 
         // Ingestor 2
-        ScheduleContext ctxIngestor2 = new ScheduleContext(time, shards);
+        ScheduleContext ctxIngestor2 = new ScheduleContext(time, shards, new DefaultClockImpl());
```
shintasmith: Dont you want to change all of these to use the constructor that takes 2 params?

usnavi:
lgtm

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/RollupService.java
```diff
@@ -135,6 +136,7 @@ public Integer getValue() {
         // higher.
         Configuration config = Configuration.getInstance();
         rollupDelayMillis = config.getLongProperty("ROLLUP_DELAY_MILLIS");
+        delayedMetricRollupDelayMillis = config.getLongProperty("DELAYED_METRICS_ROLLUP_DELAY_MILLIS");
```
izrik: I recommend using the `CoreConfig` enum here, instead of a string.

izrik:
@ChandraAddala I'm still looking through it. Did you want me to write any tests for it?

ChandraAddala:
@izrik Since you have a good understanding of the ScheduleContext/ShardStateManager, it would be great if you can verify that the code changes make sense. I am fine with the tests.

izrik:
@ChandraAddala I _think_ I fully understand it. If nothing else, I'll make sure when I write more tests for `SlotStateManager`.

PR #660, opened 2016-04-06
Adding Datastax support for Metadata and TestIO

usnavi:
- Adding Datastax support for MetadataIO
- Creating TestIO for Astyanax & Datastax as required by tests

### DIFF
--- a/blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/io/astyanax/AstyanaxWriterIntegrationTest.java
+++ b/blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/io/astyanax/AstyanaxWriterIntegrationTest.java
@@ -16,87 +16,82 @@
 
 package com.rackspacecloud.blueflood.io.astyanax;
 
+import com.rackspacecloud.blueflood.io.CassandraModel;
+import com.rackspacecloud.blueflood.io.CassandraUtilsIO;
 import com.rackspacecloud.blueflood.io.IntegrationTestBase;
-import com.rackspacecloud.blueflood.io.astyanax.AstyanaxWriter;
 import com.rackspacecloud.blueflood.types.Locator;
 import org.junit.Test;
 
+import static org.junit.Assert.*;
+
+
 public class AstyanaxWriterIntegrationTest extends IntegrationTestBase {
+
+    private CassandraUtilsIO cassandraUtilsIO = new AstyanaxCassandraUtilsIO();
 
     @Test
     public void testEnsureStringMetricsDoNotEndUpInNumericSpace() throws Exception {
-        assertNumberOfRows("metrics_string", 0);
-        assertNumberOfRows("metrics_full", 0);
-        assertNumberOfRows("metrics_locator", 0);
+
+        assertEquals( "Ensure " + CassandraModel.CF_METRICS_STRING_NAME + "is 0 before",
+                0, cassandraUtilsIO.getKeyCount( CassandraModel.CF_METRICS_STRING_NAME ) );
+        assertEquals( "Ensure " + CassandraModel.CF_METRICS_FULL_NAME + "is 0 before",
+                0, cassandraUtilsIO.getKeyCount( CassandraModel.CF_METRICS_FULL_NAME ) );
+        assertEquals( "Ensure " + CassandraModel.CF_METRICS_LOCATOR_NAME + "is 0 before",
+                0, cassandraUtilsIO.getKeyCount( CassandraModel.CF_METRICS_LOCATOR_NAME ) );
 
         writeMetric("string_metric", "This is a string test");
-        
-        assertNumberOfRows("metrics_string", 1);
-        assertNumberOfRows("metrics_full", 0);
-        assertNumberOfRows("metrics_locator", 0);
+
+        assertEquals( "Ensure " + CassandraModel.CF_METRICS_STRING_NAME + "is 1 after",
+                1, cassandraUtilsIO.getKeyCount( CassandraModel.CF_METRICS_STRING_NAME ) );
+        assertEquals( "Ensure " + CassandraModel.CF_METRICS_FULL + "is 0 after",
+                0, cassandraUtilsIO.getKeyCount( CassandraModel.CF_METRICS_FULL_NAME ) );
+        assertEquals( "Ensure " + CassandraModel.CF_METRICS_LOCATOR_NAME + "is 0 after",
+                0, cassandraUtilsIO.getKeyCount( CassandraModel.CF_METRICS_LOCATOR_NAME ) );
     }
 
     @Test
     public void testEnsureNumericMetricsDoNotEndUpInStringSpaces() throws Exception {
-        assertNumberOfRows("metrics_string", 0);
-        assertNumberOfRows("metrics_full", 0);
-        assertNumberOfRows("metrics_locator", 0);
+        assertEquals( "Ensure " + CassandraModel.CF_METRICS_STRING_NAME + "is 0 before",
+                0, cassandraUtilsIO.getKeyCount( CassandraModel.CF_METRICS_STRING_NAME ) );
+        assertEquals( "Ensure " + CassandraModel.CF_METRICS_FULL_NAME + "is 0 before",
+                0, cassandraUtilsIO.getKeyCount( CassandraModel.CF_METRICS_FULL_NAME ) );
+        assertEquals( "Ensure " + CassandraModel.CF_METRICS_LOCATOR_NAME + "is 0 before",
+                0, cassandraUtilsIO.getKeyCount( CassandraModel.CF_METRICS_LOCATOR_NAME ) );
 
         writeMetric("long_metric", 64L);
-        
-        assertNumberOfRows("metrics_string", 0);
-        assertNumberOfRows("metrics_full", 1);
-        assertNumberOfRows("metrics_locator", 1);
-    }
 
-    @Test
-    public void testMetadataGetsWritten() throws Exception {
-        assertNumberOfRows("metrics_metadata", 0);
+        assertEquals( "Ensure " + CassandraModel.CF_METRICS_STRING_NAME + "is 0 after",
+                0, cassandraUtilsIO.getKeyCount( CassandraModel.CF_METRICS_STRING_NAME ) );
+        assertEquals( "Ensure " + CassandraModel.CF_METRICS_FULL_NAME + "is 1 after",
+                1, cassandraUtilsIO.getKeyCount( CassandraModel.CF_METRICS_FULL_NAME ) );
+        assertEquals( "Ensure " + CassandraModel.CF_METRICS_LOCATOR_NAME + "is 1 after",
+                1, cassandraUtilsIO.getKeyCount( CassandraModel.CF_METRICS_LOCATOR_NAME ) );
 
-        Locator loc1 = Locator.createLocatorFromPathComponents("acONE", "entityId", "checkId", "mz", "metric");
-        Locator loc2 = Locator.createLocatorFromPathComponents("acTWO", "entityId", "checkId", "mz", "metric");
-        AstyanaxWriter writer = AstyanaxWriter.getInstance();
-
-        // multiple cols on a single locator should produce a single row.
-//        writer.writeMetadataValue(loc1, "a", new byte[]{1,2,3,4,5});
-//        writer.writeMetadataValue(loc1, "b", new byte[]{6,7,8,9,0});
-//        writer.writeMetadataValue(loc1, "c", new byte[]{11,22,33,44,55,66});
-//        writer.writeMetadataValue(loc1, "d", new byte[]{-1,-2,-3,-4});
-//        writer.writeMetadataValue(loc1, "e", new byte[]{1,2,3,4,5});
-//        writer.writeMetadataValue(loc1, "f", new byte[]{1,2,3,4,5,6,7,8,9,0});
-        writer.writeMetadataValue(loc1, "a", "Some1String");
-        writer.writeMetadataValue(loc1, "b", "Some2String");
-        writer.writeMetadataValue(loc1, "c", "Some3String");
-        writer.writeMetadataValue(loc1, "d", "Some4String");
-        writer.writeMetadataValue(loc1, "e", "Some5String");
-        writer.writeMetadataValue(loc1, "f", "Some6String");
-
-        assertNumberOfRows("metrics_metadata", 1);
-
-        // new locator means new row.
-        writer.writeMetadataValue(loc2, "a", "strrrrring");
-        assertNumberOfRows("metrics_metadata", 2);
     }
 
     @Test
     public void testExcessEnumMetricGetsWritten() throws Exception {
-        assertNumberOfRows("metrics_excess_enums", 0);
+        assertEquals( "Ensure " + CassandraModel.CF_METRICS_EXCESS_ENUMS_NAME + " 0 before",
+                0, cassandraUtilsIO.getKeyCount( CassandraModel.CF_METRICS_EXCESS_ENUMS_NAME ) );
 
         Locator loc1 = Locator.createLocatorFromPathComponents("acONE", "entityId", "checkId", "mz", "metric");
         Locator loc2 = Locator.createLocatorFromPathComponents("acTWO", "entityId", "checkId", "mz", "metric");
         AstyanaxWriter writer = AstyanaxWriter.getInstance();
 
         writer.writeExcessEnumMetric(loc1);
-        assertNumberOfRows("metrics_excess_enums", 1);
+        assertEquals(  "Ensure " + CassandraModel.CF_METRICS_ENUM_NAME + " 1 after",
+                1, cassandraUtilsIO.getKeyCount( CassandraModel.CF_METRICS_EXCESS_ENUMS_NAME ) );
 
         // new locator means new row.
         writer.writeExcessEnumMetric(loc2);
-        assertNumberOfRows("metrics_excess_enums", 2);
+        assertEquals(  "Ensure " + CassandraModel.CF_METRICS_ENUM_NAME + " 2 after",
+                2, cassandraUtilsIO.getKeyCount( CassandraModel.CF_METRICS_EXCESS_ENUMS_NAME ) );
     }
 
     @Test
     public void testExcessEnumMetricDoesNotDuplicates() throws Exception {
-        assertNumberOfRows("metrics_excess_enums", 0);
+        assertEquals( "Ensure " + CassandraModel.CF_METRICS_EXCESS_ENUMS_NAME + " 0 before",
+                0, cassandraUtilsIO.getKeyCount( CassandraModel.CF_METRICS_EXCESS_ENUMS_NAME ) );
 
         Locator loc1 = Locator.createLocatorFromPathComponents("ac", "entityId", "checkId", "mz", "metric");
         Locator loc2 = Locator.createLocatorFromPathComponents("ac", "entityId", "checkId", "mz", "metric");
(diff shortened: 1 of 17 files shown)


blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/IOContainer.java
```diff
@@ -34,6 +38,9 @@
     private static final Configuration configuration = Configuration.getInstance();
 
     private ShardStateIO shardStateIO;
+    private MetadataIO metadataIO;
+    private TestIO testIO;
```
shintasmith: I am not sure I like the TestIO class in the classes of main code. In fact, I was going to propose moving the TestIO class in the src/integration-test/java package.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxMetadataIO.java
```diff
@@ -0,0 +1,162 @@
...
+    public static final String VALUE = "value";
+
+    private static final Logger LOG = LoggerFactory.getLogger( DatastaxMetadataIO.class );
+
+    private final StringMetadataSerDes serDes = new StringMetadataSerDes();
+
+    @Override
+    public Map<String, String> getAllValues( Locator locator ) throws IOException {
+
+        Timer.Context ctx = Instrumentation.getReadTimerContext( CassandraModel.CF_METRICS_STATE_NAME );
```
shintasmith: Should this be CF_METRICS_METADATA_NAME ?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxMetadataIO.java
```diff
@@ -0,0 +1,162 @@
...
+        }
+        finally {
+            ctx.stop();
+        }
+    }
+
+    @Override
+    public Table<Locator, String, String> getAllValues( Set<Locator> locators ) throws IOException {
+
+        Timer.Context ctx = Instrumentation.getReadTimerContext( CassandraModel.CF_METRICS_STATE_NAME );
```
shintasmith: Should be METRICS_METADATA_NAME.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxMetadataIO.java
```diff
@@ -0,0 +1,162 @@
...
+        }
+        finally {
+            ctx.stop();
+        }
+    }
+
+    @Override
+    public void put( Locator locator, String key, String value ) throws IOException {
+
+        Timer.Context ctx = Instrumentation.getWriteTimerContext( CassandraModel.CF_METRICS_STATE_NAME );
```
shintasmith: Same here.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxMetadataIO.java
```diff
@@ -0,0 +1,162 @@
...
+    public void putAll( Table<Locator, String, String> meta ) throws IOException {
+
+        Timer.Context ctx = Instrumentation.getWriteTimerContext( CassandraModel.CF_METRICS_STATE_NAME );
+
+        Session session = DatastaxIO.getSession();
+
+        // TODO:  We need to look into doing async calls on each row.  Batch supposedly is stupid in Cassandra
+        // when your keys aren't on the same partition
+        try {
+            PreparedStatement ps = session.prepare( "INSERT INTO metrics_metadata (key, column1, value) VALUES (?, ?, ?)" );
```
shintasmith: Maybe you can use CF_METRICS_METADATA_NAME here too?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxTestIO.java
```diff
@@ -0,0 +1,53 @@
...
+     */
+    @Override
+    public long getKeyCount( String cf ) throws Exception {
+
+        Session session = DatastaxIO.getSession();
+
+        Select select = select( "key" )
+                .from( cf );
+
+        List<Row> rows = session.execute( select ).all();
```
shintasmith: You can do:
```
session.execute ( select  ).distinct()
```
Then just count the size of the rows.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxShardStateIO.java
```diff
@@ -77,7 +77,7 @@
     @Override
     public void putShardState(int shard, Map<Granularity, Map<Integer, UpdateStamp>> slotTimes) throws IOException {
 
-        Timer.Context ctx = Instrumentation.getReadTimerContext(CassandraModel.CF_METRICS_STATE_NAME);
+        Timer.Context ctx = Instrumentation.getWriteTimerContext(CassandraModel.CF_METRICS_STATE_NAME);
```
shintasmith: Good catch!!!! Thank you!

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/TestIO.java
```diff
@@ -0,0 +1,24 @@
+package com.rackspacecloud.blueflood.io;
+
+/**
+ * Utility class to provide some test methods existing tests use.
+ */
+public interface TestIO {
```
shintasmith: Another idea would be to create an interface, preferably with a different name, and just have the two MetadataIO classes (Astyanax and Datastax) to implement it. But the getKeyCount() would not take a String parameter for the column name. It is implied that it operates on whatever ColumnFamily that is used to store Metadata.

shintasmith:
Can you also remove the AstyanaxReader.getMetadataValues() and AstyanaxWriter.writeMetadataValue() methods? Replace all calls to these methods with the corresponding MetadataIO methods.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/IOContainer.java
usnavi: Yeah, I was thinking about that as well, but was curious if other tests might have similar test helper methods.  I'll move it into integration

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/TestIO.java
usnavi: My thought was to make a helper class that might be usable in other contexts as well, not just for MetadataIO.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxMetadataIO.java
usnavi: Wow, yeah, totally.  Should we be updating our integration tests to run with Datastax on and Astyanax on?  that would have caught this.

shintasmith: What do you mean?

usnavi: nm, i'm crazy

usnavi:
@shintasmith  - updated PR

blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/io/TestIO.java
```diff
@@ -0,0 +1,24 @@
+package com.rackspacecloud.blueflood.io;
+
+/**
+ * Utility class to provide some test methods existing tests use.
+ */
+public interface TestIO {
```
ChandraAddala: The name looks a bit odd to me. I see that it is only used in tests but I think it should still be named appropriately.

usnavi: It's a helper class for testing to be implemented by Astyanax & Datastax.  TestHelperIO?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxMetadataIO.java
```diff
@@ -0,0 +1,162 @@
...
+                    .all()
+                    .from( CassandraModel.CF_METRICS_METADATA_NAME )
+                    .where( eq( KEY, locator.toString() ));
+
+            List<Row> results = session.execute( select ).all();
+
+            Map<String, String> values = new HashMap<String, String>();
+
+            for ( Row row : results ) {
+                if ( LOG.isDebugEnabled() ) {
```
ChandraAddala: I had noticed this log statement in DatastaxShardStateIO also. It is just filling up the debug log. I felt it is too much logging even at debug level.

https://github.com/rackerlabs/blueflood/blob/1c9bebba0454c0a12ea4ea9d9d291a00ef5beaf2/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxShardStateIO.java#L61

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxMetadataIO.java
```diff
@@ -0,0 +1,162 @@
...
+        Session session = DatastaxIO.getSession();
+
+        try {
+            Insert insert = insertInto( CassandraModel.CF_METRICS_METADATA_NAME )
+                    .value( KEY, locator.toString() )
+                    .value( COLUMN1, key )
+                    .value( VALUE, serDes.serialize( value ) );
+            // TODO: This is required by the cassandra-maven-plugin 2.0.0-1, but not by cassandra 2.0.11, which we run.
+            // I believe its due to the bug https://issues.apache.org/jira/browse/CASSANDRA-6238
+            insert.setConsistencyLevel( ConsistencyLevel.ONE );
```
ChandraAddala: If we forget the maven plugin, what is the consistency level we currently use? If it is not "ConsistencyLevel.ONE", arent we sacrificing data consistency to support this plugin?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxShardStateIO.java
```diff
@@ -54,7 +54,7 @@
             Select.Where statement = QueryBuilder
                     .select()
                     .all()
-                    .from(CassandraModel.QUOTED_KEYSPACE, CassandraModel.CF_METRICS_STATE_NAME)
+                    .from(CassandraModel.CF_METRICS_STATE_NAME)
```
ChandraAddala: Where does it refer the keyspace from? I mean how does it know to look for this column family in keyspace "DATA"?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxMetadataIO.java
usnavi: The default is ConsistencyLevel.LOCAL_ONE.  I believe when this was added to Cassandra, SimpleStrategy was not updated to support it.  This was fixed in Cassandra 2.0.3.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxShardStateIO.java
usnavi: I put it  https://github.com/usnavi/blueflood/blob/metadata_datastax2/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxIO.java#L82

blueflood-core/src/test/java/com/rackspacecloud/blueflood/utils/InMemoryMetadataIO.java
```diff
@@ -6,13 +6,14 @@
...
 import com.rackspacecloud.blueflood.io.MetadataIO;
+import com.rackspacecloud.blueflood.io.TestIO;
 import com.rackspacecloud.blueflood.types.Locator;
 
 import java.io.IOException;
 import java.util.Map;
 import java.util.Set;
 
-public class InMemoryMetadataIO implements MetadataIO {
+public class InMemoryMetadataIO implements MetadataIO, TestIO {
```
ChandraAddala: Oh man, multiple inheritance..Havent seen these in a while.

blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/io/astyanax/AstyanaxWriterIntegrationTest.java
```diff
@@ -16,87 +16,66 @@
...
 
         writeMetric("string_metric", "This is a string test");
-        
-        assertNumberOfRows("metrics_string", 1);
-        assertNumberOfRows("metrics_full", 0);
-        assertNumberOfRows("metrics_locator", 0);
+
+        assertEquals( 1, testIO.getKeyCount( CassandraModel.CF_METRICS_STRING_NAME ) );
+        assertEquals( 0, testIO.getKeyCount( CassandraModel.CF_METRICS_FULL_NAME ) );
+        assertEquals( 0, testIO.getKeyCount( CassandraModel.CF_METRICS_LOCATOR_NAME ) );
```
ChandraAddala: if any of these asserts fail, it will be difficult to tell which one did.  If you write a comment as below, it can help.

`assertEquals( "Key count for METRICS_FULL", 0, testIO.getKeyCount( CassandraModel.CF_METRICS_FULL_NAME ) );`

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxMetadataIO.java
```diff
@@ -0,0 +1,162 @@
...
+            List<String> query = new ArrayList<String>();
+
+            for( Locator l : locators) {
+                query.add( l.toString() );
+            }
+
+            Select.Where select = select()
+                    .all()
+                    .from( CassandraModel.CF_METRICS_METADATA_NAME )
+                    .where( in( KEY, query ) );
```
ChandraAddala: I am not familiar with metrics metadata table but I am just curious on the load this query is going to put. Are we actually querying it for all the locators we have? Do you know?

blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/io/TestIO.java
shintasmith: How about CassandraUtilsIO, it's specificly a utility class for doing Cassandra stuff, yes?

blueflood-core/src/test/java/com/rackspacecloud/blueflood/utils/InMemoryMetadataIO.java
shintasmith: It's just an overachiever class, it implements multiple interfaces. :-) I am also creating something like that for the Enum stuff. It's perfectly ok, I think.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxMetadataIO.java
shintasmith: @ChandraAddala, do you think we should change that to TRACE ? Or get rid of it?

blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/io/datastax/DatastaxTestIO.java
```diff
@@ -0,0 +1,47 @@
...
+     * @throws Exception
+     */
+    @Override
+    public long getKeyCount( String cf ) throws Exception {
+
+        Session session = DatastaxIO.getSession();
+
+        Select select = select( "key" )
+                .distinct()
+                .from( cf );
```
shintasmith: Beautiful!

blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/io/astyanax/AstyanaxWriterIntegrationTest.java
shintasmith: :+1: I recommend adding comments for all assert statement. I usually do it with my test.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxMetadataIO.java
usnavi: Let's do TRACE.  I thought it was a lot of logging, but you can always configure by class if necessary.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxMetadataIO.java
```diff
@@ -0,0 +1,162 @@
...
+
+        Timer.Context ctx = Instrumentation.getWriteTimerContext( CassandraModel.CF_METRICS_METADATA_NAME );
+
+        Session session = DatastaxIO.getSession();
+
+        try {
+            Insert insert = insertInto( CassandraModel.CF_METRICS_METADATA_NAME )
+                    .value( KEY, locator.toString() )
+                    .value( COLUMN1, key )
+                    .value( VALUE, serDes.serialize( value ) );
```
VinnyQ: I think we should stay consistent throughout and either using PreparedStatement or the Insert class.  Discussing with shinta I think we like preparedstatement better (reads better).

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxMetadataIO.java
```diff
@@ -0,0 +1,162 @@
...
+    public void putAll( Table<Locator, String, String> meta ) throws IOException {
+
+        Timer.Context ctx = Instrumentation.getWriteTimerContext( CassandraModel.CF_METRICS_METADATA_NAME );
+
+        Session session = DatastaxIO.getSession();
+
+        // TODO:  We need to look into doing async calls on each row.  Batch supposedly is stupid in Cassandra
+        // when your keys aren't on the same partition
+        try {
+            PreparedStatement ps = session.prepare( "INSERT INTO " + CassandraModel.CF_METRICS_METADATA_NAME + " (key, column1, value) VALUES (?, ?, ?)" );
```
VinnyQ: How are you getting by without specifying a keyspace?  Do you have a default connection setup in your blueflood.conf or something?  I am getting  "No keyspace has been specified." error if I do it the same way you have done here for a different table.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxMetadataIO.java
usnavi: Datastax recommends the preparedstatement if you are executing it more than once.  I'm for updating our stuff to preparedstatement if we dont' think it will add too much work.  Otherwise we can do it in a later story.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxMetadataIO.java
usnavi: https://github.com/rackerlabs/blueflood/blob/1c9bebba0454c0a12ea4ea9d9d291a00ef5beaf2/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxShardStateIO.java#L61

blueflood-core/src/test/java/com/rackspacecloud/blueflood/utils/InMemoryMetadataIO.java
ChandraAddala: I wasnt complaining. I was excited to see it.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxMetadataIO.java
ChandraAddala: I dont see much value in those log statements but I would be okay with 'trace' level logging.

usnavi:
@shintasmith @VinnyQ @ChandraAddala - all updates save improving the queries with async queries & prepared statements.

VinnyQ:
:+1:  \0/ :dancers:

PR #661, opened 2016-04-06
Test Rollup Statics

izrik:
This PR adds tests for the static fields in the `Rollup` interface that are populated with anonymous inner classes. It brings the unit test coverage up to 100% for the `Rollup` interface and its inner classes.

### DIFF
--- a/blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/RollupStaticsTest.java
+++ b/blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/RollupStaticsTest.java
@@ -1,0 +1,120 @@
+package com.rackspacecloud.blueflood.types;
+
+import org.junit.Test;
+
+import java.io.IOException;
+
+import static org.junit.Assert.*;
+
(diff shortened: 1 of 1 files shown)


izrik:
@usnavi @ChandraAddala @shintasmith @VinnyQ Please review.

ChandraAddala:
Looks good.

ChandraAddala:
Any idea why the coverage has decreased?

izrik:
@ChandraAddala Some of the coverage numbers are inconsistent. Some of the larger tests (especially the integration tests) that involve lots of classes don't always cover exactly the same paths of execution, simply because of the dynamic nature of the code. Sometimes a timer doesn't get to trigger, sometimes a periodic cleanup operation doesn't happen before the test ends. Some particular execution path somewhere in the code base was not executed this time, but it's not related to the new tests in this PR. As we replace the big lots-of-classes-at-once tests with smaller tests for each class, and isolate them with mocks and dependency injection, the numbers should stabilize. These tests that I'm adding are stable, and will always cover the exact same execution paths.

PR #662, opened 2016-04-07
Test points

izrik:
This PR adds tests for the `Points<T>` and `Points.Point<T>` classes. It brings class-specific unit test coverage to 100% for both.

### DIFF
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/types/Points.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/types/Points.java
@@ -70,14 +70,9 @@ public class Points<T> {
         public boolean equals(Object obj) {
             if (obj == null || !(obj instanceof Point))
                 return false;
-            try {
-                Point<T> other = (Point<T>)obj;
-                return other.getTimestamp() == this.getTimestamp()
-                        && other.getData().equals(this.getData());
-            } catch (ClassCastException ex) {
-                // not a Point<T>, but a Point<X> instead?
-                return false;
-            }
+            Point<T> other = (Point<T>)obj;
+            return other.getTimestamp() == this.getTimestamp()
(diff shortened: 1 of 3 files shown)


izrik:
@ChandraAddala @usnavi @shintasmith @VinnyQ Please review.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/types/Points.java
```diff
@@ -70,14 +70,9 @@ public int hashCode() {
...
-                Point<T> other = (Point<T>)obj;
-                return other.getTimestamp() == this.getTimestamp()
-                        && other.getData().equals(this.getData());
-            } catch (ClassCastException ex) {
-                // not a Point<T>, but a Point<X> instead?
-                return false;
-            }
+            Point<T> other = (Point<T>)obj;
+            return other.getTimestamp() == this.getTimestamp()
+                    && other.getData().equals(this.getData());
```
usnavi: I'm curious, what happens if they do pass in a Point<X>?

usnavi: nm, i see the test.

usnavi:
lgtm

blueflood-core/src/main/java/com/rackspacecloud/blueflood/types/Points.java
izrik: Detail: If they pass, for example, a `Point<X>` to a `Point<Y>`, the cast won't through anything because both types get converted to just `Point`, due to type erasure of java generics. Then, in the `return` statement, `this.getData()` will return a `Y`, and pass it to `X.equals(Object)`. So the `equals` will receive the object as type `Object` and do whatever comparisons it wants to, and no cast exceptions will be thrown.

PR #663, opened 2016-04-11
enum io using datastax

shintasmith:
The main goal of this PR is to provide a set of Datastax driver classes that can read/write enum metrics. Enum is stored in two CFs: metrics_enum and metrics_preaggregated_{granularity}.

The original AstyanaxReader class that provides the functionality to read enums has code that sort of "joins" the two CFs and construct BluefloodEnumRollup objects from the two CFs. The code that does this is intertwined inside AstyanaxReader, but not necessarily all of the code is Astyanax specific. 

In the Datastax world, the code is separated out. The code that deals with strictly reading/writing enums are isolated in the ```DatastaxEnumIO``` class, thus this class can be Datastax driver specific. The code that does the "join" is isolated in the ```EnumMetricDataIO```, and this class is driver agnostic.

The ```PreaggregatedMetricsIO``` classes are classes that deal with providing the Blueflood services a way to read/write a grouping of preaggregated metrics. Currently, it can only deal with Enum, as the other preaggregated metrics IO classes (such as counters, timers, histograms, etc) are not yet written. Eventually, the ```PreaggregatedMetricsIO``` will be added to the IOContainer and Services code will access it via IOContainer.

The majority of the tests are written to ensure the output of the new methods in DatastaxEnumIO and EnumMetricsDataIO are the same as the equivalent methods in AstyanaxReader/Writer. 

Other changes include:

* update to latest Junit to fix some mocking issue
** fix MetricsIntegrationTest so that tests pass regardless of the order they are run. The order has changed with the new Junit version

* add JUnitParams package to enable running parameterized test methods (see EnumIOIntegrationTest)

### DIFF
--- a/blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/io/EnumIOIntegrationTest.java
+++ b/blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/io/EnumIOIntegrationTest.java
@@ -1,0 +1,333 @@
+/*
+ * Copyright (c) 2016 Rackspace.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * 	http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.rackspacecloud.blueflood.io;
+
+import com.datastax.driver.core.ResultSetFuture;
+import com.google.common.collect.Table;
+import com.rackspacecloud.blueflood.io.astyanax.AstyanaxEnumIO;
+import com.rackspacecloud.blueflood.io.astyanax.AstyanaxReader;
+import com.rackspacecloud.blueflood.io.astyanax.AstyanaxWriter;
+import com.rackspacecloud.blueflood.io.datastax.DatastaxEnumIO;
+import com.rackspacecloud.blueflood.outputs.formats.MetricData;
+import com.rackspacecloud.blueflood.rollup.Granularity;
+import com.rackspacecloud.blueflood.types.*;
+import junitparams.JUnitParamsRunner;
+import junitparams.Parameters;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import java.util.*;
+
+import static org.junit.Assert.*;
+
+/**
+ * Integration tests covering EnumIO
+ */
+public class EnumIOIntegrationTest extends IntegrationTestBase  {
+
+    protected Map<Locator, List<IMetric>> locatorToMetrics;
+    protected DatastaxEnumIO datastaxEnumIO = new DatastaxEnumIO();
+    protected AstyanaxEnumIO astyanaxEnumIO = new AstyanaxEnumIO();
+
+    /**
+     * This method is to supply the granularity parameter to some test methods below
+     *
+     * @return
+     */
+    protected Object getGranularitiesToTest() {
+        return new Object[] {
+                Granularity.FULL,
+                Granularity.MIN_5,
+                Granularity.MIN_20,
+                Granularity.MIN_60,
+                Granularity.MIN_240,
+                Granularity.MIN_1440
+        };
+    }
+
+
+    @Before
+    public void generateEnum() throws Exception {
+        locatorToMetrics = generateEnumForTenants();
+    }
+
+    @RunWith(JUnitParamsRunner.class)
+    public static class WriteDatastaxReadAstyanaxReader extends EnumIOIntegrationTest {
+
+        /**
+         * Writes enums using Datastax.
+         * <p/>
+         * Get the locator -> enum string mapping using Astyanax
+         * (i.e: call astyanaxReader.getEnumStringMappings()).
+         * <p/>
+         * Verify that the metrics_enum CF is written properly
+         *
+         * @throws Exception
+         */
+        @Test
+        public void shouldExistEnumHashToValueMapping() throws Exception {
+
+            // write using Datastax
+            for (Map.Entry<Locator, List<IMetric>> entry : locatorToMetrics.entrySet()) {
+                Locator locator = entry.getKey();
+                List<IMetric> metrics = entry.getValue();
+                for (IMetric metric : metrics) {
+                    datastaxEnumIO.putAsync(locator, metric.getCollectionTime(), (BluefloodEnumRollup) metric.getMetricValue(), Granularity.FULL);
+                }
+            }
+
+            // read using AstyanaxReader
+            AstyanaxReader reader = AstyanaxReader.getInstance();
+            final List<Locator> locators = new ArrayList<Locator>() {{
+                addAll(locatorToMetrics.keySet());
+            }};
+
+            // read the enums from metrics_enum CF, via getEnumStringMappings()
+            Map<Locator, List<String>> mapping = reader.getEnumStringMappings(locators);
+            org.junit.Assert.assertEquals("locator -> enum string mapping size", locatorToMetrics.size(), mapping.size());
+
+            for (Map.Entry<Locator, List<String>> entry : mapping.entrySet()) {
+                List<String> enumValues = entry.getValue();
+                assertTrue(String.format("metrics size for locator %s is not bigger than possible enum values defined for these tests", entry.getKey()),
+                        enumValues.size() <= enumValueList.size());
+            }
+        }
+
+        /**
+         * Writes enums using Datastax.
+         * <p/>
+         * Get the locator -> metric data mapping using Astyanax
+         * (i.e: call astyanaxReader.getEnumMetricDataForRangeForLocatorList())
+         *
+         * @throws Exception
+         */
+        @Test
+        @Parameters(method = "getGranularitiesToTest")
+        public void shouldExistInMetricsPreaggregatedGran(Granularity granularity) throws Exception {
+
+            // write using Datastax
+            for (Map.Entry<Locator, List<IMetric>> entry : locatorToMetrics.entrySet()) {
+                Locator locator = entry.getKey();
+                List<IMetric> metrics = entry.getValue();
+                for (IMetric metric : metrics) {
+                    ResultSetFuture future = datastaxEnumIO.putAsync(locator, metric.getCollectionTime(), (BluefloodEnumRollup) metric.getMetricValue(), granularity);
+                    future.get().all();
+                }
+            }
+
+            // read delay!
+            Thread.sleep(500);
+
+            // read using Astyanax
+            AstyanaxReader reader = AstyanaxReader.getInstance();
+            final List<Locator> locators = new ArrayList<Locator>() {{
+                addAll(locatorToMetrics.keySet());
+            }};
+
+            // ask for 5 minutes back
+            Map<Locator, MetricData> results = reader.getEnumMetricDataForRangeForLocatorList(locators,
+                                                        getRangeFromMinAgoToNow(5), granularity);
+            assertEquals("Results size", locatorToMetrics.size(), results.size());
+
+            for (Locator locator : locators) {
+                MetricData metricData = results.get(locator);
+                Points points = metricData.getData();
+
+                Set<Map.Entry> entries = points.getPoints().entrySet();
+                List<IMetric> expectedMetrics = locatorToMetrics.get(locator);
+                assertEquals("metrics size for locator " + locator, expectedMetrics.size(), entries.size());
+
+                for (IMetric metric : expectedMetrics) {
+                    Points.Point point = (Points.Point) points.getPoints().get(metric.getCollectionTime());
+                    assertNotNull(String.format("point for timestamp %d exists", metric.getCollectionTime()), point);
+                }
+            }
+        }
+    }
+
+    @RunWith(JUnitParamsRunner.class)
+    public static class WriteDatastaxReadAstyanaxEnumIO extends EnumIOIntegrationTest {
+
+        /**
+         * Writes enums using Datastax.
+         * <p/>
+         * Get the locator -> enum string mapping using AstyanaxEnumIO
+         * <p/>
+         * Verify that metrics_enum hash to string mapping is written properly
+         *
+         * @throws Exception
+         */
+        @Test
+        public void shouldExistEnumHashToValueMapping() throws Exception {
+
+            // write using Datastax
+            for (Map.Entry<Locator, List<IMetric>> entry : locatorToMetrics.entrySet()) {
+                Locator locator = entry.getKey();
+                List<IMetric> metrics = entry.getValue();
+                for (IMetric metric : metrics) {
+                    datastaxEnumIO.putAsync(locator, metric.getCollectionTime(), (BluefloodEnumRollup) metric.getMetricValue(), Granularity.FULL);
+                }
+            }
+
+            // read using Astyanax
+            final List<Locator> locators = new ArrayList<Locator>() {{
+                addAll(locatorToMetrics.keySet());
+            }};
+
+            // read the enums from metrics_enum CF, via getEnumStringMappings()
+            Table<Locator, Long, String> table = astyanaxEnumIO.getEnumHashValuesForLocators(locators);
+            Set<Locator> locatorSet = table.rowKeySet();
+            assertEquals("Table(locator, hash, value): # of locators", locatorToMetrics.size(), locatorSet.size());
+            for (Map.Entry<Locator, Map<Long, String>> entry : table.rowMap().entrySet()) {
+                Locator locator = entry.getKey();
+                Map<Long, String> hashValueMap = entry.getValue();
+                assertTrue(String.format("metrics size for locator %s is not bigger than possible enum values defined for these tests", locator),
+                        hashValueMap.size() <= enumValueList.size());
+            }
+        }
+
+        /**
+         * Writes enums using Datastax.
+         * <p/>
+         * Get the locator -> enum string mapping using AstyanaxEnumIO
+         *
+         * @throws Exception
+         */
+        @Test
+        @Parameters(method = "getGranularitiesToTest")
+        public void shouldExistInMetricsPreaggregatedGran(Granularity granularity) throws Exception {
+
+            // write using Datastax
+            for (Map.Entry<Locator, List<IMetric>> entry : locatorToMetrics.entrySet()) {
+                Locator locator = entry.getKey();
+                List<IMetric> metrics = entry.getValue();
+                for (IMetric metric : metrics) {
+                    datastaxEnumIO.putAsync(locator, metric.getCollectionTime(), (BluefloodEnumRollup) metric.getMetricValue(), granularity);
+                }
+            }
+
+            // read delay!
+            Thread.sleep(500);
+
+            // read using Astyanax
+            final List<Locator> locators = new ArrayList<Locator>() {{
+                addAll(locatorToMetrics.keySet());
+            }};
+
+            // read the enums from metrics_enum CF, via getEnumStringMappings()
+            // ask for 5 minutes back
+            Table<Locator, Long, BluefloodEnumRollup> table = astyanaxEnumIO.getEnumRollupsForLocators(locators,
+                                                                    CassandraModel.getColumnFamily(BluefloodEnumRollup.class, granularity).getName(),
+                                                                    getRangeFromMinAgoToNow(5));
+            Set<Locator> locatorSet = table.rowKeySet();
+            assertEquals("Table(locator, timestamp, rollup): # of locators", locatorToMetrics.size(), locatorSet.size());
+            for (Map.Entry<Locator, Map<Long, BluefloodEnumRollup>> entry : table.rowMap().entrySet()) {
+                Locator locator = entry.getKey();
+                Map<Long, BluefloodEnumRollup> enumRollupMap = entry.getValue();
+
+                // check and make sure all timestamp in our test data is there
+                List<IMetric> expectedMetrics = locatorToMetrics.get(locator);
+                for (IMetric expected : expectedMetrics) {
+                    BluefloodEnumRollup enumRollup = enumRollupMap.get(expected.getCollectionTime());
+                    assertNotNull(String.format("enumRollup exists at timestamp %s", expected.getCollectionTime()), enumRollup);
+                }
+                assertTrue(String.format("metrics size for locator %s is not bigger than possible enum values defined for these tests", locator),
+                        enumRollupMap.size() <= enumValueList.size());
+            }
+        }
+    }
+
+    @RunWith(JUnitParamsRunner.class)
+    public static class WriteAstyanaxReadDatastax extends EnumIOIntegrationTest {
+
+        @Test
+        public void shouldExistEnumHashToValueMapping() throws Exception {
+
+            // organize our test data in the way the API wants it
+            List<IMetric> metrics = new ArrayList<IMetric>() {{
+                for (Locator locator : locatorToMetrics.keySet()) {
+                    addAll(locatorToMetrics.get(locator));
+                }
+            }};
+
+            // write using Astyanax
+            AstyanaxWriter writer = AstyanaxWriter.getInstance();
+            writer.insertMetrics(metrics, CassandraModel.CF_METRICS_PREAGGREGATED_FULL);
+
+            // read using Datastax
+            final List<Locator> locators = new ArrayList<Locator>() {{
+                addAll(locatorToMetrics.keySet());
+            }};
+            Table<Locator, Long, String> table = datastaxEnumIO.getEnumHashValuesForLocators(locators);
+            Set<Locator> locatorSet = table.rowKeySet();
+            assertEquals("Table(locator, hash, value): # of locators", locatorToMetrics.size(), locatorSet.size());
+
+            for (Locator locator : locators) {
+                Map<Long, String> map = table.row(locator);
+                // This test will be run multiple times so it's possible previous
+                // runs of this test will have inserted different enum values.
+                // Enum values for these tests are randomly generated, see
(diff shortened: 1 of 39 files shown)


shintasmith:
@usnavi @ChandraAddala @VinnyQ @izrik , please review. There is a compilation error with JDK 1.7 about ArrayList and List incompatibility. I will fix it when I get in, but you can start reviewing.

blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/io/EnumIOIntegrationTest.java
```diff
@@ -0,0 +1,333 @@
...
+    }
+
+
+    @Before
+    public void generateEnum() throws Exception {
+        locatorToMetrics = generateEnumForTenants();
+    }
+
+    @RunWith(JUnitParamsRunner.class)
+    public static class WriteDatastaxReadAstyanaxReader extends EnumIOIntegrationTest {
```
usnavi: I had no idea that you could extend the enclosing class, but I guess that it does make sense for static inner classes.

I assume you did this so you didn't have to have multiple files?  Are there other reasons?

blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/io/EnumIOIntegrationTest.java
```diff
@@ -0,0 +1,333 @@
...
+
+            // read using AstyanaxReader
+            AstyanaxReader reader = AstyanaxReader.getInstance();
+            final List<Locator> locators = new ArrayList<Locator>() {{
+                addAll(locatorToMetrics.keySet());
+            }};
+
+            // read the enums from metrics_enum CF, via getEnumStringMappings()
+            Map<Locator, List<String>> mapping = reader.getEnumStringMappings(locators);
+            org.junit.Assert.assertEquals("locator -> enum string mapping size", locatorToMetrics.size(), mapping.size());
```
usnavi: minor, do we want to remove ```org.junit.Assert.``` prefix?

blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/io/EnumIOIntegrationTest.java
shintasmith: I am doing this to show I'm cool 😉.

JK, I'm doing it because I can organize the tests into groups this way. They are all testing Enum stuff, so in a way I think they need to be together, they share some common utility code. But the static inner classes groups several tests together. This class groups the tests related to writing with datastax, reading with the old AstyanaxReader. There's another one for writing with datastax, reading with the new AstayanxEnumIO, etc.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxEnumIO.java
```diff
@@ -0,0 +1,322 @@
...
+ * using Datastax driver.
+ */
+public class DatastaxEnumIO implements AsyncWriter, EnumReader {
+
+    private static final Logger LOG = LoggerFactory.getLogger(DatastaxEnumIO.class);
+    private static final EnumSerDes serDes = new EnumSerDes();
+
+    private static final String INSERT_KEY_COLUMN_VALUE_FORMAT = "INSERT INTO %s (key, column1, value) VALUES (?, ?, ?)";
+    private static final String SELECT_FOR_KEY = "SELECT * FROM %s WHERE key = :locator";
+    private static final String SELECT_FOR_KEY_RANGE_FORMAT = "SELECT * FROM %s WHERE key = :locator AND column1 >= :tsStart AND column1 <= :tsEnd";
```
usnavi: If you want, you can use ```QueryBuilder``` and ```bindMarker()``` to create these strings.  E.g., https://gist.github.com/usnavi/b7920c76b9b36d582d43df22a72c349c#file-gistfile1-txt-L45

shintasmith: Any advantage to using QueryBuilder and bindMarker, instead of just the String like above? I find it easier to read with the String , to be honest...

blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/io/EnumIOIntegrationTest.java
usnavi: and that would prompt @ChandraAddala to come up with an even longer method name :-P

blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/io/EnumIOIntegrationTest.java
shintasmith: Done.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/EnumMetricDataIO.java
```diff
@@ -0,0 +1,199 @@
...
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
+
+/**
+ * This class uses the driver specific {@link EnumReader} to
+ * read/write enum metrics from Cassandra and convert them to
+ * {@link com.rackspacecloud.blueflood.outputs.formats.MetricData}
+ * objects that are used by services such as Rollup, Ingest and Query
+ */
+public class EnumMetricDataIO {
```
usnavi: I was under the impression that all *IO classes were direct implementations of of a particular Cassandra driver.  I think that's a pretty good thing to follow, albeit no better names are really coming to mind.

Some ideas:
* EnumMetrics
* EnumMetricData
* EnumMetricAccessor

shintasmith: Yeah, we can follow that and I can rename the class. I think I like EnumMetricData. MetricData is actually a class of the return object in most of these methods.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/EnumMetricDataIO.java
```diff
@@ -0,0 +1,199 @@
...
+                populateEnumValueToCountMap(enumHashRollup.row(locator), enumHashValues.row(locator));
+                Points points = convertToPoints(enumHashRollup.row(locator));
+                MetricData metricData = new MetricData(points, getUnitString(locator), MetricData.Type.ENUM);
+                resultMap.put(locator, metricData);
+            }
+        } catch (InterruptedException e) {
+            LOG.error("Interrupted while reading Enum metrics for locators " + locators, e);
+        } catch (ExecutionException e) {
+            LOG.error("Execution error while reading Enum metrics for locators " + locators, e);
+        }
```
usnavi: I assume ignoring the error and continuing was what we did before.  Do we want to make a story to go back and think how best to handle errors and what to return the consumer?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/IOContainer.java
```diff
@@ -34,21 +35,31 @@
...
     // more IO classes to follow
 
     /**
-     * Creates an instance of this class based on what configuration says our
+     * Returns an instance of this class based on what configuration says our
      * driver should be.
      *
      * @return IOContainer
      */
     public static IOContainer fromConfig() {
```
ChandraAddala: I think you have synchronize the method also. If not, it wont be thread safe.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/IOContainer.java
```diff
@@ -34,21 +35,31 @@
...
-        String driver = configuration.getStringProperty(CoreConfig.CASSANDRA_DRIVER);
-        return new IOContainer(DriverType.getDriverType(driver));
+        if ( FROM_CONFIG_INSTANCE == null ) {
+            synchronized (IOContainer.class) {
+                String driver = configuration.getStringProperty(CoreConfig.CASSANDRA_DRIVER);
```
usnavi: i think you need one more ```== null`` check in case 2 threads got through the if before the singleton is initialized.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/astyanax/PreaggregatedMetricsIO.java
```diff
@@ -0,0 +1,64 @@
...
+import com.rackspacecloud.blueflood.types.IMetric;
+
+import java.io.IOException;
+import java.util.Collection;
+
+/**
+ * This class deals with reading/writing metrics to the metrics_preaggregated_* column families
+ * using Astyanax driver
+ */
+public class PreaggregatedMetricsIO extends AbstractMetricsIO {
```
usnavi: Would it make it easier to follow if started with Astyanax, like ```AstyanaxPreAgMetricsIO``` or something like that?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/AsyncWriter.java
```diff
@@ -0,0 +1,42 @@
...
+import com.rackspacecloud.blueflood.rollup.Granularity;
+import com.rackspacecloud.blueflood.types.Locator;
+import com.rackspacecloud.blueflood.types.Rollup;
+
+/**
+ * An interface describing the behavior of classes who can
+ * asynchronously insert a metric to cassandra. This only
+ * applies to classes that use datastax driver.
+ */
+public interface AsyncWriter {
```
usnavi: Why do we need this interface?  If its tied to the Datastax implementation (like returning ```ResultSetFuture```) it makes it less necessary, right?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxEnumIO.java
usnavi: No advantage, and it looks like you are you using ```%s```, so this works even better.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxEnumIO.java
```diff
@@ -0,0 +1,321 @@
...
+
+    /**
+     * Read the metrics_enum column family for the specified locators. Organize
+     * the data as a table of locator, enum value hash, and enum value.
+     * This is a representation on how the data looks in the column family.
+     *
+     * @param locators
+     * @return
+     */
+    public Table<Locator, Long, String> getEnumHashValuesForLocators(final List<Locator> locators) {
```
usnavi: Stick an ```@Override``` on this.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxEnumIO.java
```diff
@@ -0,0 +1,321 @@
...
+    /**
+     * Read the metrics_preaggregated_{granularity} for specific locators.
+     * This is to be implemented by driver specific class.
+     *
+     * @param locators
+     * @param columnFamily
+     * @param range
+     * @return
+     */
+    public Table<Locator, Long, BluefloodEnumRollup> getEnumRollupsForLocators(final List<Locator> locators, String columnFamily, Range range) {
```
usnavi: @Override

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/PreaggregatedMetricsIO.java
```diff
@@ -0,0 +1,114 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ExecutionException;
+
+/**
+ * This class deals with reading/writing metrics to the metrics_preaggregated_* column families
+ * using Datastax driver
+ */
+public class PreaggregatedMetricsIO extends AbstractMetricsIO {
```
usnavi: same question as I had with the Astyanax implementation.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/utils/Locators.java
```diff
@@ -0,0 +1,37 @@
...
+
+import com.rackspacecloud.blueflood.types.Locator;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Utilities to operate on {@link com.rackspacecloud.blueflood.types.Locator} objects
+ */
+public class Locators {
```
usnavi: I like ```LocatorUtils``` but I can do what the group wants.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/IOContainer.java
shintasmith: I think I'll just synchronize the whole method :-P

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/AsyncWriter.java
shintasmith: I need it in Datastax's PreaggregatedMetrics. I am keeping a map of rollupType => corresponding Async Writer. 
https://github.com/rackerlabs/blueflood/pull/663/files/f4170585379e101e680979d9051da390b9aeac1a#diff-579f34fdebe138b9a4ab835a28731440R47

When PreaggregatedMetrics gets a list of IMetrics objects, it needs to be able to call the right AsyncWriter to write that to the CFs. It is a Datastax specific interface only, thus it lives in the datastax package. I am not writing re-writing the Astyanax logic to do the same, because eventually we will be getting rid of them.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/astyanax/PreaggregatedMetricsIO.java
shintasmith: I guess. I was trying to make the name shorter. I don't like PreAg. So it will have to be AstyanaxPreaggregatedMetrics.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/EnumMetricDataIO.java
shintasmith: Yes, that was what we did before. Here's the equivalent code in AstyanaxReader:
https://github.com/rackerlabs/blueflood/blob/master/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/astyanax/AstyanaxReader.java#L471

I am torn as to what is the right behavior. If we read and got Interrupted or Execution exception, do we bail out the whole thing? What if the read of metrics_enum gets exception but the other read (of metrics_preaggregated_*) works ok? Do we retry? How long do we retry before giving up? In the mean time, caller is waiting for this result.

shintasmith:
@usnavi @ChandraAddala , I incorporated your comments:
* rename EnumMetricsDataIO to EnumMetricsData
* rename datastax.PreaggregatedMetricsIO to DatastaxPreaggregatedMetrics, same as astyanax one
* rename Locators to LocatorsUtil
* synchronize the IOContainer.getInstance()

blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/io/EnumIOIntegrationTest.java
```diff
@@ -0,0 +1,333 @@
...
+         */
+        @Test
+        public void shouldExistEnumHashToValueMapping() throws Exception {
+
+            // write using Datastax
+            for (Map.Entry<Locator, List<IMetric>> entry : locatorToMetrics.entrySet()) {
+                Locator locator = entry.getKey();
+                List<IMetric> metrics = entry.getValue();
+                for (IMetric metric : metrics) {
+                    datastaxEnumIO.putAsync(locator, metric.getCollectionTime(), (BluefloodEnumRollup) metric.getMetricValue(), Granularity.FULL);
```
ChandraAddala: Just a question. I see that below test calls `rollup delay 10 mins.` after the putAsync call. I dont see this method calling it. Also there is a `Thread.sleep` going on in the below test. What is the difference between the two tests in writing using datastax.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/EnumReader.java
```diff
@@ -0,0 +1,79 @@
...
+     * Locates all enum values from metrics_enum column family by their
+     * {@link com.rackspacecloud.blueflood.types.Locator}.
+     *
+     * The result is organized in a map of Locator -> list of String. The string
+     * is the string enum values.
+     *
+     * @param locators
+     * @return
+     */
+    public Map<Locator, List<String>> getEnumStringMappings(final List<Locator> locators);
```
ChandraAddala: Should this be `public Map<Locator, Set<String>>` ?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/EnumReader.java
```diff
@@ -0,0 +1,79 @@
...
+ * Row (key)    | Column1         | Value
+ * -------------|-----------------|-------------------
+ * metric name  | enum hash code  | enum string value
+ *
+ * The metrics_preaggregated_{granularity} column family is a table of:
+ * Row (key)    | Column1         | Value
+ * -------------|-----------------|-------------------
+ * metric name  | timestamp       | serialized BluefloodEnumRollup object
+ */
+public interface EnumReader {
```
usnavi: Should the be EnumReaderIO or am I being too naming convention crazy?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/astyanax/AstyanaxPreaggregatedMetrics.java
```diff
@@ -0,0 +1,64 @@
...
+import com.rackspacecloud.blueflood.types.IMetric;
+
+import java.io.IOException;
+import java.util.Collection;
+
+/**
+ * This class deals with reading/writing metrics to the metrics_preaggregated_* column families
+ * using Astyanax driver
+ */
+public class AstyanaxPreaggregatedMetrics extends AbstractMetricsIO {
```
usnavi: Should this end in IO as well?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/EnumReader.java
shintasmith: IO implies both read and write, no?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/EnumReader.java
shintasmith: Lmao! I just converted some methods from having input type ArrayList to List. This technically should be Set. But if I change this, I have to change some Service code that calls this method (I think it's EnumValidator). I was trying to avoid changing too much Service code. Do you think I should in this PR?

blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/io/IntegrationTestBase.java
```diff
@@ -270,8 +275,64 @@ protected void generateEnumRollups(Locator locator, long from, long to, Granular
...
+        long now = System.currentTimeMillis();
+        return new Range(now - (deltaMin*60*1000), now);
+    }
+
+    protected Map<Locator, List<IMetric>> generateEnumForTenants() throws Exception {
+        Map<Locator, List<IMetric>> locatorToMetrics = new HashMap<Locator, List<IMetric>>();
+
+        // setup some Enum data
+        long startTime = System.currentTimeMillis();
+        final Locator locator1 = Locator.createLocatorFromPathComponents(TENANT1, "my", "enum", "values");
```
ChandraAddala: Should we put test name in the metric name?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/astyanax/AstyanaxPreaggregatedMetrics.java
shintasmith: You certainly have something against IO! :-) Good catch. I will rename.

blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/io/IntegrationTestBase.java
shintasmith: These metrics are being deleted after every test. But I can add test name for clarity.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxPreaggregatedMetrics.java
```diff
@@ -0,0 +1,113 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ExecutionException;
+
+/**
+ * This class deals with reading/writing metrics to the metrics_preaggregated_* column families
+ * using Datastax driver
+ */
+public class DatastaxPreaggregatedMetrics extends AbstractMetricsIO {
```
usnavi: IO?

blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/io/EnumIOIntegrationTest.java
shintasmith: I think this one was a miss. I will add future.get().all(). 

The sleep below was needed because I wasn't getting the results. But I will re-test after I add the future().get().all and see if I still need the sleep. If not, I will remove.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/EnumReader.java
usnavi: I was going under the heuristic that the [driver]*IO were all driver-specific implementations of a generic *IO interface.

I think it could go either way, whatever convention is most descriptive and easiest to follow.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/astyanax/AstyanaxPreaggregatedMetrics.java
usnavi: against IO? I feel like I'm IO's biggest fan!

blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/io/EnumMetricDataIOIntegrationTest.java
```diff
@@ -0,0 +1,113 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+import static org.junit.Assert.*;
+
+/**
+ * This is the tests for {@link EnumMetricData} class
+ */
+public class EnumMetricDataIOIntegrationTest extends IntegrationTestBase {
```
ChandraAddala: Test has to be renamed as well.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/EnumMetricData.java
```diff
@@ -0,0 +1,199 @@
...
+     * The result is organized in a map of Locator -> {@link com.rackspacecloud.blueflood.outputs.formats.MetricData}.
+     *
+     * @param locators
+     * @param range
+     * @param gran
+     * @return
+     */
+    public Map<Locator, MetricData> getEnumMetricDataForRangeForLocatorList(final List<Locator> locators, final Range range, final Granularity gran) {
+        String columnFamily = CassandraModel.getColumnFamily(BluefloodEnumRollup.class, gran).getName();
+        if ( columnFamily == null ) {
```
ChandraAddala: Does this ever happen?

shintasmith: Am I hearing Chandra say "no validation needed" ???? 😱

shintasmith: Or ```throw new ShouldNeverHappenException()``` ?

ChandraAddala: :) Looking at this method `CassandraModel.getColumnFamily`, it didnt give me an impression that we will get `null` as return value.

ChandraAddala:
I only have some minor comments. But apart from that everything looks good.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/EnumReader.java
shintasmith: This is a bigger change that I'd like to bite with this PR. What happens is that this method is called from EnumValidator. What EnumValidator does is: it reads enums from Cassandra, then sort+compare with enums it reads from ElasticSearch. The API from ElasticSearch returns a List. So everytime there's a code that compares the two (EnumValidatorTest has quite a few places) results, I have to convert the result from Cassandra from Set to List.

So for now, let's leave it as List.

shintasmith:
@usnavi @ChandraAddala , I incorporated your comments:
* rename MetricsIO, AbstractMetricsIO to MetricsRW, AbstractMetricsRW, RW for ReaderWriter
* rename EnumReader to EnumReaderIO
* rename EnumMetricDataIOIntegrationTest to EnumMetricDataIntegrationTest
* fix issues with tests not calling future.get().all(), get rid of unnecessary sleep
* add class name to the metrics generated by the tests

usnavi:
I assume the idea is that that the difference between RW & IO is that IO is for wrapping driver functions, while RW has more blueflood-specific logic in it?

shintasmith:
@usnavi, yes, IO is for classes that deal with specific drivers. RW is for classes that uses IO classes to construct the right objects for Blueflood Services. For you, I added a package-info.java:
https://github.com/rackerlabs/blueflood/pull/663/commits/cca5b15420c1bf752b4b597f671922860d713ce2

usnavi:
AWESOME! lgtm

PR #667, opened 2016-04-14
LocatorIO interaface with Datastax and Astyanax implementation

VinnyQ:
# What
Create new classes to reimplement the AstyanaxWriter.insertLocator() and AstyanaxReader.getLocatorsToRollup() methods with both an Asytanax implementation and Datastax implementation.

#  Why
Preparation for switching blueflood to use Datastax instead of Astyanax driver for communicating with cassandra.

#  How
Create LocatorIO interface with Datastax and Astyanax implementation,
  wire new LocatorIO classes in IOContainer,
  replace wherever AstyanaxReader.getLocatorsToRollup is called

### DIFF
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/LocatorIO.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/LocatorIO.java
@@ -1,0 +1,39 @@
+/*
+ * Copyright (c) 2016 Rackspace.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * 	http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.rackspacecloud.blueflood.io;
+
+import com.rackspacecloud.blueflood.types.Locator;
+
+import java.io.IOException;
+import java.util.Collection;
+
+public interface LocatorIO {
+
+    /**
+     * Insert a locator for a shard, calculate by hashing of the locator string value
+     * @param locator
+     * @throws IOException
+     */
+    public void insertLocator(Locator locator) throws IOException;
+
+    /**
+     * @param shard
+     * @return a collection of the locators objects corresponding to the given shard
+     * @throws IOException
+     */
+    public Collection<Locator> getLocators(long shard) throws IOException;
+
+}
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/astyanax/AstyanaxLocatorIO.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/astyanax/AstyanaxLocatorIO.java
@@ -1,0 +1,88 @@
+/*
+ * Copyright (c) 2016 Rackspace.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * 	http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.rackspacecloud.blueflood.io.astyanax;
+
+import com.codahale.metrics.Timer;
+import com.netflix.astyanax.MutationBatch;
+import com.netflix.astyanax.connectionpool.exceptions.ConnectionException;
+import com.netflix.astyanax.connectionpool.exceptions.NotFoundException;
+import com.netflix.astyanax.query.RowQuery;
+import com.rackspacecloud.blueflood.io.CassandraModel;
+import com.rackspacecloud.blueflood.io.Instrumentation;
+import com.rackspacecloud.blueflood.io.LocatorIO;
+import com.rackspacecloud.blueflood.types.Locator;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.util.Collection;
+import java.util.Collections;
+
+/**
+ * This class uses the Astyanax driver to read/write locators from
+ * Cassandra metrics_locator Column Family.
+ */
+public class AstyanaxLocatorIO implements LocatorIO {
+
+    private static final Logger LOG = LoggerFactory.getLogger(AstyanaxLocatorIO.class);
+
+    /**
+     * Insert a locator with key = shard long value calculated using Util.getShard()
+     * @param locator
+     * @throws IOException
+     */
+    @Override
+    public void insertLocator(Locator locator) throws IOException {
+        try {
+            MutationBatch mutationBatch = AstyanaxIO.getKeyspace().prepareMutationBatch();
+            AstyanaxWriter.getInstance().insertLocator(locator, mutationBatch);
+            mutationBatch.execute();
+        } catch (Exception e) {
+            throw new IOException(e);
+        }
+    }
+
+    /**
+     * Returns the locators for a shard, i.e. those that should be rolled up, for a given shard.
+     * 'Should' means:
+     *  1) A locator is capable of rollup (it is not a string/boolean metric).
+     *  2) A locator has had new data in the past LOCATOR_TTL seconds.
+     *
+     * @param shard Number of the shard you want the locators for. 0-127 inclusive.
+     * @return Collection of locators
+     * @throws IOException
+     */
+    @Override
+    public Collection<Locator> getLocators(long shard) throws IOException {
+        Timer.Context ctx = Instrumentation.getReadTimerContext(CassandraModel.CF_METRICS_LOCATOR_NAME);
+        try {
+            RowQuery<Long, Locator> query = AstyanaxIO.getKeyspace()
+                    .prepareQuery(CassandraModel.CF_METRICS_LOCATOR)
+                    .getKey(shard);
+            return query.execute().getResult().getColumnNames();
+        } catch (NotFoundException e) {
+            Instrumentation.markNotFound(CassandraModel.CF_METRICS_LOCATOR_NAME);
+            return Collections.emptySet();
+        } catch (ConnectionException ex) {
+            Instrumentation.markReadError(ex);
+            LOG.error("Connection exception during getLocators(" + Long.toString(shard) + " )", ex);
+            throw new IOException("Error reading locators", ex);
+        } finally {
+            ctx.stop();
+        }
+    }
+
+}
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxLocatorIO.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxLocatorIO.java
@@ -1,0 +1,148 @@
+/*
+ * Copyright (c) 2016 Rackspace.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * 	http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.rackspacecloud.blueflood.io.datastax;
+
+import com.codahale.metrics.Timer;
+import com.datastax.driver.core.*;
+import com.datastax.driver.core.querybuilder.Insert;
+import com.datastax.driver.core.querybuilder.QueryBuilder;
+import com.datastax.driver.core.querybuilder.Select;
+import com.rackspacecloud.blueflood.cache.TenantTtlProvider;
+import com.rackspacecloud.blueflood.io.CassandraModel;
+import com.rackspacecloud.blueflood.io.Instrumentation;
+import com.rackspacecloud.blueflood.io.LocatorIO;
+import com.rackspacecloud.blueflood.types.Locator;
+import com.rackspacecloud.blueflood.utils.Util;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+
+import static com.datastax.driver.core.querybuilder.QueryBuilder.bindMarker;
+import static com.datastax.driver.core.querybuilder.QueryBuilder.eq;
+import static com.datastax.driver.core.querybuilder.QueryBuilder.ttl;
+
+/**
+ * This class uses the Datastax driver to read/write locators from
+ * Cassandra metrics_locator Column Family.
+ */
+public class DatastaxLocatorIO implements LocatorIO {
+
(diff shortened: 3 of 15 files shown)


VinnyQ:
@usnavi @shintasmith @ChandraAddala please review when you have a chance.  Thanks!

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/astyanax/AstyanaxIO.java
```diff
@@ -116,7 +116,7 @@ private static ConnectionPoolConfigurationImpl createPreferredConnectionPoolConf
         return connectionPoolConfiguration;
     }
 
-    protected static Keyspace getKeyspace() {
+    public static Keyspace getKeyspace() {
```
shintasmith: Why do we need to make this public?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/astyanax/AstyanaxLocatorIO.java
```diff
@@ -0,0 +1,88 @@
...
+                    .prepareQuery(CassandraModel.CF_METRICS_LOCATOR)
+                    .getKey(shard);
+            return query.execute().getResult().getColumnNames();
+        } catch (NotFoundException e) {
+            Instrumentation.markNotFound(CassandraModel.CF_METRICS_LOCATOR_NAME);
+            return Collections.emptySet();
+        } catch (ConnectionException ex) {
+            Instrumentation.markReadError(ex);
+            LOG.error("Connection exception during getLocatorsToRollup", ex);
+            throw new IOException("Error reading locators", ex);
```
shintasmith: Print the shard in the exception message.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/astyanax/AstyanaxLocatorIO.java
```diff
@@ -0,0 +1,88 @@
...
+     * 'Should' means:
+     *  1) A locator is capable of rollup (it is not a string/boolean metric).
+     *  2) A locator has had new data in the past LOCATOR_TTL seconds.
+     *
+     * @param shard Number of the shard you want the locators for. 0-127 inclusive.
+     * @return Collection of locators
+     * @throws IOException
+     */
+    @Override
+    public Collection<Locator> getLocatorsToRollup(long shard) throws IOException {
```
shintasmith: I noticed that in insertLocator() it is just a wrapper that calls AstyanaxWriter.getInstance().insertLocator(). But here, the  code is copied from AstyanaxReader. Curious why the different scheme. I am fine with either, but let's do it one way. :-)

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxLocatorIO.java
```diff
@@ -0,0 +1,148 @@
...
+ * This class uses the Datastax driver to read/write locators from
+ * Cassandra metrics_locator Column Family.
+ */
+public class DatastaxLocatorIO implements LocatorIO {
+
+    private static final Logger LOG = LoggerFactory.getLogger(DatastaxLocatorIO.class);
+
+    public static final String KEY = "key";
+    public static final String COLUMN1 = "column1";
+    public static final String VALUE = "value";
```
shintasmith: Do these need to be public?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxLocatorIO.java
```diff
@@ -0,0 +1,148 @@
...
+     * Create all prepared statements use in this class for metrics_locator
+     */
+    private void createPreparedStatements()  {
+
+        // create a generic select statement for retrieving from metrics_locator
+        Select.Where select = QueryBuilder
+                .select()
+                .all()
+                .from( CassandraModel.CF_METRICS_LOCATOR_NAME )
+                .where( eq ( "key", bindMarker() ));
```
shintasmith: Maybe use your public static final String KEY here?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxLocatorIO.java
```diff
@@ -0,0 +1,148 @@
...
+        getValue = DatastaxIO.getSession().prepare( select );
+
+        // create a generic insert statement for inserting into metrics_locator
+        Insert insert = QueryBuilder.insertInto(CassandraModel.QUOTED_KEYSPACE, CassandraModel.CF_METRICS_LOCATOR_NAME)
```
shintasmith: You don't need to use the QUOTED_KEYSPACE anymore.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxLocatorIO.java
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    /**
+     * Insert a locator with key = shard long value calculated using Util.getShard()
+     * @param locator
+     * @throws IOException
+     */
+    @Override
+    public void insertLocator(Locator locator) throws IOException {
+        Session session = DatastaxIO.getSession();
```
shintasmith: Do we need to put some Instrumentation here?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxLocatorIO.java
```diff
@@ -0,0 +1,148 @@
...
+     *  2) A locator has had new data in the past LOCATOR_TTL seconds.
+     *
+     * @param shard Number of the shard you want the locators for. 0-127 inclusive.
+     * @return Collection of locators
+     * @throws IOException
+     */
+    @Override
+    public Collection<Locator> getLocatorsToRollup(long shard) throws IOException {
+
+        Timer.Context ctx = Instrumentation.getReadTimerContext(CassandraModel.CF_METRICS_LOCATOR.getName());
```
shintasmith: Use the CassandraModel.CF_METRICS_LOCATOR_NAME.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/io/IOContainerTest.java
```diff
@@ -33,7 +33,7 @@
         "com.rackspacecloud.blueflood.utils.Metrics",
         "com.codahale.metrics.*"})
 @PrepareForTest({Configuration.class})
-@SuppressStaticInitializationFor( "com.rackspacecloud.blueflood.io.datastax.DatastaxIO")
+@SuppressStaticInitializationFor( { "com.rackspacecloud.blueflood.io.datastax.DatastaxIO", "com.rackspacecloud.blueflood.io.CassandraModel" })
```
shintasmith: Why do we need to add the CassandraModel here?

shintasmith:
@VinnyQ , in general, I think I am fine with the approach you have taken for the LocatorIO. Just a few minor comments. :-)

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/astyanax/AstyanaxLocatorIO.java
usnavi: my vote is to call it from AstyanaxReader, no need to copy.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/LocatorFetchRunnableDrainExecutionContextTest.java
```diff
@@ -17,6 +24,10 @@
 
 import static org.mockito.Mockito.*;
 
+@PowerMockIgnore({"javax.management.*", "com.rackspacecloud.blueflood.utils.Metrics", "com.codahale.metrics.*"})
+@PrepareForTest({ IOContainer.class })
+@RunWith(PowerMockRunner.class)
+@SuppressStaticInitializationFor( "com.rackspacecloud.blueflood.io.IOContainer" )
```
usnavi: is this required?

blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/LocatorFetchRunnableRunTest.java
```diff
@@ -17,14 +24,18 @@
 
 import static org.mockito.Mockito.*;
 
+@PowerMockIgnore({"javax.management.*", "com.rackspacecloud.blueflood.utils.Metrics", "com.codahale.metrics.*"})
+@PrepareForTest({ IOContainer.class })
+@RunWith(PowerMockRunner.class)
+@SuppressStaticInitializationFor( "com.rackspacecloud.blueflood.io.IOContainer" )
```
usnavi: required?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/LocatorIO.java
```diff
@@ -0,0 +1,39 @@
...
+     * @throws IOException
+     */
+    public void insertLocator(Locator locator) throws IOException;
+
+    /**
+     * @param shard
+     * @return a collection of the locators objects corresponding to the given shard
+     * @throws IOException
+     */
+    public Collection<Locator> getLocatorsToRollup(long shard) throws IOException;
```
ChandraAddala: I think this method name should just be `getLocators(long shard)`

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/astyanax/AstyanaxIO.java
VinnyQ: I need it to set up a mutation batch for LocatorIOIntegrationTest to call astyanaxWriter.insertLocator.  That test class is in package com.rackspacecloud.blueflood.io, and test both Datastax and Astyanax stuff.  Do you rather we get rid of that test there?  (see test writeOldAstyanaxReadDatastax in LocatorIOIntegrationTest).  Or leave this method public?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/astyanax/AstyanaxLocatorIO.java
VinnyQ: @shintasmith @usnavi 
I originally did it this way AND removed the method () from AstyanaxReader.  Nothing should be calling the AstyanaxReader.getLocatorsToRollup() now except as a test in LocatorIOIntegrationTest to verify that it's behaving the same.  Would it be better if I remove the method from AstyanaxReader and just get rid of the "Old" tests?  Or leaveing it and make this a wrapper method instead?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxLocatorIO.java
VinnyQ: Cool!  This was the cause of the instantiate error.  Now I can get rid of that too!

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxLocatorIO.java
VinnyQ: There was no instrumentation fir insertLocator in the original code, so I didn't add it here.  Don't want to instrument/send more things to graphite if we are not tracking it.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/io/IOContainerTest.java
VinnyQ: I was getting static initializing error for the CassandModel.QUOTED_KEYSPACE call in DatastaxLocatorIO.  But not having to use that anymore makes this no longer necessary.  Removed!  Thanks!

blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/LocatorFetchRunnableDrainExecutionContextTest.java
VinnyQ: @usnavi 
Yes if we still want to truly test verifyZeroInteractions(locatorIO);

But looking at these tests it seems a bit bloated with all these verifyZeroInteractions, so we can just get rid of the assert for verifyZeroInteractions(locatorIO); and we wouldn't need to set up the static mock (because of the change in LocatorFetchRunnable).

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxLocatorIO.java
VinnyQ: nope!  fixed!

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxLocatorIO.java
VinnyQ: fixed!

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/LocatorIO.java
VinnyQ: Good call, it's not use for anything else.  Changing it!

blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/LocatorFetchRunnableRunTest.java
VinnyQ: see above.  Only if we still want to test verifyZeroInteractions(locatorIO) with true mocking.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/astyanax/AstyanaxIO.java
shintasmith: I vote for removing the tests: writeDatastaxReadOldAstyanax() and writeOldAstyanaxReadDatastax(), because those 2 tests are not much different than writeDatastaxReadAstyanax() and writeAstyanaxReadDatastax(). Eventually they will call the methods in AstyanaxWriter.insertLocator(), yes? Or am I completely wrong on that?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/astyanax/AstyanaxLocatorIO.java
shintasmith: I vote for removing it from AstyanaxReader.java

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxLocatorIO.java
shintasmith: Let's leave it without Instrumentation, then. I was thinking we instrument all reads/writes to Cassandra, but it looks like we only do it for selected CFs.

VinnyQ:
@shintasmith @usnavi @ChandraAddala please review again when you have a chance!  thanks!

VinnyQ:
@shintasmith @usnavi @ChandraAddala please review again when you have a chance.  Thanks!

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/astyanax/AstyanaxLocatorIO.java
```diff
@@ -0,0 +1,88 @@
...
+     *  1) A locator is capable of rollup (it is not a string/boolean metric).
+     *  2) A locator has had new data in the past LOCATOR_TTL seconds.
+     *
+     * @param shard Number of the shard you want the locators for. 0-127 inclusive.
+     * @return Collection of locators
+     * @throws IOException
+     */
+    @Override
+    public Collection<Locator> getLocators(long shard) throws IOException {
+        Timer.Context ctx = Instrumentation.getReadTimerContext(CassandraModel.CF_METRICS_LOCATOR.getName());
```
shintasmith: Please use CassandraModel.CF_METRIC_LOCATOR_NAME

shintasmith: Why not use CassandraModel.CF_METRICS_LOCATOR_NAME ?

shintasmith:
Other than one minor comment, it looks good. :+1:

PR #669, opened 2016-04-14
Test ScheduleContext

izrik:
This PR adds tests for the `ScheduleContext` class. It brings the class-specific unit test coverage up to 86%.

### DIFF
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/ScheduleContext.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/ScheduleContext.java
@@ -156,6 +156,18 @@ public class ScheduleContext implements IngestionContext, ScheduleContextMBean {
         this(currentTimeMillis, managedShards, new DefaultClockImpl());
         registerMBean();
     }
+    @VisibleForTesting
+    public ScheduleContext(long currentTimeMillis,
+                           Collection<Integer> managedShards,
+                           Clock clock,
+                           ShardStateManager shardStateManager,
+                           ShardLockManager shardLockManager) {
+        this.scheduleTime = currentTimeMillis;
+        this.shardStateManager = shardStateManager;
+        this.lockManager = shardLockManager;
+        this.clock = clock;
+        registerMBean();
+    }
 
     public void setCurrentTimeMillis(long millis){ scheduleTime = millis; }
     public long getCurrentTimeMillis() { return scheduleTime; }
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/ScheduleContext.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/ScheduleContext.java
@@ -264,7 +264,7 @@ public class ScheduleContext implements IngestionContext, ScheduleContextMBean {
         }
     }
 
-    private boolean areChildKeysOrSelfKeyScheduledOrRunning(SlotKey slotKey) {
+    boolean areChildKeysOrSelfKeyScheduledOrRunning(SlotKey slotKey) {
         // if any ineligible (children and self) keys are running or scheduled to run, we shouldn't work on this.
         Collection<SlotKey> ineligibleKeys = slotKey.getChildrenKeys();
 
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/ScheduleContext.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/ScheduleContext.java
@@ -268,13 +268,19 @@ public class ScheduleContext implements IngestionContext, ScheduleContextMBean {
         // if any ineligible (children and self) keys are running or scheduled to run, we shouldn't work on this.
         Collection<SlotKey> ineligibleKeys = slotKey.getChildrenKeys();
 
-        if (runningSlots.keySet().contains(slotKey) || scheduledSlots.contains(slotKey)) {
+        if (runningSlots.keySet().contains(slotKey)) {
+            return true;
+        }
+        if (scheduledSlots.contains(slotKey)) {
             return true;
         }
 
         // if any ineligible keys are running or scheduled to run, do not schedule this key.
         for (SlotKey childrenKey : ineligibleKeys) {
-            if (runningSlots.keySet().contains(childrenKey) || scheduledSlots.contains(childrenKey)) {
+            if (runningSlots.keySet().contains(childrenKey)) {
+                return true;
+            }
+            if (scheduledSlots.contains(childrenKey)) {
                 return true;
             }
         }
--- a/blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/ScheduleContextAddRemoveShardsTest.java
+++ b/blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/ScheduleContextAddRemoveShardsTest.java
@@ -1,0 +1,121 @@
+package com.rackspacecloud.blueflood.service;
+
+import com.rackspacecloud.blueflood.utils.DefaultClockImpl;
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.invocation.InvocationOnMock;
+import org.mockito.stubbing.Answer;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.mockito.Matchers.anyInt;
+import static org.mockito.Mockito.*;
+
+public class ScheduleContextAddRemoveShardsTest {
+
+
+    long currentTime;
+    List<Integer> managedShards;
+
+    ShardStateManager shardStateManager;
+    ShardLockManager lockManager;
+
+    ScheduleContext ctx;
+
+    @Before
+    public void setUp() {
+
+        // given
+        currentTime = 1234000L;
+        managedShards = new ArrayList<Integer>() {{
+            add(0);
+        }};
+
+        shardStateManager = mock(ShardStateManager.class);
+        lockManager = mock(ShardLockManager.class);
+
+        ctx = new ScheduleContext(currentTime, managedShards,
+                new DefaultClockImpl(), shardStateManager, lockManager);
+    }
+
+    @Test
+    public void addShardAddsShard() {
+
+        // given
+        final Integer[] capture1 = new Integer[1];
+        doAnswer(new Answer() {
+            @Override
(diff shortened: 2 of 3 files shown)


izrik:

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/ScheduleContext.java
```diff
@@ -252,17 +264,23 @@ void scheduleEligibleSlots(long maxAgeMillis, long delayedMetricsMaxAgeMillis) {
...
+    boolean areChildKeysOrSelfKeyScheduledOrRunning(SlotKey slotKey) {
         // if any ineligible (children and self) keys are running or scheduled to run, we shouldn't work on this.
         Collection<SlotKey> ineligibleKeys = slotKey.getChildrenKeys();
 
-        if (runningSlots.keySet().contains(slotKey) || scheduledSlots.contains(slotKey)) {
+        if (runningSlots.keySet().contains(slotKey)) {
+            return true;
+        }
+        if (scheduledSlots.contains(slotKey)) {
             return true;
```
usnavi: any reason you did this?  personally, i prefer using the ||

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/ScheduleContext.java
```diff
@@ -252,17 +264,23 @@ void scheduleEligibleSlots(long maxAgeMillis, long delayedMetricsMaxAgeMillis) {
...
         }
 
         // if any ineligible keys are running or scheduled to run, do not schedule this key.
         for (SlotKey childrenKey : ineligibleKeys) {
-            if (runningSlots.keySet().contains(childrenKey) || scheduledSlots.contains(childrenKey)) {
+            if (runningSlots.keySet().contains(childrenKey)) {
+                return true;
+            }
+            if (scheduledSlots.contains(childrenKey)) {
```
usnavi: same here.  I usually put the ||...on the second line, seems more readable to me.

izrik: @usnavi It has to do with the way jacoco calculates branch coverage. With both conditions in the same if-statement, with a `||` between them, there are theoretically four possible setups: `(T || T)`, `(T || F)`, `(F || T)`, `(F || F)`. The first one is not possible, because a slot should never be both scheduled and running. By breaking up the conditionals into two independent branches, they can be tracked by jacoco separately. Each one says "all of 2 branches covered" instead of "1 of 4 branches missed" in the combined case.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/ScheduleContext.java
izrik: See below.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/ScheduleContext.java
ChandraAddala: I am not in favor of changing code to increase jococo coverage.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/ScheduleContextAddRemoveShardsTest.java
```diff
@@ -0,0 +1,121 @@
...
+        // then
+        verify(shardStateManager).add(anyInt());
+        verifyNoMoreInteractions(shardStateManager);
+        assertNotNull(capture1[0]);
+        assertEquals(shard, capture1[0].intValue());
+
+        verify(lockManager).addShard(anyInt());
+        verifyNoMoreInteractions(lockManager);
+        assertNotNull(capture2[0]);
+        assertEquals(shard, capture2[0].intValue());
```
ChandraAddala: Should we verify that `getManagedShards()` returns this shard?

blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/ScheduleContextAddRemoveShardsTest.java
```diff
@@ -0,0 +1,121 @@
...
+        // then
+        verify(shardStateManager).remove(anyInt());
+        verifyNoMoreInteractions(shardStateManager);
+        assertNotNull(capture1[0]);
+        assertEquals(shard, capture1[0].intValue());
+
+        verify(lockManager).removeShard(anyInt());
+        verifyNoMoreInteractions(lockManager);
+        assertNotNull(capture2[0]);
+        assertEquals(shard, capture2[0].intValue());
```
ChandraAddala: Should we verify that `getManagedShards()` does not return this?

blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/ScheduleContextAddRemoveShardsTest.java
izrik: @ChandraAddala That's a method on `ShardStateManager`. Such a test case would be more suited to `ShardStateManagerTest`.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/ScheduleContextAreKeysRunningTest.java
```diff
@@ -0,0 +1,212 @@
...
+        ctx.scheduleEligibleSlots(1, 7200000);
+
+        SlotKey runningSlot = ctx.getNextScheduled();
+
+        // precondition
+        assertEquals(0, ctx.getScheduledCount());
+        assertEquals(1, ctx.getRunningCount());
+
+        ctx.clearFromRunning(runningSlot);
+        ctx.pushBackToScheduled(runningSlot, false);
```
ChandraAddala: I beliece only one of these happen at a time. If execution was successful we call `clearFromRunning` but if it is not we call `pushBackToScheduled`.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/ScheduleContext.java
izrik: @ChandraAddala I understand the sentiment, and I'm not normally one for changing the _meaning_ of the code, under normal circumstances. But this doesn't change the logic at all. For all I know, _it will compile to exactly the same bytecode_. And if not, it will compile to trivial equivalent bytecode. It's less of a change than what jacoco does to instrument the code. Personally, I find it a little easier to read because it doesn't run off the right side of the screen.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/ScheduleContextAddRemoveShardsTest.java
ChandraAddala: sounds good.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/ScheduleContextAreKeysRunningTest.java
izrik: @ChandraAddala Kinda. The code is not consistent on this. If `clearFromRunning` doesn't get called, then the slot never gets removed from `runningSlots`, and that means that after `pushBackToScheduled`, the slot would be in both `runningSlots` and `scheduledSlots` at the same time. The code doesn't handle the failure mode very well.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/ScheduleContextAddRemoveShardsTest.java
izrik: @ChandraAddala See above.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/ScheduleContextAreKeysRunningTest.java
ChandraAddala: If the behavior of code is that it is not removing from `runningSlots` shouldn't the test case reflect that? Why do we have to compensate it by calling both when in reality only one of these gets called? Do I make sense or am I missing something?

izrik: @ChandraAddala The discrepancy is already documented in `ScheduleContextPushBackToScheduledTest`. See here: https://github.com/izrik/blueflood/blob/test-schedule-context/blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/ScheduleContextPushBackToScheduledTest.java#L66-L69 
I'm compensating for it here based on what I believe the intention of the code is. I'lltry to clear it up some.

izrik: @ChandraAddala I tried to clarify the setup of the test case. Does that answer your question sufficiently?

ChandraAddala: I undestand your explanation but looking at it just gives an impression that these two calls happens one after another in a workflow which isnt the case. It might be appropriate to your test case as you documented here. I was just wondering what would `areChildKeysOrSelfKeyScheduledOrRunning` return when we only do pushBackToScheduled which is a valid scenario.

izrik: @ChandraAddala It's a _possible_ scenario, although I wouldn't consider it a _valid_ one. Calling `pushBackToScheduled` while the slot is still in the running category makes it both running and scheduled. I don't believe that was the ultimate intention of the original authors, given how `getNextScheduled` is so careful to separate the two. Admittedly, this is all testing around failure conditions which I don't think the original authors adequately considered. Nevertheless, I've split it out into two separate test cases, one with and one without the call to `clearFromRunning`.

ChandraAddala: I agree it is a possible scenario. My argument is during that workflow. `clearfromRunning` will not be called.

izrik: @ChandraAddala I added a test for that.

ChandraAddala:
Looks good.

izrik:
@usnavi Any other question/comments/suggestions/concerns?

usnavi:
lgtm

PR #670, opened 2016-04-14
stub out MetricsRW interface

shintasmith:
This is to allow parallel development between @usnavi (who's doing datastax metrics_{gran}) and me (who's finishing up datastax metrics_preaggregated_{gran})


usnavi:
lgtm

PR #671, opened 2016-04-14
Extract the SlotStateManager class to its own file.

izrik:
The `SlotStateManager` was previously an inner class within `ShardStateManager`. However, inner classes can't be instantiated apart from an instance of their containing class. This makes them difficult to test. In order to write tests specifically for `SlotStateManager`, it needs to be a non-nested class in its own file. This PR does just that. After this PR is merged, I can start writing more tests.


izrik:
@usnavi @ChandraAddala @shintasmith @VinnyQ Please review.

izrik:
This may have to be rebased after #669 gets merged.

PR #673, opened 2016-04-15
Test Locator

izrik:
This PR adds tests for the `Locator` class. It brings the class-specific unit test coverage from zero up to 96% of lines.

### DIFF
--- a/blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/LocatorTest.java
+++ b/blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/LocatorTest.java
@@ -307,6 +307,17 @@ public class LocatorTest {
     }
 
     @Test
+    public void equalsThisNullThrowsException() {
+
+        // given
+        Locator locator = new Locator();
+        Locator other = Locator.createLocatorFromDbKey("a.b.d");
+
+        // expect
+        assertFalse(locator.equals((Object)other));
+    }
+
+    @Test
     public void equalsOtherNullThrowsException() {
 
         // given
--- a/blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/LocatorTest.java
+++ b/blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/LocatorTest.java
@@ -296,7 +296,7 @@ public class LocatorTest {
     }
 
     @Test
-    public void equalsBothNullReturnsTrue() {
+    public void equalsBothUninitializedReturnsTrue() {
 
         // given
         Locator locator = new Locator();
@@ -307,7 +307,7 @@ public class LocatorTest {
     }
 
     @Test
-    public void equalsThisNullThrowsException() {
+    public void equalsThisUninitializedReturnsFalse() {
 
         // given
         Locator locator = new Locator();
@@ -318,7 +318,7 @@ public class LocatorTest {
     }
 
     @Test
-    public void equalsOtherNullThrowsException() {
+    public void equalsOtherUninitializedReturnsFalse() {
 
         // given
         Locator locator = Locator.createLocatorFromDbKey("a.b.c");
--- a/blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/LocatorTest.java
+++ b/blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/LocatorTest.java
@@ -432,7 +432,4 @@ public class LocatorTest {
         // then
         assertEquals(-1, comparison);
     }
-
-
-
 }
--- a/blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/LocatorTest.java
+++ b/blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/LocatorTest.java
@@ -1,0 +1,4 @@
+package com.rackspacecloud.blueflood.types;
+
+public class LocatorTest {
+}
--- a/blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/LocatorTest.java
+++ b/blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/LocatorTest.java
@@ -307,17 +307,6 @@ public class LocatorTest {
     }
 
     @Test
-    public void equalsThisNullThrowsException() {
-
-        // given
-        Locator locator = new Locator();
-        Locator other = Locator.createLocatorFromDbKey("a.b.d");
-
-        // expect
-        assertFalse(locator.equals((Object)other));
-    }
-
-    @Test
     public void equalsOtherNullThrowsException() {
 
         // given
--- a/blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/LocatorTest.java
+++ b/blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/LocatorTest.java
@@ -147,6 +147,7 @@ public class LocatorTest {
 
     @Test(expected = NullPointerException.class)
     public void isValidDBKeyThrowsExceptionOnNullDbKey() {
+        // TODO: This is incorrect behavior. The test is merely documenting it. It should be fixed in the future.
 
         // given
         Locator locator = new Locator();  // the method is not static, so we have to create an instance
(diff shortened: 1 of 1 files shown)


izrik:

blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/LocatorTest.java
```diff
@@ -0,0 +1,435 @@
...
+        // given
+        Locator locator = Locator.createLocatorFromDbKey("a.b.c");
+        Locator other = new Locator();
+
+        // expect
+        assertFalse(locator.equals(other));
+    }
+
+    @Test(expected = NullPointerException.class)
+    public void equals2OtherNullThrowsException() {
```
usnavi: I assume we'd rather not have these.  If so, would you make a comment that you are verifying current behavior and we'd rather return a false rather than null.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/LocatorTest.java
```diff
@@ -0,0 +1,435 @@
...
+public class LocatorTest {
+
+    final String tenant = "tenant";
+    final String metricName = "some.metric.name";
+    final String fullyQualifiedName = tenant + "." + metricName;
+
+    @Test
+    public void publicConstructorDoesNotSetStringRepresentation() {
+        // when
+        Locator locator = new Locator();
```
ChandraAddala: My recommendation is to remove the empty Locator constructor.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/LocatorTest.java
```diff
@@ -0,0 +1,435 @@
...
+    }
+
+    @Test(expected = NullPointerException.class)
+    public void isValidDBKeyThrowsExceptionOnNullDbKey() {
+
+        // given
+        Locator locator = new Locator();  // the method is not static, so we have to create an instance
+
+        // when
+        locator.isValidDBKey(null, "a");
```
ChandraAddala: Can you fix isValidDBKey to do the right thing?

blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/LocatorTest.java
```diff
@@ -0,0 +1,435 @@
...
+
+    @Test
+    public void equalsDifferentStringReturnsFalse() {
+
+        // given
+        Locator locator = Locator.createLocatorFromDbKey("a.b.c");
+        Locator other = Locator.createLocatorFromDbKey("a.b.d");
+
+        // expect
+        assertFalse(locator.equals((Object)other));
```
ChandraAddala: Do you have to cast it to "Object"? I dont think thats necessary.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/LocatorTest.java
```diff
@@ -0,0 +1,435 @@
...
+    }
+
+    @Test(expected = NullPointerException.class)
+    public void equals2OtherNullThrowsException() {
+
+        // given
+        Locator locator = Locator.createLocatorFromDbKey("a.b.c");
+
+        // when
+        boolean result = locator.equals((Locator)null);
```
ChandraAddala: I dont understand why this throws exception? Looks like "null" is handled in the equals method. 

``
@Override
    public boolean equals(Object obj) {
        return obj != null && obj instanceof Locator && obj.hashCode() == this.hashCode();
    }
``

blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/LocatorTest.java
izrik: Yes. There are two `equals` methods, one that takes an `Object` and one that takes a `Locator`. I wanted this test to call the first one, so I explicitly cast to `Object`. Otherwise, it would resolve to the second overload.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/LocatorTest.java
izrik: @ChandraAddala That's not really part of the story. I'll add a comment about it and add a new story to change it.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/LocatorTest.java
izrik: @ChandraAddala We can do that in a separate story.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/LocatorTest.java
ChandraAddala: That doesnt sound right. May be we should remove the equals method that is not being overridden.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/LocatorTest.java
izrik: @usnavi Will do.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/LocatorTest.java
izrik: @ChandraAddala It's calling the following overload:
```

    public boolean equals(Locator other) {
        return stringRep.equals(other.toString());
    }
```
If `other` is `null`, then an exception will be thrown.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/LocatorTest.java
izrik: @usnavi Done.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/LocatorTest.java
izrik: @ChandraAddala Fair enough, but as above, that's not a part of the current story.

izrik:
@usnavi @ChandraAddala Any other questions, or shall I merge this?

ChandraAddala:
Looks good to me.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/LocatorTest.java
```diff
@@ -0,0 +1,441 @@
...
+        // given
+        Locator locator = new Locator();
+        Locator other = new Locator();
+
+        // expect
+        assertTrue(locator.equals((Object)other));
+    }
+
+    @Test
+    public void equalsThisNullThrowsException() {
```
usnavi: i might have missed this.  Is this test named correctly?  It doesn't throw an exception.

izrik: @usnavi That was an oversight on my part. I have removed that test case. The coverage numbers stayed the same.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/LocatorTest.java
```diff
@@ -0,0 +1,441 @@
...
+        // given
+        Locator locator = new Locator();
+        Locator other = Locator.createLocatorFromDbKey("a.b.d");
+
+        // expect
+        assertFalse(locator.equals((Object)other));
+    }
+
+    @Test
+    public void equalsOtherNullThrowsException() {
```
usnavi: FYI - there is another one here.

izrik: @usnavi I take it back. They weren't duplicate tests. You were right the first time. They're misnamed. I have clarified the names so they describe what the tests actually check for.

izrik:
@usnavi Any other questions?

usnavi:
lgtm

PR #681, opened 2016-04-21
Only call register once

izrik:
This PR prevents `registerMBean` from being called multiple times.

### DIFF
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/ScheduleContext.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/ScheduleContext.java
@@ -151,9 +151,11 @@ public class ScheduleContext implements IngestionContext, ScheduleContextMBean {
         this.clock = clock;
         registerMBean();
     }
-
     public ScheduleContext(long currentTimeMillis, Collection<Integer> managedShards) {
-        this(currentTimeMillis, managedShards, new DefaultClockImpl());
+        this.scheduleTime = currentTimeMillis;
+        this.shardStateManager = new ShardStateManager(managedShards, asMillisecondsSinceEpochTicker());
+        this.lockManager = new NoOpShardLockManager();
+        this.clock = new DefaultClockImpl();
         registerMBean();
     }
     @VisibleForTesting
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/ScheduleContext.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/ScheduleContext.java
@@ -483,7 +483,12 @@ public class ScheduleContext implements IngestionContext, ScheduleContextMBean {
         return results;
     }
 
-    private void registerMBean() {
+    private boolean registered = false;
+    private synchronized void registerMBean() {
+
+        if (registered) return;
+        registered = true;
+
         try {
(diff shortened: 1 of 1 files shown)


izrik:
@VinnyQ @ChandraAddala @shintasmith @usnavi Please review.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/ScheduleContext.java
```diff
@@ -151,9 +151,11 @@ public ScheduleContext(long currentTimeMillis, Collection<Integer> managedShards
         this.clock = clock;
         registerMBean();
     }
-
     public ScheduleContext(long currentTimeMillis, Collection<Integer> managedShards) {
-        this(currentTimeMillis, managedShards, new DefaultClockImpl());
+        this.scheduleTime = currentTimeMillis;
+        this.shardStateManager = new ShardStateManager(managedShards, asMillisecondsSinceEpochTicker());
+        this.lockManager = new NoOpShardLockManager();
+        this.clock = new DefaultClockImpl();
```
shintasmith: why repeat this? calling this() should be the correct thing to do, just don't call registerMBean() in the subsequent constructor.

izrik: My first instinct is to have one and only one constructor that sets fields, and to have all other constructors call that one. However, in this case, java didn't like the call to `asMillisecondsSinceEpochTicker` (i.e. referencing `this`) happening before the call to the supertype constructor is called. Also, it seemed strange to have the other constructors defer to one that was `@VisibleForTesting`. My second choice was to have all of the constructors set fields, and none defer to any other, because that way at least they're consistent. I have changed it to just remove the `registerMBean` call in the second constructor.

izrik:
@shintasmith how does it look now?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/ScheduleContext.java
```diff
@@ -481,7 +480,12 @@ public long read() {
         return results;
     }
 
-    private void registerMBean() {
+    private boolean registered = false;
```
ChandraAddala: The name of `isRegistered` is more appropriate for naming boolean fields.

shintasmith:
looks much better, @izrik ! :+1: I will let you settle the variable name with @ChandraAddala . PR looks good otherwise. :-)

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/ScheduleContext.java
izrik: @ChandraAddala what about `mbeanHasBeenRegistered`? I think that would clarify *what* it is that is being registered. Or what about `isMbeanRegistered`?

ChandraAddala: My only concern was with preceding the variable name with "is" which is typical of naming boolean fields in java.  I am fine with the rest of it.

izrik: @ChandraAddala done

ChandraAddala:
Looks good.

PR #682, opened 2016-04-22
Test BluefloodCounterRollup

izrik:
This PR adds tests for the `BluefloodCounterRollup` class. It brings the class-specific unit test coverage from 63% up to 98% of lines, and from 47% to 93% of branches. Also, the old tests were pretty bad, and did not properly test the functionality, so I replaced them.

### DIFF
--- a/blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/BluefloodCounterRollupTest.java
+++ b/blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/BluefloodCounterRollupTest.java
@@ -7,7 +7,6 @@ import org.junit.Test;
 import java.io.IOException;
 import java.util.Random;
 
-import static org.hamcrest.CoreMatchers.instanceOf;
 import static org.hamcrest.CoreMatchers.is;
 import static org.junit.Assert.*;
 
@@ -282,7 +281,7 @@ public class BluefloodCounterRollupTest {
         assertNotNull(count);
         assertThat(count, is(CoreMatchers.<Number>instanceOf(Long.class)));
         assertEquals(3L, count.longValue());
-        assertEquals(Double.POSITIVE_INFINITY, rollup.getRate(), 0.00001d); //(3)/0=inf
+        assertEquals(Double.POSITIVE_INFINITY, rollup.getRate(), 0.00001d); //(3)/(1-1)=inf
         assertEquals(1, rollup.getSampleCount());
     }
 
@@ -299,7 +298,7 @@ public class BluefloodCounterRollupTest {
         assertNotNull(count);
         assertThat(count, is(CoreMatchers.<Number>instanceOf(Long.class)));
         assertEquals(7L, count.longValue());
-        assertEquals(7.0d, rollup.getRate(), 0.00001d); // (3+4)/1=7
+        assertEquals(7.0d, rollup.getRate(), 0.00001d); // (3+4)/(2-1)=7
         assertEquals(2, rollup.getSampleCount());
     }
 
@@ -371,7 +370,7 @@ public class BluefloodCounterRollupTest {
         assertNotNull(count);
         assertThat(count, is(CoreMatchers.<Number>instanceOf(Long.class)));
         assertEquals(12L, count.longValue());
-        assertEquals(6.0d, rollup.getRate(), 0.00001d); // (3+4+5)/2=6
+        assertEquals(6.0d, rollup.getRate(), 0.00001d); // (3+4+5)/(3-1)=6
         assertEquals(3, rollup.getSampleCount());
     }
 
@@ -391,7 +390,7 @@ public class BluefloodCounterRollupTest {
         assertNotNull(count);
         assertThat(count, is(CoreMatchers.<Number>instanceOf(Long.class)));
         assertEquals(18L, count.longValue());
-        assertEquals(6.0d, rollup.getRate(), 0.00001d); // (3+4+5+6)/3=6
+        assertEquals(6.0d, rollup.getRate(), 0.00001d); // (3+4+5+6)/(4-1)=6
         assertEquals(4, rollup.getSampleCount());
     }
 
@@ -449,7 +448,7 @@ public class BluefloodCounterRollupTest {
         Points<SimpleNumber> input = new Points<SimpleNumber>();
         input.add(new Points.Point<SimpleNumber>(1234L, new SimpleNumber(Double.MIN_NORMAL)));
         input.add(new Points.Point<SimpleNumber>(1235L, new SimpleNumber(Double.MIN_NORMAL)));
-        double expectedCount = Double.MIN_NORMAL * 2; // Long.MIN_VALUE + 2;
+        double expectedCount = Double.MIN_NORMAL * 2;
         // when
         BluefloodCounterRollup rollup = BluefloodCounterRollup.buildRollupFromRawSamples(input);
         // then
@@ -552,7 +551,7 @@ public class BluefloodCounterRollupTest {
         assertNotNull(count);
         assertThat(count, is(CoreMatchers.<Number>instanceOf(Long.class)));
         assertEquals(3L, count.longValue()); // 1+2
-        assertEquals(3.0d, rollup.getRate(), 0.00001d); // (1+2+3+4)/2=5
+        assertEquals(3.0d, rollup.getRate(), 0.00001d); // (1+2+3+4)/(3-1)=5
         assertEquals(2, rollup.getSampleCount());
     }
 
@@ -582,7 +581,7 @@ public class BluefloodCounterRollupTest {
         assertNotNull(count);
         assertThat(count, is(CoreMatchers.<Number>instanceOf(Long.class)));
         assertEquals(10L, count.longValue()); // 1+2+3+4
-        assertEquals(5.0d, rollup.getRate(), 0.00001d); // (1+2+3+4)/(1+1)=5
+        assertEquals(5.0d, rollup.getRate(), 0.00001d); // (1+2+3+4)/((2-1)+(2-1))=5
         assertEquals(4, rollup.getSampleCount());
     }
 
@@ -611,7 +610,7 @@ public class BluefloodCounterRollupTest {
         assertNotNull(count);
         assertThat(count, is(CoreMatchers.<Number>instanceOf(Long.class)));
         assertEquals(6L, count.longValue()); // 1+2+3
-        assertEquals(6.0d, rollup.getRate(), 0.00001d); // (1+2+3)/(1+0)=6
+        assertEquals(6.0d, rollup.getRate(), 0.00001d); // (1+2+3)/((2-1)+(1-1))=6
         assertEquals(3, rollup.getSampleCount());
     }
 
@@ -636,7 +635,7 @@ public class BluefloodCounterRollupTest {
         assertNotNull(count);
         assertThat(count, is(CoreMatchers.<Number>instanceOf(Long.class)));
         assertEquals(6L, count.longValue()); // 1+2+3
-        assertEquals(3.0d, rollup.getRate(), 0.00001d); // (1+2+3)/(2)=3
+        assertEquals(3.0d, rollup.getRate(), 0.00001d); // (1+2+3)/(3-1)=3
         assertEquals(3, rollup.getSampleCount());
     }
 
--- a/blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/BluefloodCounterRollupTest.java
+++ b/blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/BluefloodCounterRollupTest.java
@@ -6,10 +6,7 @@ import org.junit.Test;
 import java.io.IOException;
 import java.util.Random;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNotSame;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.*;
 
 public class BluefloodCounterRollupTest {
     private static final Random random = new Random(72262L);
@@ -163,5 +160,15 @@ public class BluefloodCounterRollupTest {
         BluefloodCounterRollup a = new BluefloodCounterRollup().withCount(1).withSampleCount(1);
         //when
         assertTrue(a.equals(b));
+    }
+
+    @Test
+    public void constructorSetsInitialValues() {
+        // when
+        BluefloodCounterRollup rollup = new BluefloodCounterRollup();
+        // then
+        assertNull(rollup.getCount());
+        assertEquals(0.0d, rollup.getRate(), 0.000001d);
+        assertEquals(0, rollup.getSampleCount());
     }
 }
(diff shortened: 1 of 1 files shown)


izrik:
@ChandraAddala @VinnyQ @usnavi @shintasmith Please review. In particular, I'm concerned about the description of the need for clarification of the rate calculation mentioned at [line 224](https://github.com/rackerlabs/blueflood/pull/682/files#diff-93a8723c3c19d4ca1cf7ed02864e9177R224). Would this go better somewhere else? Maybe in the `BluefloodCounterRollup` code or javadoc? Maybe in a separate github issue or JIRA story? Or just leave it where it is?

blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/BluefloodCounterRollupTest.java
```diff
@@ -1,93 +1,758 @@
...
+        assertFalse(a.equals(b));
+    }
+
+    @Test
+    public void equalsSameRateReturnsFalse() {
+        //given
+        BluefloodCounterRollup a = new BluefloodCounterRollup().withCount(1).withRate(1.0d);
+        BluefloodCounterRollup b = new BluefloodCounterRollup().withCount(1).withRate(1.0d);
+        //when
+        assertTrue(a.equals(b));
```
VinnyQ: should this be assertFalse?  Why are these two equals?

blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/BluefloodCounterRollupTest.java
```diff
@@ -1,93 +1,758 @@
...
+        Points<SimpleNumber> input = new Points<SimpleNumber>();
+        input.add(new Points.Point<SimpleNumber>(1234L, new SimpleNumber(3L)));
+        // when
+        BluefloodCounterRollup rollup = BluefloodCounterRollup.buildRollupFromRawSamples(input);
+        // then
+        Number count = rollup.getCount();
+        assertNotNull(count);
+        assertThat(count, is(CoreMatchers.<Number>instanceOf(Long.class)));
+        assertEquals(3L, count.longValue());
+        assertEquals(Double.POSITIVE_INFINITY, rollup.getRate(), 0.00001d); //(3)/(1-1)=inf
```
VinnyQ: Wait what?  You can test for infinity?  Why is getRate() returning infinite here?

blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/BluefloodCounterRollupTest.java
```diff
@@ -1,93 +1,758 @@
...
+        input.add(new Points.Point<SimpleNumber>(1236L, new SimpleNumber(5L)));
+        // when
+        BluefloodCounterRollup rollup = BluefloodCounterRollup.buildRollupFromRawSamples(input);
+        // then
+        Number count = rollup.getCount();
+        assertNotNull(count);
+        assertThat(count, is(CoreMatchers.<Number>instanceOf(Long.class)));
+        assertEquals(12L, count.longValue());
+        assertEquals(6.0d, rollup.getRate(), 0.00001d); // (3+4+5)/(3-1)=6
+        assertEquals(3, rollup.getSampleCount());
```
VinnyQ: Hi @izrik, you're right in that this is a tricky problem.  Personally, for rate, I think it's should measure values over time and average on number of points, not interval.

e.g.
if you start walking at 1am and travel 3 miles the first hour, 4 miles the 2nd hour, and 5 miles the 3rd hour, by 4am you traveled a total of 12 miles.  Your "rate" (or average velocity) per hour would be the total of all the miles walked: 12 miles (3 + 4 + 5), in the interval of 3 hours (4 - 1), so the rate would = 12 / 3 = 4.  

But the current code is actually treating rate as interva, because the three data points are collected at the 2am, 3am. and 4am mark (4 - 2 = 2).

For rate, we have to assume that the first data point is achieved by an interval occurring a second before the first point, which is the same as the number of data points we received at the regular interval (i.e. input.getPoints().size()).

So the code should look like this instead (quoted here with added comments by me):

```
public static BluefloodCounterRollup buildRollupFromRawSamples(Points<SimpleNumber> input) throws IOException {
        long minTime = Long.MAX_VALUE;
        long maxTime = Long.MIN_VALUE;
        BluefloodCounterRollup rollup = new BluefloodCounterRollup();
        Number count = 0L;
        for (Points.Point<SimpleNumber> point : input.getPoints().values()) {
            // count = sum of point values (3 + 4 + 5 = 12)
            count = sum(count, point.getData().getValue());

            // minTime = 1234
            minTime = Math.min(minTime, point.getTimestamp());

            // maxTime = 1236
            maxTime = Math.max(maxTime, point.getTimestamp());
        }

        // the problem is here.  Instead of using numSeconds of maxTime minus minTime (1236 - 1234 = 2)
        // it should actually be dividing by number of points from input.getPoints.size();
        // XXX double numSeconds = (double)(maxTime - minTime);
        // XXX double rate = count.doubleValue() / numSeconds;
        double rate = count.doubleValue() / input.getPoints().size();
        return rollup.withCount(count).withRate(rate).withSampleCount(input.getPoints().size());
    }
```


Can you step through

blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/BluefloodCounterRollupTest.java
izrik: @VinnyQ They both have the same count and rate (and sample count, because the field stays at its initial value). Therefore they represent the same value, even though they're not the same object.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/BluefloodCounterRollupTest.java
```diff
@@ -1,93 +1,758 @@
...
+        assertEquals(4, rollup.getSampleCount());
+    }
+
+    @Test(expected = NullPointerException.class)
+    public void rawSampleBuilderWithNullInputsThrowsException() throws IOException {
+
+        // when
+        BluefloodCounterRollup rollup = BluefloodCounterRollup.buildRollupFromRawSamples(null);
+        // then
+        // the exception is thrown
```
VinnyQ: NIT, i would combine the above into one line of comment.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/BluefloodCounterRollupTest.java
VinnyQ: Ah ok, then you should rename the test name to equalsSameRateReturnsTrue

blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/BluefloodCounterRollupTest.java
izrik: @VinnyQ Yes, you can test for infinity, but I'm not sure if this is the best way to do so. Floating point arithmetic is notoriously tricky, with plenty of special cases and rules. Now that I think about it, this should be replaced with `assertTrue(Double.isInfinite(rollup.getRate())`. `getRate()` is returning infinity because that's what was set as the rate by `buildRollupFromRawSamples`. `buildRollupFromRawSamples` divides a non-zero number by zero, the result of which is infinite, under floating point math rules.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/BluefloodCounterRollupTest.java
izrik: @VinnyQ Ah, I see what you mean now. Yes, the name is misleading. Error on my part.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/BluefloodCounterRollupTest.java
VinnyQ: oh nm, it's returning a division by 0.  Weird that it didn't throw an error, it should.  Single data point should return a rate the same as its values, not INFINITY.  But anyway this problem will be fixed (never happen) if we change the code to use number of points to calculate rate, instead of maxTime - minTime.  See comments below for my thoughts on that.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/BluefloodCounterRollupTest.java
izrik: @VinnyQ You can't use `input.getPoints().size()` to calculate the rate, because the samples aren't always uniformly spaced. There could be gaps, for example. In that case, the number of samples would not match the total amount of transpired time.

Your example of walking is fine, but doesn't entirely fit for our purposes. That example treats time as a continuous variable. However, we're dealing with time as a discrete variable (see [this](https://en.wikipedia.org/wiki/Discrete_time_and_continuous_time)).
In continuous time, we could treat the start and end of the walk as infinitesimal points in time. In discrete time, on the other hand, we can only consider whole time units (seconds, in this case). The sample represents a value not at a given moment, but across the whole time unit, and our math needs to be aware of that.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/BluefloodCounterRollupTest.java
izrik: @VinnyQ Fixed.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/BluefloodCounterRollupTest.java
izrik: Eh, it doesn't bother me so much. My use of given/when/then/expect is a holdover from using spock. The labels made the tests more readable. They are one of my favorite features of the framework.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/BluefloodCounterRollupTest.java
VinnyQ: Ok rethinking this you're right, there's an alternate way of looking at rate that the current implementation is correct for.

If we were to capture the speed of walking at time 1, 2, and 3, then the average speed for the entire interval is val(1) + val(2) + val(3) / 3 - 1.  So the rate here means the average speed over the interval, which is possibly how the original devs want to treat "rate" as for metrics.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/BluefloodCounterRollupTest.java
VinnyQ: Maybe have a special case in buildRollupFromRawSamples() to calculate rate so that if input.getPoints.size() = 1 { rate = count; } ?  instead of returning division by 0.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/BluefloodCounterRollupTest.java
VinnyQ: Ok rethinking this you're right, there's an alternate way of looking at rate that the current implementation is correct for.

If we are looking at diskusage growth or shrinkage over time:
at 1am it's at 3% full
at 2am it's at 4% full
at 3am it's at 5% full
the rate of disk usage growing is 6% per hour, from 1am to 3am.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/BluefloodCounterRollupTest.java
izrik: @VinnyQ Integer division would throw an `ArithmeticException` for divide by zero. However, floating point math allows for division by zero, and gives an explicit value of `+Infinity`. See [here](http://grouper.ieee.org/groups/754/faq.html#exceptions).

VinnyQ:
👍

blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/BluefloodCounterRollupTest.java
izrik: @VinnyQ I don't think that's correct. It's more like an average. If we want to determine change over time, we can calculate `(end value - start value) / (end time - start time)`, which would give us `(5-3)/(3-1)`, which is 1 percentage point per hour. That makes sense to me for a variable named "`rate`", given the data. However, rollups calculate things like averages, not rates of change. That's what we return from the query service; that's what we see on our grafana boards. The average of those three samples is `(3+4+5)/3=4`, which seems right to me for an average.

The numerator is what we'd expect of an average. The denominator is what we'd expect of a delta. It's unclear to me what it's supposed to be. =/

I just realized: I think what it's trying to calculate, specifically, is "average value per time unit", rather than "average of all samples". But if that's the case, then there's an additional problem in that missing samples (i.e. time units for which there is no value recorded) are effectively treated as zero, bringing the whole average down.
If our time unit is one hour

izrik:
@shintasmith @ChandraAddala @usnavi Would y'all care to weigh in?

blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/BluefloodCounterRollupTest.java
```diff
@@ -1,93 +1,761 @@
...
+        //    ------+-----------------------+-----------------------+-----------------------+-----
+        //          1                       1                       1                       1
+        //          2                       2                       2                       2
+        //          3                       3                       3                       3
+        //          4                       5                       6                       7
+        //          L                       L                       L                       L
+        //          |                                                                       |
+        //          | <------------------------- three time units ------------------------> |
+        //                                         (3+4+5)/3=4
+        //
```
shintasmith: :+1: great explanation!

blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/BluefloodCounterRollupTest.java
shintasmith: I actually think counter is a measurement of something at a single point of time, as supposed to over a period of time. So at this point, I'm leaning towards the first (i.e: current) calculation.

izrik: @shintasmith Fair enough. I've created a story to sort it all out. The tests in this PR just document existing behavior. Is that sufficient for now?

shintasmith: I am actually saying there might not be a need to create a story :-) Can you maybe add this topic to the Dev meeting? We should probably discuss in a team setting.

So yes, this PR is sufficient. :-) :+1:

izrik: @shintasmith We still need the story, because none of it is documented. It wasn't even tested very well.

PR #683, opened 2016-04-22
Test BluefloodTimerRollup.Percentile

izrik:
This PR adds tests for the `BluefloodTimerRollup.Percentile` class. It brings the class-specific unit test coverage from zero up to 100% of both lines and branches.

### DIFF
--- a/blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/BluefloodTimerRollupPercentileTest.java
+++ b/blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/BluefloodTimerRollupPercentileTest.java
@@ -117,4 +117,28 @@ public class BluefloodTimerRollupPercentileTest {
         assertEquals(BigDecimal.class, result.getClass());
         assertEquals(value, result);
     }
+
+    @Test
+    public void constructorSetsMeanAndGetMeanReturnsMean() {
+        // when
+        BluefloodTimerRollup.Percentile value = new BluefloodTimerRollup.Percentile((Number)123L);
+        // then
+        assertEquals(123L, value.getMean());
+    }
+
+    @Test
+    public void longConstructorSetsMean() {
+        // when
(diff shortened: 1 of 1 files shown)


izrik:
@usnavi @shintasmith @VinnyQ @ChandraAddala Please review.

usnavi:
lgtm

blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/BluefloodTimerRollupPercentileTest.java
```diff
@@ -0,0 +1,220 @@
...
+    @Test
+    public void maybePromoteLongYieldsSameValue() {
+        // given
+        Long value = 123L;
+        // when
+        Number result = BluefloodTimerRollup.Percentile.maybePromote(value);
+        // then
+        assertNotNull(result);
+        assertEquals(Long.class, result.getClass());
+        assertEquals(123L, (long)(Long)result);
```
shintasmith: why not do ```result.longValue()``` ?

izrik: @shintasmith No particular reason. In the first two tests, I had used a simple cast, so I think I was just sticking with that. I'm not entirely sure what goes on under the hood for casts or for `longValue()`, at least not in Java. Would one be better, in particular to another?

shintasmith: Just for better readability. The first time I read that code, it wasn't obvious to me why there are 2 casts. Anyways, it's a minor comment. I am ok with either way.

izrik: @shintasmith According to [this](http://stackoverflow.com/questions/36869186/what-is-the-difference-between-casting-to-a-long-and-callint-longvalue), they are equivalent.

PR #685, opened 2016-04-25
add datastax connection metrics to our existing metrics registry

shintasmith:
**WHAT**
Datastax connection has its own collection of metrics, as documented here:
http://docs.datastax.com/en/drivers/java/3.0/com/datastax/driver/core/Metrics.html

This PR is to grab all of those metrics and register them with our Blueflood MetricsRegistry, so we have metrics related to datastax in our graphite. 

I started Blueflood with Datastax configured. Then from JConsole, I was able to see all the datastax.* metrics under our metrics.

### DIFF
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxIO.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxIO.java
@@ -15,9 +15,11 @@
  */
 package com.rackspacecloud.blueflood.io.datastax;
 
+import com.codahale.metrics.*;
 import com.datastax.driver.core.*;
 import com.datastax.driver.core.exceptions.NoHostAvailableException;
 import com.datastax.driver.core.policies.DCAwareRoundRobinPolicy;
+
 import com.datastax.driver.core.policies.TokenAwarePolicy;
 import com.rackspacecloud.blueflood.io.CassandraModel;
 import com.rackspacecloud.blueflood.io.IOConfig;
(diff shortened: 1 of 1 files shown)


shintasmith:

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxIO.java
```diff
@@ -113,34 +106,58 @@ private static PoolingOptions getPoolingOptions(int numHosts){
...
+            public void onTimerAdded(String name, Timer timer) {
+                bfMetricsRegistry.register(metricsPrefix + name, timer);
+            }
+
+            @Override
+            public void onTimerRemoved(String name) {
+                bfMetricsRegistry.remove(metricsPrefix + name);
             }
-        }, 1, 1, TimeUnit.MINUTES);
+        });
```
ChandraAddala: I am having hard time understand the relation between code being removed and the new code added. Can you explain it a little bit?

ChandraAddala:
Looks good.

usnavi:
lgtm

izrik:
@shintasmith Would we want to specify separate prefixes for ingest vs. query vs. rollup? They all use datastax, right?

shintasmith:
@izrik, good point, but I am not sure we separate our metrics that way today with Astyanax. All of them just go to the same bucket.

izrik:
@shintasmith Fair enough. I just wanted to bring that up, in case it might become an issue later. Other than that, it looks good to me.

PR #686, opened 2016-04-28
Test Granularity

izrik:
This PR adds tests for the `Granularity` class. It brings the class-specific unit test coverage from 67% up to 98% of lines, and from 65% to 98% of branches.


izrik:
@shintasmith @VinnyQ @ChandraAddala @usnavi Please review. There are some method in this class that I'm still not entirely sure what they do. The existing tests were complex and probably brittle, but I left them in in case we need to decipher what some method is doing in the future. On the whole, the existing tests were sub-par in my opinion, in that they are complicated and contain lots of calculations and logic. In the future, it would be good to get rid of them completely.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/rollup/Granularity.java
```diff
@@ -171,29 +171,55 @@ public Range deriveRange(int slot, long referenceMillis) {
...
+     *                  {@code "GEOMETRIC"}.
+     * @param assumedIntervalMillis FULL resolution is tricky because we don't
+     *                              know the period of check in question.
+     *                              Assume the minimum period and go from there.
+     * @param checkTtl check the specified time against the default ttl values,
+     *                 skipping granularities that don't fit.
      * @return
      */
-    public static Granularity granularityFromPointsInInterval(String tenantid, long from, long to, int points) {
+    public static Granularity granularityFromPointsInInterval(String tenantid, long from, long to, int points, String algorithm, long assumedIntervalMillis, boolean checkTtl) {
```
shintasmith: I am not sure I understand why you added the ```checkTtl``` here. Can you elaborate?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/rollup/Granularity.java
```diff
@@ -254,23 +280,25 @@ private static Granularity granularityFromPointsLinear(double requestedDuration,
...
      * @param requestedDuration (milliseconds)
      */
-    private static Granularity granularityFromPointsGeometric(String tenantid, long from, long to, double requestedDuration, int requestedPoints) {
+    private static Granularity granularityFromPointsGeometric(String tenantid, long from, long to, double requestedDuration, int requestedPoints, long assumedIntervalMillis, boolean checkTtl) {
         double minimumPositivePointRatio = Double.MAX_VALUE;
         Granularity gran = null;
-        if (SAFETY_TTL_PROVIDER == null) {
+        if (checkTtl && SAFETY_TTL_PROVIDER == null) {
             SAFETY_TTL_PROVIDER = SafetyTtlProvider.getInstance();
         }
```
ChandraAddala: These code changes are not mere refactoring. They are altering the functionality. What story are these changes part of?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/rollup/Granularity.java
```diff
@@ -171,29 +171,55 @@ public Range deriveRange(int slot, long referenceMillis) {
...
      * @param from beginning of interval (millis)
      * @param to end of interval (millis)
      * @param points count of desired data points
+     * @param algorithm the algorithm to use. Valid values are
+     *                  {@code "GEOMETRIC"}, {@code "LINEAR"}, and
+     *                  {@code "LESSTHANEQUAL"}. Any other value is treated as
+     *                  {@code "GEOMETRIC"}.
+     * @param assumedIntervalMillis FULL resolution is tricky because we don't
+     *                              know the period of check in question.
+     *                              Assume the minimum period and go from there.
```
usnavi: what do you mean by "period of check in question"?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/rollup/Granularity.java
izrik: @shintasmith 
The `granularityFromPointsGeometric` method uses the `SafetyTtlProvider.getFinalTTL` method to (I think) get some kind of TTL value. It uses this TTL value to (I think) automatically skip granularities that wouldn't work for the specified `from` time. But that means that the behavior changes based on what the current time is. And that, in turn, means that the tests will fail unless they provide input data based on the current time. The tests I was writing at the time all failed because the method returned `Granularity.MIN_1440`, which didn't make any sense given the inputs. A return value of null or an exception would have been more appropriate. Also, because it was skipping granularities, there were large sections of the method that weren't being executed. It should be noted that this TTL check only happens when the `GEOMETRIC` algorithm is selected. That is, the behavior of the other two algorithms is _different_ in that regard.

I've run into this sort of thing in the past, and it's never made things nice or easy. I've had tests (on other a different project) fail because my machine clock was not properly sync'd to the server's, and there was no way to really know that was the cause. Time-sensitive testing is a losing proposition.

So I decided to provide an overload that allows me to turn off the TTL-checking behavior of the function. That way, I can provide fixed inputs to the method, and confirm that at least the math behind the method is correct. And I don't have to worry about what time it is when I run the tests.

IIRC, the `SafetyTtlProvider` class was also relying on functionality within the `Granularity` class. So trying to test one specific part of `Granularity` was indirectly involving larger chunks of  the class.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/rollup/Granularity.java
izrik: @ChandraAddala This is intended to be related to the testing. Please, see above. The overload with the `checkTtl` parameter is only called by the tests. The changes only minimally change the functionality, and only when the tests are running. Would it be better if I added a `@VisibleForTesting` annotation? Failing that, I could create a separate story to add the overload.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/rollup/Granularity.java
izrik: @usnavi I copied that from `granularityFromPointsGeometric`, on line 299. I don't know what the original author meant. I'm still not sure what that parameter is for.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/rollup/Granularity.java
ChandraAddala: My concern is not with the new method added, it is with the changes to the existing method. I am looking at the old code and the changed code. The code changes does make the existing method little more complex that it was before and I dont see a compelling reason to do so. I think you should be able to test this without making these changes. If ``Calendar.getInstance()`` is causing the problem, change it with ``Clock`` instance which you can mock.

shintasmith: I am leaning towards agreeing with @ChandraAddala . I think all 3 of us agree that the existing implementation will need to be changed slightly to accommodate testing. The challenge is how do we change the existing implementation so that we leave the behavior/logic of the class as in-tact as possible.

Using the ```Clock``` seems reasonable. @ChandraAddala had refactored ```ScheduledContext``` this way. He added a constructor that takes an instance of Clock:
```
public ScheduleContext(long currentTimeMillis, Collection<Integer> managedShards)
public ScheduleContext(long currentTimeMillis, Collection<Integer> managedShards, Clock clock)
```
Our normal services code will use the first constructor, which will use a default ```Clock``` implementation. Our test code can mock ```Clock``` and can use the second constructor. This leaves the remaining logic of the code in-tact.

Adding ```checkTtl``` parameter to ```granularityFromPointsInInterval()``` changes the logic of the method. It opens up a different behavior of the method, which is to skip checking TTL. I don't think that's good.

izrik: @ChandraAddala @shintasmith I think the `Clock` option is the best. I'll do that.

izrik:
@shintasmith @ChandraAddala @usnavi Code updated. Is it better with the `Clock` paramter?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/utils/CalendarClock.java
```diff
@@ -0,0 +1,21 @@
+package com.rackspacecloud.blueflood.utils;
+
+import org.joda.time.Instant;
+
+import java.util.Calendar;
+
+public class CalendarClock implements Clock {
```
ChandraAddala: Cant we reuse the existing one? ``DefaultClockImpl``?

izrik: `DefaultClockImpl` returns `Instant.now()`, which returns an `Instant` initialized from other sources within jodatime. I think that `Instant` by default ultimately gets its values from `System.currentTimeMillis()`, but I'm not entirely sure.

The code that I replaced, however, used `Calendar.getInstance().getTimeInMillis()`. I dug around inside `Calendar` for a while, but I couldn't find where it was getting its values from by default.

I can't be sure that `DefaultClockImpl` would provide exactly the same behavior in all cases, especially considering that the two time sources can be configured separately. For example, I'm not sure if the same time zone is used in both cases. Thus, in order to ensure that the refactoring is the same, I decided to create another clock implementation. That way, if the two time libraries initialize differently, or if some other distant part of blueflood changes time configurations somewhere somehow, it will still behave exactly as before.

shintasmith: I actually don't know why we call ```Calendar.getInstance().getTimeInMillis()```. That was the first question I asked when I saw that code.

shintasmith: http://stackoverflow.com/questions/368094/system-currenttimemillis-vs-new-date-vs-calendar-getinstance-gettime

ChandraAddala: Both return the current time in millis which is just a number of millis since epoch. I think we should be fine. 

https://docs.oracle.com/javase/7/docs/api/java/util/Calendar.html#getTimeInMillis()
http://joda-time.sourceforge.net/apidocs/org/joda/time/DateTimeUtils.html#currentTimeMillis()

izrik: @ChandraAddala @shintasmith I agree that it's probably close enough. My only concern is that they can be configured differently:

1. `new Instant()` calls `DateTimeUtils.currentTimeMillis()` which calls `MillisProvider.getMillis()` on the `cMillisProvider` field. `cMillisProvider` in turn can be changed by any piece of code that calls the various setter methods, chiefly `DateTimeUtils.setCurrentMillisProvider`.
2. `Calendar.getTimeInMillis()` calls `updateTime()` which calls `computeTime` on the class that extends `Calendar`. The `Calendar.getInstance()` call gets a `Calendar` instance via a big complex process, which can be customize... _somewhere_. I don't know where; trying to find it exhausted my patience.

Anyways, it is possible that some part of the codebase, somewhere, would try to change the way time values are calculated via either route. That would affect how the system operates at run-time, and may break something. I think that is very unlikely, however. **If y'all are ok with that, then it's fine with me; I'll change it.**

ChandraAddala: We currently use the ``DefaultClock`` in ``ScheduleContext`` and ``ShardStateManager``. I dont see any reason why someone would mess with joda-time to configure a different ``millisProvider``. But if they do, they better evaluate the impact properly.

I think its pretty safe to use DefaultClockImpl.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/rollup/Granularity.java
usnavi: Can we make a comment that this was from a previous author and we are not sure what it means at this time.  I was staring at this forever thinking i was missing something.

usnavi:
other then my one comment, lgtm.

VinnyQ:
👍  LGREATTM!  look at that coverage climb!  GJ @izrik

blueflood-core/src/main/java/com/rackspacecloud/blueflood/rollup/Granularity.java
izrik: @usnavi I added a note about that. How does it look?

usnavi: thanks!  lgtm

PR #687, opened 2016-04-28
Test maxvalue

izrik:
This PR adds tests for the `MaxValue` classes. It brings the class-specific unit test coverage from 72% up to 100% of lines, and from 78% to 100% of branches.


izrik:

blueflood-core/src/main/java/com/rackspacecloud/blueflood/types/MaxValue.java
```diff
@@ -100,17 +100,17 @@ void handleRollupMetric(IBasicRollup basicRollup) throws RuntimeException {
         }
 
         if (this.isFloatingPoint() && !other.isFloatingPoint()) {
-            if (this.toDouble() < (double)other.toDouble()) {
+            if (this.toDouble() < (double)other.toLong()) {
```
shintasmith: Why is this changed?

blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/MaxValueTest.java
```diff
@@ -32,97 +39,375 @@ public void setUp() {
...
-    public void testMaxValueForDoubleMetrics() throws IOException {
-        for (double val : TestData.DOUBLE_SRC) {
-            max.handleFullResMetric(val);
-        }
-        Assert.assertTrue(max.isFloatingPoint());
-        Assert.assertEquals(90.48232472545334, max.toDouble(), 0);
+    public void fullResInitialDoubleSetsValue() {
+
+        // when
+        max.handleFullResMetric(123.45d);
```
shintasmith: I know this is an existing test. But I'm wondering why we use class attribute and not just declaring MaxValue inside each test method? Having it in the class level makes parallelizing test a little difficult, no? Should we make this better?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/types/MaxValue.java
izrik: `MaxValue` can store a floating-point (`Double`) or a non-floating-point (`Long`) values. It does this by storing each in a separate field. If you call `toDouble()`, it returns the `double` field. If you call `toLong()`, it returns the `long` field. It is the job of the caller to determine which should be used, by calling `isFloatingPoint()`. [I don't like this system. That's just how the code behaves right now.]
In this particular case, it first checks that `this` holds a floating point number and that `other` holds a non-floating-point number. If that's the case, it should compare the respective `double` and `long` values. However, if `other` is not a floating-point number, and its `double` field has never been set, calling `toDouble()` here will return `0.0`, instead of the non-floating-point value which _was_ set.
In summary: `MaxValue` requires a two-step process for getting the correct stored value, and the original developers called the wrong method in step 2. Compare this to `MinValue`, which already used the correct `toLong()` at this point: https://github.com/izrik/blueflood/blob/test-maxvalue/blueflood-core/src/main/java/com/rackspacecloud/blueflood/types/MinValue.java#L103

blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/MaxValueTest.java
izrik: @shintasmith I'm not sure what you're referring to as "class attribute". Do you mean the `TestData.DOUBLE_SRC`?

shintasmith: I mean ```max``` is a class attribute, as supposed to line 204, where you new a MaxValue object inside the test method.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/types/MaxValue.java
shintasmith: Ah, gotcha, I neglected to see the first if statement:
```
if (this.isFloatingPoint() && !other.isFloatingPoint()) {
```
It makes sense now.

No need to revert. Fixing it was the right thing to do. I just didn't see it at first :-)

I agree about creating a new story to look into this later.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/MaxValueTest.java
izrik: @shintasmith The `max` field is initialized in the `setUp()` method. Initializing it in the separate test methods instead would add a line for each test. Using a local variable for each test instead of a field would save one line total. Functionally, the tests would work exactly the same; there would be no difference. The two approaches are equivalent, and more a matter of taste. I can change it if you prefer.

The `other` variable, on the other hand, is only used in two test methods. It wouldn't really gain anything to have it as a field. Of course, that is possible to do, and would be equivalent, as above. Again, it's a matter of taste. I can change it if you think that would look better or be more readable.

shintasmith: I think initializing in each test method is a good thing, right? It allows us to parallelize at test method level, if we ever want to.

But... since it was like that from the beginning, and maybe we won't ever parallelize our tests at method level, I am ok with leaving it as is.

izrik: The `max` field is an instance field, not a static one. It will not interfere with any attempt to parallelize the tests. we could run the tests in this file in parallel _now_ and they would all pass.

shintasmith: I meant, running tests parallel at method level, not at class level. But its ok.

ChandraAddala:
Looks good.

shintasmith:
lgtm too

blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/MaxValueTest.java
izrik: Even running parallel at method level, each test method gets its own instance of `MaxValueTest`, with its own `max` field. It won't interfere with parallelization at all. I'll look for a reference.

izrik: @shintasmith http://junit.org/junit4/faq.html#atests_2

shintasmith: Ah gotcha. No, it's ok. I understand now. It's really not worth spending anymore time. :-)

PR #689, opened 2016-05-02
modify multi ingestion handler to return 207 multi-status

VinnyQ:
- WHAT
When handling mutliple metrics ingestion, return 207 (MULTI_STATUS) when partial errors are encountered; i.e. when some of the metrics failed validation.  Valid metrics should still be posted.  For now the response should be empty if 207 is returned.

Still return 200 for 100% valid payload and 400 for 100% invalid payload (along with validation error messages).

- HOW
Modify HttpMetricsIngestionHandler and JSONMetricsContainer to test for partial validation failures and return 207 after persisting the valid metrics.

- TEST
To test manually, modify the timestamp in the following post examples and POST to a blueflood ingest node.  Use http://www.epochconverter.com/ to obtain timestamp in seconds (and then multiple it by 1000 to convert it to ms for use in the payload).

```
# single metric
curl -i -X POST 'http://localhost:19000/v2.0/836986/ingest' -d \
  '[
      {
        "collectionTime": 1461873511000,
        "ttlInSeconds": 172800,
        "metricValue": 1,
        "metricName": "p3.metric.one"
      }
    ]'

# multi metrics
curl -i -X POST 'http://localhost:19000/v2.0/836986/ingest/multi' -d \
  '[
      {
        "tenantId": 836986,
        "collectionTime": 1461873511000,
        "ttlInSeconds": 172800,
        "metricValue": 2,
        "metricName": "p3.metric.two"
      },
      {
        "tenantId": 836986,
        "collectionTime": 1461873511000,
        "ttlInSeconds": 172800,
        "metricValue": 3,
        "metricName": "p3.metric.three"
      },
      {
        "tenantId": 836986,
        "collectionTime": 1461873511000,
        "ttlInSeconds": 172800,
        "metricValue": 4,
        "metricName": "p3.metric.four"
      }
    ]'

# aggregated single
curl -i -X POST 'http://localhost:19000/v2.0/836986/ingest/aggregated' -d \
  '{
    "tenantId":"836986",
    "timestamp":1462312693000,
    "flushInterval": 15000,
    "counters":[
        {
            "name":"internal.bad_lines_seen",
            "value":0,
            "rate":0
        },{
            "name":"internal.packets_received",
            "value":317,
            "rate":21.133333333333333
        },{
            "name":"new.C5s",
            "value":16,
            "rate":1
        }
    ],
    "gauges":[
        {
            "name":"836986.G200ms",
            "value":145
        },{
            "name":"836986.G1s",
            "value":397
        },{
            "name":"836986.G10s",
            "value":56
        },{
            "name":"internal.timestamp_lag",
            "value":0
        }
    ]
  }'

# aggregated multi
curl -i -X POST 'http://localhost:19000/v2.0/836986/ingest/aggregated/multi' -d \
  '[
  {
    "tenantId":"836986",
    "timestamp":14623126930,
    "flushInterval": 15000,
    "counters":[
        {
            "name":"ap1.counter.one",
            "value":1,
            "rate":0
        },{
            "name":"ap1.counter.two",
            "value":2,
            "rate":21.133333333333333
        },{
            "name":"ap1.counter.three",
            "value":3,
            "rate":1
        }
    ],
    "gauges":[
        {
            "name":"ap1.gauge.one",
            "value":100
        },{
            "name":"ap1.gauge.two",
            "value":200
        },{
            "name":"ap1.gauge.three",
            "value":300
        }
    ]
  },
  {
    "tenantId":"836986",
    "timestamp":1462312693000,
    "flushInterval": 15000,
    "counters":[
        {
            "name":"ap2.counter.one",
            "value":1,
            "rate":0
        },{
            "name":"ap2.counter.two",
            "value":2,
            "rate":21.133333333333333
        },{
            "name":"ap2.counter.three",
            "value":3,
            "rate":1
        }
    ],
    "gauges":[
        {
            "name":"ap2.gauge.one",
            "value":100
        },{
            "name":"ap2.gauge.two",
            "value":200
        },{
            "name":"ap2.gauge.three",
            "value":300
        }
    ]
  },
  {
    "tenantId":"836986",
    "timestamp":146231269300000,
    "flushInterval": 15000,
    "counters":[
        {
            "name":"ap3.counter.one",
            "value":1,
            "rate":0
        },{
            "name":"ap3.counter.two",
            "value":2,
            "rate":21.133333333333333
        },{
            "name":"ap3.counter.three",
            "value":3,
            "rate":1
        }
    ],
    "gauges":[
        {
            "name":"ap3.gauge.one",
            "value":100
        },{
            "name":"ap3.gauge.two",
            "value":200
        },{
            "name":"ap3.gauge.three",
            "value":300
        }
    ]
  }
  ]'
```

### DIFF
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/formats/AggregatedPayload.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/formats/AggregatedPayload.java
@@ -14,25 +14,24 @@
  *    limitations under the License.
  */
 
-package com.rackspacecloud.blueflood.inputs.handlers.wrappers;
+package com.rackspacecloud.blueflood.inputs.formats;
 
 import com.rackspacecloud.blueflood.service.Configuration;
 import com.rackspacecloud.blueflood.service.CoreConfig;
 import com.rackspacecloud.blueflood.types.*;
+import org.codehaus.jackson.annotate.JsonIgnore;
 
-import java.util.AbstractList;
-import java.util.Collection;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 
 // Using nested classes for now. Expect this to be cleaned up.
 public class AggregatedPayload {
-    private static final long pastDiff = Configuration.getInstance().getLongProperty( CoreConfig.BEFORE_CURRENT_COLLECTIONTIME_MS );
-    private static final long futureDiff = Configuration.getInstance().getLongProperty( CoreConfig.AFTER_CURRENT_COLLECTIONTIME_MS );
+
+    private final long BEFORE_CURRENT_COLLECTIONTIME_MS = Configuration.getInstance().getLongProperty( CoreConfig.BEFORE_CURRENT_COLLECTIONTIME_MS );
+    private final long AFTER_CURRENT_COLLECTIONTIME_MS = Configuration.getInstance().getLongProperty( CoreConfig.AFTER_CURRENT_COLLECTIONTIME_MS );
 
     private String tenantId;
     private long timestamp; // millis since epoch.
-    
+
     // this field is optional
     private long flushInterval = 0;
     
@@ -61,16 +60,17 @@ public class AggregatedPayload {
     public Collection<BluefloodSet> getSets() { return safeAsList(sets); }
     public Collection<BluefloodEnum> getEnums() { return safeAsList(enums); }
 
-
     public List<String> getValidationErrors() {
-
         List<String> errors = new java.util.ArrayList<String>();
-        long current = System.currentTimeMillis();
 
-        if( timestamp > current + futureDiff )
-            errors.add( "'timestamp' '" + timestamp + "' is more than '" + futureDiff + "' milliseconds into the future." );
-        else if( timestamp < current - pastDiff )
-            errors.add( "'timestamp' '" + timestamp + "' is more than '" + pastDiff + "' milliseconds into the past." );
+        long currentTime = System.currentTimeMillis();
+        if ( timestamp > currentTime + AFTER_CURRENT_COLLECTIONTIME_MS ) {
+            // timestamp is too far in the future
+            errors.add( "'timestamp' '" + timestamp + "' is more than '" + AFTER_CURRENT_COLLECTIONTIME_MS + "' milliseconds into the future." );
+        } else if ( timestamp < currentTime - BEFORE_CURRENT_COLLECTIONTIME_MS ) {
+            // timestamp is too far in the past
+            errors.add( "'timestamp' '" + timestamp + "' is more than '" + BEFORE_CURRENT_COLLECTIONTIME_MS + "' milliseconds into the past." );
+        }
 
         return errors;
     }
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/formats/JSONMetric.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/formats/JSONMetric.java
@@ -1,0 +1,90 @@
+/*
+ * Copyright 2013 Rackspace
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+package com.rackspacecloud.blueflood.inputs.formats;
+
+import com.rackspacecloud.blueflood.service.Configuration;
+import com.rackspacecloud.blueflood.service.CoreConfig;
+
+import java.util.List;
+
+// Jackson compatible class. Jackson uses reflection to call these methods and so they have to match JSON keys.
+public class JSONMetric {
+
+    final long BEFORE_CURRENT_COLLECTIONTIME_MS = Configuration.getInstance().getLongProperty( CoreConfig.BEFORE_CURRENT_COLLECTIONTIME_MS );
+    final long AFTER_CURRENT_COLLECTIONTIME_MS = Configuration.getInstance().getLongProperty( CoreConfig.AFTER_CURRENT_COLLECTIONTIME_MS );
+
+    private String metricName;
+    private Object metricValue;
+    private long collectionTime;
+    private int ttlInSeconds;
+    private String unit;
+
+    public String getMetricName() {
+        return metricName;
+    }
+
+    public void setMetricName(String metricName) {
+        this.metricName = metricName;
+    }
+
+    public String getUnit() {
+        return unit;
+    }
+
+    public void setUnit(String unit) {
+        this.unit = unit;
+    }
+
+    public Object getMetricValue() {
+        return metricValue;
+    }
+
+    public void setMetricValue(Object metricValue) {
+        this.metricValue = metricValue;
+    }
+
+    public long getCollectionTime() {
+        return collectionTime;
+    }
+
+    public void setCollectionTime(long collectionTime) {
+        this.collectionTime = collectionTime;
+    }
+
+    public int getTtlInSeconds() {
+        return this.ttlInSeconds;
+    }
+
+    public void setTtlInSeconds(int ttlInSeconds) {
+        this.ttlInSeconds = ttlInSeconds;
+    }
+
+    public List<String> getValidationErrors() {
+        List<String> errors = new java.util.ArrayList<String>();
+
+        long currentTime = System.currentTimeMillis();
+        if ( collectionTime < currentTime - BEFORE_CURRENT_COLLECTIONTIME_MS ) {
+            // collectionTime is too far in the past
+            errors.add( "'" + metricName + "': 'collectionTime' '" + collectionTime + "' is more than '" + BEFORE_CURRENT_COLLECTIONTIME_MS + "' milliseconds into the past." );
+        } else if ( collectionTime > currentTime + AFTER_CURRENT_COLLECTIONTIME_MS ) {
+            // collectionTime is too far in the future
+            errors.add( "'" + metricName + "': 'collectionTime' '" + collectionTime + "' is more than '" + AFTER_CURRENT_COLLECTIONTIME_MS + "' milliseconds into the future." );
+        }
+
+        return errors;
+    }
+
+}
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/formats/JSONMetricsContainer.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/formats/JSONMetricsContainer.java
@@ -22,8 +22,6 @@ import com.rackspacecloud.blueflood.service.CoreConfig;
 import com.rackspacecloud.blueflood.types.Locator;
 import com.rackspacecloud.blueflood.types.Metric;
 import com.rackspacecloud.blueflood.utils.TimeValue;
-import org.apache.commons.lang.StringUtils;
-import org.codehaus.jackson.annotate.JsonIgnore;
 import org.joda.time.DateTime;
 
 import java.util.ArrayList;
@@ -33,68 +31,79 @@ import java.util.concurrent.TimeUnit;
 public class JSONMetricsContainer {
     private final String tenantId;
     private final List<JSONMetric> jsonMetrics;
-    private List<Metric> delayedMetrics;
 
     private static final long TRACKER_DELAYED_METRICS_MILLIS = Configuration.getInstance().getLongProperty(CoreConfig.TRACKER_DELAYED_METRICS_MILLIS);
-
     private static final long MAX_AGE_ALLOWED = Configuration.getInstance().getLongProperty(CoreConfig.ROLLUP_DELAY_MILLIS);
     private static final long SHORT_DELAY = Configuration.getInstance().getLongProperty(CoreConfig.SHORT_DELAY_METRICS_ROLLUP_DELAY_MILLIS);
 
     private static final long pastDiff = Configuration.getInstance().getLongProperty( CoreConfig.BEFORE_CURRENT_COLLECTIONTIME_MS );
     private static final long futureDiff = Configuration.getInstance().getLongProperty( CoreConfig.AFTER_CURRENT_COLLECTIONTIME_MS );
 
+    private final List<Metric> metrics = new ArrayList<Metric>();
+    private final List<Metric> delayedMetrics = new ArrayList<Metric>();
+    private final List<String> errors = new ArrayList<String>();
+
     public JSONMetricsContainer(String tenantId, List<JSONMetric> metrics) {
         this.tenantId = tenantId;
         this.jsonMetrics = metrics;
+        processJson();
     }
 
     public List<String> getValidationErrors() {
-
-        List<String> errors = new ArrayList<String>();
-
-        // Validate that any ScopedJSONMetric is actually scoped to a tenant.
-        for (JSONMetric jsonMetric : this.jsonMetrics) {
-            errors.addAll( jsonMetric.getValidationErrors() );
-        }
-
         return errors;
     }
 
-    public List<Metric> toMetrics() {
+    public List<Metric> getValidMetrics() {
+        return metrics;
+    }
+
+    private List<Metric> processJson() {
+
         if (jsonMetrics == null || jsonMetrics.isEmpty()) {
             return null;
         }
 
-        final List<Metric> metrics = new ArrayList<Metric>();
-        delayedMetrics = new ArrayList<Metric>();
         for (JSONMetric jsonMetric : jsonMetrics) {
+
+            // validate metric and retrieve error message if failed
+            List<String> metricValidationErrors = jsonMetric.getValidationErrors();
+            if ( !metricValidationErrors.isEmpty() ) {
+                // has metric has validation errors, do not convert metric and add to errors list and go to next metric
+                errors.addAll(metricValidationErrors);
+                continue;
+            }
+
+            if (jsonMetric.getMetricValue() == null) {
+                // skip null value
+                continue;
+            }
+
+            // no error, create metric from json values, but skip null metricValue
             Locator locator;
-            if (jsonMetric instanceof ScopedJSONMetric) {
-                ScopedJSONMetric scopedMetric = (ScopedJSONMetric)jsonMetric;
+            if (jsonMetric instanceof JSONMetricScoped) {
+                JSONMetricScoped scopedMetric = (JSONMetricScoped)jsonMetric;
                 locator = Locator.createLocatorFromPathComponents(scopedMetric.getTenantId(), jsonMetric.getMetricName());
             } else {
                 locator = Locator.createLocatorFromPathComponents(tenantId, jsonMetric.getMetricName());
             }
 
-            if (jsonMetric.getMetricValue() != null) {
-                final Metric metric = new Metric(locator, jsonMetric.getMetricValue(), jsonMetric.getCollectionTime(),
-                        new TimeValue(jsonMetric.getTtlInSeconds(), TimeUnit.SECONDS), jsonMetric.getUnit());
-                long delay = new DateTime().getMillis() - metric.getCollectionTime();
+            final Metric metric = new Metric(locator, jsonMetric.getMetricValue(), jsonMetric.getCollectionTime(),
+                    new TimeValue(jsonMetric.getTtlInSeconds(), TimeUnit.SECONDS), jsonMetric.getUnit());
+            long delay = new DateTime().getMillis() - metric.getCollectionTime();
 
-                if (delay > TRACKER_DELAYED_METRICS_MILLIS) {
-                    delayedMetrics.add(metric);
-                }
-
-                if (delay > MAX_AGE_ALLOWED) {
-                    if (delay <= SHORT_DELAY) {
-                        Instrumentation.markMetricsWithShortDelayReceived();
-                    } else {
-                        Instrumentation.markMetricsWithLongDelayReceived();
-                    }
-                }
-
-                metrics.add(metric);
+            if (delay > TRACKER_DELAYED_METRICS_MILLIS) {
+                delayedMetrics.add(metric);
             }
+
+            if (delay > MAX_AGE_ALLOWED) {
+                if (delay <= SHORT_DELAY) {
+                    Instrumentation.markMetricsWithShortDelayReceived();
+                } else {
+                    Instrumentation.markMetricsWithLongDelayReceived();
+                }
+            }
+
+            metrics.add(metric);
         }
 
         return metrics;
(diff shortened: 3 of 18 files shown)


blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/formats/JSONMetricsContainer.java
```diff
@@ -142,6 +147,13 @@ public void setTtlInSeconds(int ttlInSeconds) {
             this.ttlInSeconds = ttlInSeconds;
         }
 
+        public boolean isValid() {
+            long current = System.currentTimeMillis();
+
+            // valid if collectionTime is greater than past window or less than future window
+            return ((collectionTime >= current - pastDiff ) && ( collectionTime <= current + futureDiff ));
```
shintasmith: Can you somehow consolidate the isValid() and getValidationErrors() so that the check for what's valid timestamp is done only in one place? 

One way of doing that is to maybe provide isTooOld() or isWayInFuture() and use these methods in both isValid() and getValidationErrors()? That way we don't replicate the logic in 2 different places.

shintasmith: As I look at the code more, I think having isValid() and getValidationErrors() will produce inconsistent results. Because each of those methods does its own call of System.currentTimeMillis(). 

Imagine us ingesting metrics that are on the borderline of being valid, just a few milliseconds shy of being invalid. When we call the isValid() on those, we get true. But a few milliseconds later, when we call getValidationErrors(), the System.currentTimeMillis() will be different, thus may cause those metrics to show up as being invalid. 

I think you should just keep one method to do the timestamp validation check. You will have to keep the invalid metrics as a class member variable, i.e: add:
```
List<Metric> invalidMetrics;
```
that gets populated when someone calls toMetrics(). And then you just do a getInvalidMetricsCount() or some method like that. If the count is greater than zero, return 207. If the count is all metrics are invalid, return 400. etc.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/formats/JSONMetricsContainer.java
```diff
@@ -64,14 +65,18 @@ public JSONMetricsContainer(String tenantId, List<JSONMetric> metrics) {
         delayedMetrics = new ArrayList<Metric>();
         for (JSONMetric jsonMetric : jsonMetrics) {
             Locator locator;
+            boolean isValid;
```
ChandraAddala: Cant u directly do this?

```
boolean isValid = jsonMetric.isValid();
```

blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/formats/JSONMetricsContainer.java
ChandraAddala: I agree we should combine both these methods. But do we need to maintain another data structure with the data. I think we can get away from declaring another variable.

blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/inputs/handlers/HttpHandlerIntegrationTest.java
```diff
@@ -108,6 +108,34 @@ public void testHttpIngestionInvalidFutureCollectionTime() throws Exception {
...
+        jsonBody = updateTimeStampJson(jsonBody, "\"%FUTURE_TIMESTAMP%\"", futureTime);
+        jsonBody = jsonBody.replaceAll("%TENANT_ID_.%", TID);
+
+        HttpResponse response = httpPost( TID, postMultiPath, jsonBody );
+        String[] output = getBodyArray( response );
+
+        try {
+            assertEquals("Should get status 207 from ingestion server for POST with invalid collectionTime", 207, response.getStatusLine().getStatusCode() );
+            assertEquals( 4, output.length );
+            assertTrue( Pattern.matches( PAST_COLLECTION_TIME_REGEX, output[ 1 ] ) );
```
ChandraAddala: I thought we are doing empty response.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/formats/JSONMetricsContainer.java
```diff
@@ -64,14 +65,18 @@ public JSONMetricsContainer(String tenantId, List<JSONMetric> metrics) {
...
                 locator = Locator.createLocatorFromPathComponents(scopedMetric.getTenantId(), jsonMetric.getMetricName());
+                isValid = scopedMetric.isValid();
             } else {
                 locator = Locator.createLocatorFromPathComponents(tenantId, jsonMetric.getMetricName());
+                isValid = jsonMetric.isValid();
             }
 
-            if (jsonMetric.getMetricValue() != null) {
+            // convert and add metric only if it's valid (no validation error)
+            if ((jsonMetric.getMetricValue() != null) && isValid ) {
```
ChandraAddala: ``toMetrics()`` method now has a side-affect that it will only return valid metrics. May be we can create a separate method to get only valid metrics.

shintasmith: If there are no other consumers of toMetrics() other than HttpMetricsIngestionHandler, I think it's ok to change the behavior.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionHandler.java
```diff
@@ -166,7 +157,14 @@ public void handle(ChannelHandlerContext ctx, HttpRequest request) {
...
+            // verify metrics after parsing json and converting to metrics
+            List<String> errors = jsonMetricsContainer.getValidationErrors();
+            if ( !errors.isEmpty() && containerMetrics.isEmpty()) {
+                // there was validation errors and all metrics were invalid
+                DefaultHandler.sendResponse( ctx, request, getResponseBody( errors ), HttpResponseStatus.BAD_REQUEST );
+                return;
+            } else if (containerMetrics == null || containerMetrics.isEmpty()) {
+                // empty container
                 log.warn(ctx.getChannel().getRemoteAddress() + " No valid metrics");
                 DefaultHandler.sendResponse(ctx, request, "No valid metrics", HttpResponseStatus.BAD_REQUEST);
```
ChandraAddala: Should we ``return`` after this? There is ``return`` in the if condition above.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionHandler.java
```diff
@@ -183,7 +181,18 @@ public void handle(ChannelHandlerContext ctx, HttpRequest request) {
                         return;
                     }
                 }
-                DefaultHandler.sendResponse(ctx, request, null, HttpResponseStatus.OK);
+
+                // after processing metrics
+                // return either OK or MULTI_STATUS depending on number of valid metrics
+                if( !errors.isEmpty() ) {
+                    // has some validation errors, return MULTI_STATUS
+                    DefaultHandler.sendResponse(ctx, request, getResponseBody( errors ), HttpResponseStatus.MULTI_STATUS);
```
ChandraAddala: Same note. I thought this story is supposed to send empty response body.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/formats/JSONMetricsContainer.java
VinnyQ: No because of the scoped extension of the metric.  See the static subclass of JsonMetric called ScopedJSONMetric in the same file (/JSONMetricsContainer.java).

blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/inputs/handlers/HttpHandlerIntegrationTest.java
VinnyQ: fixed:  returning empty response.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/formats/JSONMetricsContainer.java
VinnyQ: I have checked and the only place that uses this method is HttpMetricsIngestionHandler.  Since we do ACTUALLY want to change the behavior (let valid metrics through and reject the invalid), I figure it'll be cleaner to just modify it here.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionHandler.java
VinnyQ: good catch, other if in place prevented bad things from happening after, but safer to add it.

VinnyQ:
@shintasmith @ChandraAddala @usnavi @izrik please review again this complete story of returning 207 multi-status for partially invalid payload.  Thanks!

blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/formats/JSONMetricsContainer.java
VinnyQ: What if we instead add an "ingestTime" property at the handler level which retrieve the System.currentTimeMillis() value once when the request handling started, and use that value for all basis of validity, delays, etc?  We will have to pass this value to the validation and delay tracking/segregation code to use instead of having them call System.currentTimeMillis() when needed.

Let me think about this a little more.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/formats/JSONMetricsContainer.java
ChandraAddala: That has to work right? Its polymorphism.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/formats/JSONMetricsContainer.java
ChandraAddala: I agree that it has to change. But if I look at the method name ``toMetrics()``, It gives an impression that all it does is convert to metrics. No one will expect it to filter invalid metrics unless they look inside.  I would have named it ``getValidMetrics()``. Its not a big deal though. You can leave the way it is.

shintasmith: I vote for changing the method name to getValidMetrics() :-) Should be easy to have IntelliJ do the rename for you.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/formats/JSONMetricsContainer.java
```diff
@@ -142,19 +146,21 @@ public void setTtlInSeconds(int ttlInSeconds) {
             this.ttlInSeconds = ttlInSeconds;
         }
 
+        public boolean isValid() {
+            // valid if collectionTime is NOT too old AND NOT too far in the future
+            return !tooPast(collectionTime) && !tooFuture(collectionTime);
```
shintasmith: Hmmm... This wont work either. Because again, isValid() and getValidationErrors() call these tooPast() and tooFuture() at different times. So a metric can return isValid() true, but got included in the list returned by getValidationErrors(), if its collection time is in the border line of being valid.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/formats/JSONMetricsContainer.java
VinnyQ: Made currentTime a param that has to be passed in for validation and initialize it for the same metric for isValid() and getValidationErrors().

blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/formats/JSONMetricsContainer.java
VinnyQ: You were right, changed back to directly calling isValid().  It caused tests failure earlier for tenantId validation, but I think I was doing something else then that caused it to fail.  It's working now.

VinnyQ:
@shintasmith @ChandraAddala I have added a fix for the current time edge case for validation.  I think it can be improved/optimize more later with some refactoring, but it should be good enough for now.

Let me know if you agree and +1 if you do, thanks!

blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/formats/MetricValidator.java
```diff
@@ -0,0 +1,55 @@
...
+ *    limitations under the License.
+ */
+
+package com.rackspacecloud.blueflood.inputs.formats;
+
+import com.rackspacecloud.blueflood.service.Configuration;
+import com.rackspacecloud.blueflood.service.CoreConfig;
+import org.apache.commons.lang.StringUtils;
+
+public class MetricValidator {
```
shintasmith: I am not sure I like this class. It seems to be doing something very specific, very narrow, and yet the name is so generic. It will be come a dumping ground of unrelated things. It also has static methods. Do you really need to make this class? Why can't we keep the methods in the class that calls them?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/formats/JSONMetricsContainer.java
```diff
@@ -22,21 +22,21 @@
...
 public class JSONMetricsContainer {
+
     private final String tenantId;
     private final List<JSONMetric> jsonMetrics;
     private List<Metric> delayedMetrics;
     private static final long delayedMetricsMillis = Configuration.getInstance().getLongProperty(CoreConfig.DELAYED_METRICS_MILLIS);
-    private static final long pastDiff = Configuration.getInstance().getLongProperty( CoreConfig.BEFORE_CURRENT_COLLECTIONTIME_MS );
-    private static final long futureDiff = Configuration.getInstance().getLongProperty( CoreConfig.AFTER_CURRENT_COLLECTIONTIME_MS );
 
     public JSONMetricsContainer(String tenantId, List<JSONMetric> metrics) {
```
shintasmith: I am proposing this. In the constructor of ```JSONMetricsContainer```, parse all the JSONMetric and organize them to buckets/lists of metrics:
* valid JSON Metrics
* invalid JSON Metrics
* delayed Metrics

Your ```getValidMetrics()``` should simply fetch the list of valid metrics.

Is that possible to do?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/formats/MetricValidator.java
VinnyQ: @shintasmith 
Because both the JSONMetricsContainer (use by HttpMetricsIngestionHandler) and AggregatedPayload (use by HttpAggregatedMultiIngestionHandler) classes need the same validation code.  Where do you propose I put these validation code so both classes can use it, or do you rather I just have duplicate code like before?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/formats/JSONMetricsContainer.java
VinnyQ: That was the refactoring I was thinking of.  I guess we can do it now.  I was also planning to yank all the parsing code in HttpAggregatedMultiIngestionHandler to be its own class similar to JSONMetricsContainer, and have them both inherit from a base class that handles validation.

Give me a day and I'll get these refactoring done.

ChandraAddala: Just a thought, Instead of having three data structures which possibly could contain repeat data, can we have flags in the ``JsonMetric`` indicating what kinda metric it is? So when you want to grab say valid json metric, you filter only the ones based on the flags?.

ChandraAddala: @VinnyQ Instead of the refactoring you proposed, I think its okay to just have that one line of code in two places for now. What do others think?

shintasmith: I like the idea of having a flag in ```JSONMetric```! :+1: This should also solve the problem of having to replicate the logic to check for valid time in 2 places (for normal vs aggregated). We just move the logic to inside ```JSONMetric```

So it will be at least 2 flags, right? Something like ```isDelayed``` and ```hasValidTime```?

VinnyQ: @shintasmith @ChandraAddala 
The body of json for metrics and aggregated metrics are completely different, so we can't just have JSONMetric represent both normal and aggregated message (so the properties would be different for each class, with some common one such as tenantid).  The equivalent of JSONMetric is the AggregatedPayload.  

However, maybe I can move the createBundleList() call in HttpAggregatedMultiIngestionHandler to  JSONMetricsContainer and have both a 

List<JSONMetric> jsonMetrics,  and a 
List<AggregatedPayload> jsonMetricsAggregated Variable there, that can share the same validation code, and not have two classes refactor from a common base class. 

What do you guys think of that?

VinnyQ: @ChandraAddala @shintasmith 
Hey, forget about the refactoring and the flags.  I moved a few classes and methods and was able to remove the MetricValidator class.  See latest commit and tell me what you think.

VinnyQ: @ChandraAddala @shintasmith 
Also, I kinda like shinta idea of parsing the json as the container is created, but instead of having invalidMetrics we just have an error message list that can be return for getValidationErrors().  I refactored it just so and made it so that the validation is done only once per metric (when the container is created and as each metric is processed).

Check it out!

blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/formats/JSONMetricsContainer.java
VinnyQ: combined the isValid() and getValidationErrors() into one new method validate();  Which will only be call once and do a validation on the time and add errors to a returnable errors list.  Convert JSONMetricsContainer to process json when created/initializing, which validates and populate the errors list of any validation failure.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/formats/AggregatedPayload.java
```diff
@@ -61,18 +55,15 @@ public String toString() {
...
-        List<String> errors = new java.util.ArrayList<String>();
-        long current = System.currentTimeMillis();
-
-        if( timestamp > current + futureDiff )
-            errors.add( "'timestamp' '" + timestamp + "' is more than '" + futureDiff + "' milliseconds into the future." );
-        else if( timestamp < current - pastDiff )
-            errors.add( "'timestamp' '" + timestamp + "' is more than '" + pastDiff + "' milliseconds into the past." );
-
-        return errors;
+    public String validate() {
```
shintasmith: This method does more than just validate. Can we name it ```getValidationError``` or ```validateAndGetError``` or maybe just split it to 2 methods?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/formats/JSONMetric.java
```diff
@@ -0,0 +1,98 @@
...
+        if ( tooFuture(currentTime, collectionTime) ) {
+            error = "'" + metricName + "': 'collectionTime' '" + collectionTime + "' is more than '" + AFTER_CURRENT_COLLECTIONTIME_MS + "' milliseconds into the future.";
+        } else if ( tooPast(currentTime, collectionTime) ) {
+            error = "'" + metricName + "': 'collectionTime' '" + collectionTime + "' is more than '" + BEFORE_CURRENT_COLLECTIONTIME_MS + "' milliseconds into the past.";
+        }
+
+        return error;
+    }
+
+    public static boolean tooPast(long currentTime, long collectionTime) {
```
shintasmith: I am not sure I like this being static method. If you think AggregatedPayload and JSONMetric are 2 different classes that conceptually they do not share any common behavior, then having AggregatedPayload calls static methods of JSONMetric is kind of sharing for the sake of sharing. I prefer having the code duplicated than to introduce inter-dependency of classes that do not have anything common.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/formats/AggregatedPayload.java
VinnyQ: I don't like splitting it to 2 methods, which was what I had before, because we have the code do that timestamp compare and calculation against the diffs twice for every metric.  I'll rename it.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/formats/JSONMetric.java
VinnyQ: I really hate duplicating logic.  The classes structure are different, but our idea of what metrics (for either type regular or aggregated) are too far in the past or too far in the future for ingestion is the same.  That's why I originally thought of either:

1)  put the validation code in its own class that can be share with any metrics type (regular or aggregated or otherwise),

2)  create a base class with common validation and properties methods but extend it to radically different metrics (which I tried but didn't like so much so I scrapped it), or

So we can do those two or just go back to option

3) have the tooPast/tooFuture methods which will have the exact validation logics in both AggregratedPayload and JSONMetric class.  

The most correct me is 1) a common validation class, or 2)  have a common base class for both, and maybe rename AggreggatedPayload to be JSONMetricAggregated or something.  

For rebasing, the main thing is that at the moment, we have a JSONMetricContainer that parses and hold json metrics and mapped metric, which is used by HttpMetricsIngestionHandler, but we don't have one for AggregatedPayload class (the same logics is just thrown into the handle() method of HttpAggregatedMultiIngestionHandler class).

But give me another +1 on duplicating the logic of tooPast, tooFuture and I'll do it.

shintasmith: Okay. Let's just keep it as is. :-)

VinnyQ: @shintasmith  Actually nevermind, I have something in mind that'll make us both happy.

Those boolean checks are so simple and the only reason I had them split up to different method is because of the refactoring and how complicated the validation and errors retrieving before.  But since I have simplified it a lot, I think we can just throw the check in each of the validation method.  Hold for change :)

VinnyQ:
Ok @shintasmith @ChandraAddala please review, thanks.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/formats/JSONMetric.java
VinnyQ: @shintasmith I revert it to almost it's original form.  I think at the time we wanted to extract the logic out to it's own methods but we didn't realized they were in two places.  I think it reads better now with comments and with the processJson() called during the JSONMetricsContainer to validate and map the metrics, adding the errors to a class property that is returned by getValidationErrors() -- as oppose to doing the validation right in there, makes the code much cleaner and simpler.  I think you'll like it now.

shintasmith:
lgtm 👍 , thanks @VinnyQ !!

ChandraAddala:
Looks good.

PR #691, opened 2016-05-05
Delayed metrics part2

ChandraAddala:
This is step 2 of handling delayed metrics as described in the wiki page. https://one.rackspace.com/pages/viewpage.action?pageId=162810603#Delayedmetrics-Analysis&Design-Step2:Schedulingslotsofdelayedmetrics

We had decided to go with a combination of configs, one for metrics with short delay(DELAYED_METRICS_ROLLUP_DELAY_MILLIS) and another for metrics with long delay(DELAYED_METRICS_ROLLUP_WAIT_PERIOD_MILLIS)

The changes include
- Adding new config DELAYED_METRICS_ROLLUP_WAIT_PERIOD_MILLIS
- Logic to schedule metrics with longer dealy only after waiting certain time to prevent frequent re-rolls.
- Added meters/graphs for identifying short/long delayed metrics, re-rolls waiting to happen. 
- Removing usage of config DELAYED_METRICS_MILLIS as we could use ROLLUP_DELAY_MILLIS

I created a separate dashboard for delayed metrics. Here is the link for perf01 env.
http://graphite00.dfw.perf01.bf.k1k.me/grafana/#/dashboard/db/delayed-metrics


blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/service/ShardStateIntegrationTest.java
```diff
@@ -38,14 +41,17 @@
...
 @RunWith(Parameterized.class)
 public class ShardStateIntegrationTest extends IntegrationTestBase {
     
     private ShardStateIO io;
     private List<Integer> shards;
 
-    private final int DELAYED_METRICS_ROLLUP_DELAY_MILLIS = 300000;
+    private final int DELAYED_METRICS_ROLLUP_DELAY_MILLIS = 600000;
     private final int ROLLUP_DELAY_MILLIS = 300000;
+    private final int ROLLUP_WAIT_PERIOD_MILLIS = 500000;
```
shintasmith: If you are going to rename the dashboard graphs: short vs long delay, why don't we name the config the same?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/CoreConfig.java
```diff
@@ -164,7 +165,6 @@
     ROLLUP_ON_READ_REPAIR_SIZE_PER_THREAD( "5" ),
     ROLLUP_ON_READ_TIMEOUT_IN_SECONDS("10"),
 
-    DELAYED_METRICS_MILLIS("300000"),
```
shintasmith: Why is this removed?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/CoreConfig.java
```diff
@@ -151,6 +151,7 @@
     // how long we typically wait to schedule a rollup.
     ROLLUP_DELAY_MILLIS("300000"),
     DELAYED_METRICS_ROLLUP_DELAY_MILLIS("300000"),
+    DELAYED_METRICS_ROLLUP_WAIT_PERIOD_MILLIS("300000"),
```
shintasmith: Should we name the config with short & long delays, like the graphs?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/ShardStateManager.java
```diff
@@ -304,9 +312,41 @@ protected UpdateStamp getAndSetState(int slot, UpdateStamp.State state) {
...
+         *     |                                        |
+         *     |<-- ROLLUP_DELAY--->|                   |     |<--ROLLUP_WAIT-->|                |<--ROLLUP_WAIT-->|
+         *     |                    |                   |     |                 |                |                 |
+         * -------------------------------------------------------------------------------------------------------------
+         * |slot|                   ^          ^        ^     ^                 ^    ^   ^   ^   ^                 ^
+         * | X  |                   |          |        |     |                 |    |   |   |   |                 |
+         *                          |   delayed metric  | delayed metric        |   delayed metrics                |
+         *                          |                   |                       |                                  |
+         *                      1st rollup         1st re-roll              2nd re-roll                      3rd re-roll
+         *
```
shintasmith: niiiiiicccccceeee!

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/CoreConfig.java
ChandraAddala: This one is only being used for one purpose. Just to determine if a metric is delayed or not and we use it in the tracking code. Since we plan to keep ROLLUP_DELAY_MILLIS the same 10 mins all the time, it makes sense to use that instead. One less config.

shintasmith: If I am not mistaken, the other one is used by Tracker, to print/log whether or not a metric is delayed. I know it seems extra/redundant but I think we have used it in the past to just catch metrics that are delayed by certain time, without actually changing the actual ROLLUP_DELAY_MILLIS. 

I agree the name should be more clear that it's used by Tracker.

contrib/grinder/scripts/utils.py
```diff
@@ -28,7 +28,7 @@
   'enum_num_values': 10,
   'singleplot_per_interval': 10,
   'annotations_queries_per_interval': 8,
-  'ingest_delay_millis': 0}
+  'ingest_delay_millis': ""}
```
shintasmith: looks like you are now expecting a comma separated value? can you put comments here or the properties file the format? Is it always 3 values? What are the 1st, 2nd, 3rd values for, etc?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/CoreConfig.java
ChandraAddala: In that case, I will put it back.

shintasmith:
other than my existing comments, PR looks good

ChandraAddala:
@shintasmith I had addressed the changes you has asked for. Pls take a look.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/ScheduleContext.java
```diff
@@ -214,22 +214,12 @@ public void update(long millis, int shard) {
...
-     * If the difference between {@link #scheduleTime} and a given slot's
-     * {@link UpdateStamp} in milliseconds is greater than
-     * {@code maxAgeMillis}, then that slot will scheduled.
-     *
-     * @param maxAgeMillis
-     * @param delayedMetricsMaxAgeMillis
      */
     // only one thread should be calling in this puppy.
-    void scheduleEligibleSlots(long maxAgeMillis, long delayedMetricsMaxAgeMillis) {
+    void scheduleEligibleSlots(long maxAgeMillis, long rollupDelayForMetricsWithShortDelay, long rollupWaitForMetricsWithLongDelay) {
```
usnavi: Would you please add brief description of these 3 parameters.  At the every least I was confused with maxAgeMillis since that term isn't used elsewhere (that I noticed).  It's rollupDelayMillis in other places, right?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/ShardStateManager.java
```diff
@@ -317,34 +357,51 @@ protected UpdateStamp getAndSetState(int slot, UpdateStamp.State state) {
...
+
+                        if (update.getLastIngestTimestamp() > 0 ) {
+                            long delayOfLastIngestedMetric = update.getLastIngestTimestamp() - update.getTimestamp();
+                            final long timeElapsedSinceLastIngest = now - update.getLastIngestTimestamp();
 
-                            granToReRollMeters.get(granularity).mark();
-                            if (nowMillis - update.getTimestamp() >= millisInADay) {
-                                granToDelayedMetricsMeter.get(granularity).mark();
+                            //long delay
+                            if (delayOfLastIngestedMetric > rollupDelayForMetricsWithShortDelay &&
```
usnavi: Is this check necessary?  If we already checked that we've passed the short delay above (line 369) is this necessary?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/ScheduleContext.java
ChandraAddala: description is added in the subsequent method where actual scheduling happens.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/ShardStateManager.java
```diff
@@ -222,61 +235,56 @@ protected void updateSlotOnRead(SlotState slotState) {
...
             final long lastUpdateTimestamp = slotState.getLastUpdatedTimestamp();
 
-            UpdateStamp stamp = slotToUpdateStampMap.get(slot);
-            long nowMillis = new DateTime().getMillis();
-            if (stamp == null) {
+            UpdateStamp stampInMemory = slotToUpdateStampMap.get(slot);
+            if (stampInMemory == null) {
                 // haven't seen this slot before, take the update. This happens when a blueflood service is just started.
-                slotToUpdateStampMap.put(slot, new UpdateStamp(timestamp, state, false));
-            } else if (stamp.getTimestamp() != timestamp && state.equals(UpdateStamp.State.Active)) {
```
usnavi: I have questions about this section.  Just making a note so we don't forget to cover this.  Probably easiest to talk.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/ScheduleContext.java
usnavi: cool!

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/ShardStateManager.java
```diff
@@ -222,61 +235,56 @@ protected void updateSlotOnRead(SlotState slotState) {
...
             final long timestamp = slotState.getTimestamp();
             UpdateStamp.State state = slotState.getState();
+
+            //For slots in state "A", this would be last ingest time
+            //For slots in state "X", this would be last rollup time
             final long lastUpdateTimestamp = slotState.getLastUpdatedTimestamp();
 
-            UpdateStamp stamp = slotToUpdateStampMap.get(slot);
-            long nowMillis = new DateTime().getMillis();
-            if (stamp == null) {
```
usnavi: Can we put a comment how the "last rolled timestamp" and "last ingested timestamps" are being persisted to the UpdateStamp object in memory?  I think that would be enough for me to follow this when I look at it later.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/ShardStateManager.java
ChandraAddala: So the first time we cross the short delay(2hr) barrier, we are supposed to rollup. The problem was how do we know its the first time? The condition is, when we receive a delayed metric, we check if it was ingested within the short delay. If it is, its probably the first time.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/ShardStateManager.java
ChandraAddala: @usnavi Added some comments to the method description.

usnavi: thanks.  lgtm

PR #692, opened 2016-05-09
Datastax support for BasicRollup, Numeric, Strings & Booleans

usnavi:
This feels like my life's work! :-P

Does the following
- Adds datastax support for BasicRollup, Numeric, Strings & Booleans
- Allow AstyanaxReader to read Booleans (previously it alway read 'false')
- Refactored DPreaggregatedMetricsRW to allow basic numerics to leverage its methods.
- Renamed DatastaxMetadataIO -> DMetadataIO
- Renamed DatastaxShardStateIO -> DShardStateIO
- Updated datastax exception logging for DMetadataIO & DShardStateIO


blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AbstractMetricsRW.java
```diff
@@ -148,22 +149,19 @@ protected int getTtl(Locator locator, RollupType rollupType, Granularity granula
      * @param timestampToRollupMap  a map of timestamp to rollup
      * @return
      */
-    protected <T extends Rollup> Points<T> convertToPoints(final Map<Long, T> timestampToRollupMap) {
-        Points<T> rollupPoints =  new Points<T>();
-        for (Map.Entry<Long, T> rollup : timestampToRollupMap.entrySet() ) {
-            rollupPoints.add(new Points.Point<T>(rollup.getKey(), rollup.getValue()));
+    protected Points convertToPoints(final Map<Long, Object> timestampToRollupMap) {
```
shintasmith: Is there any reason for not doing it this way? 
```protected Points<T> convertToPoints(final Map<Long, T> timestampToRollupMap)```

File: blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AbstractMetricsRW.java [hunk 2]
```diff
@@ -148,22 +149,19 @@ protected int getTtl(Locator locator, RollupType rollupType, Granularity granula
...
-     * Only our tests should call this method. Services should call either insertMetrics(Collection metrics)
-     * or insertRollups()
-     *
-     * @param metrics
-     * @throws IOException
-     */
-    @VisibleForTesting
-    public abstract void insertMetrics(Collection<IMetric> metrics, Granularity granularity) throws IOException;
+    protected Points.Point createPoint( Long timestamp, Object value ) {
+        if( value instanceof Rollup || value instanceof String)
```
shintasmith: I missed bounded type parameters already :-) I don't know if it would be better to create separate versions of ```createPoint``` methods? Like this?
```
protected Points.Point createPoint(Long timestamp, Rollup value);
protected Points.Point createPoint(Long timestamp, String value);
protected Points.Point createPoint(Long timestamp, SimpleNumber value);
```

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/EnumMetricData.java
```diff
@@ -117,17 +117,17 @@ public MetricData getEnumMetricDataForRange(final Locator locator, final Range r
 
         });
 
-        Future<Table<Locator, Long, BluefloodEnumRollup>> enumHashRollupFuture = taskExecutor.submit(new Callable() {
+        Future<Table<Locator, Long, Object>> enumHashRollupFuture = taskExecutor.submit(new Callable() {
```
shintasmith: I am not sure why this needs to be changed from ```BluefloodEnumRollup``` to ```Object```. Can you elaborate?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AbstractMetricsRW.java
usnavi: Is that really necessary?  For Basic metrics we'll have to have an instanceof check for Strings no matter what.  We could make it a abstract method, but that would either mean implementing it for Datastax and Astyanax or making more parent classes to hold the methods, one for Preaggregated and one for Basics.

usnavi:
@shintasmith - re-added some of the generic method signatures.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AbstractMetricsRW.java
```diff
@@ -148,22 +150,19 @@ protected int getTtl(Locator locator, RollupType rollupType, Granularity granula
      * @param timestampToRollupMap  a map of timestamp to rollup
      * @return
      */
-    protected <T extends Rollup> Points<T> convertToPoints(final Map<Long, T> timestampToRollupMap) {
-        Points<T> rollupPoints =  new Points<T>();
-        for (Map.Entry<Long, T> rollup : timestampToRollupMap.entrySet() ) {
-            rollupPoints.add(new Points.Point<T>(rollup.getKey(), rollup.getValue()));
+    protected <T extends Object> Points<T> convertToPoints(final Map<Long, T> timestampToRollupMap) {
```
ChandraAddala: ``<T extends Object>`` doesnt make much sense. I thought everything extends Object. Besides what happened to ``<T extends Rollup>``?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/PreaggregatedRW.java
```diff
@@ -0,0 +1,26 @@
...
+    /**
+     * Inserts a collection of rolled up metrics to the metrics_preaggregated_{granularity} column family.
+     * Only our tests should call this method. Services should call either insertMetrics(Collection metrics)
+     * or insertRollups()
+     *
+     * @param metrics
+     * @throws IOException
+     */
+    @VisibleForTesting
+    public abstract void insertMetrics(Collection<IMetric> metrics, Granularity granularity) throws IOException;
```
ChandraAddala: The only method in this interface is created for the sole purpose of testing?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AbstractMetricsRW.java
shintasmith: I agree it's kind of silly. But I don't like the fact that a lot of method signatures have to be changed from specific class to just ```Object```. For example, this method in EnumMetricData:
```
    private void populateEnumValueToCountMap(Map<Long, BluefloodEnumRollup> enumRollupMap, Map<Long, String> enumHashValuesMap)
```
had to be changed to:
```
    private void populateEnumValueToCountMap(Map<Long, Object> enumRollupMap, Map<Long, String> enumHashValuesMap) 
```
in Greg's first iteration of this PR.

It does not make sense to change that method to take a ```Map<Long, Object>```. It is a class about Enum.

The problem is he wants to use this method for simple numbers too, like Long, Float, Double. There is no way to achieve that because the Rollup objects don't share a common parent class with those Long, Float, Double.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/PreaggregatedRW.java
usnavi: Yep, that is correct.  The ```*BasicMetricsRW``` can't easily support it, so I took it out of the base class that they share with the ```*PreaggregatedMetricsRW```.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AbstractMetricsRW.java
shintasmith: I guess it's not necessary, it's just more explicit. It would also get rid of the instanceof. Why do you say for Basic metrics you'll have to do instanceof check? If we make a ```createPoint()``` method that takes ```String```, we still have to do instanceof String check?

blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/io/datastax/DEnumIOIntegrationTest.java
```diff
@@ -69,14 +69,14 @@ public void testWriteAndReadJoinConsistency() throws Exception {
             List<ResultSetFuture> results = dEnumIO.selectForLocatorAndRange(CassandraModel.CF_METRICS_PREAGGREGATED_FULL_NAME, locator, range);
             Assert.assertEquals("ResultSetFuture list size", 2, results.size());
 
-            Table<Locator, Long, Rollup> locatorRollup = dEnumIO.toLocatorTimestampRollup(results, locator, Granularity.FULL);
-            Map<Long, Rollup> resultMap = locatorRollup.row(locator);
+            Table<Locator, Long, Object> locatorRollup = dEnumIO.toLocatorTimestampValue( results, locator, Granularity.FULL );
+            Map<Long, Object> resultMap = locatorRollup.row(locator);
```
shintasmith: Does these need to stay ```Map<Long, Object>``` ?

usnavi: Yeah, I can fix this.  The fact that I had changed the method name confused me.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DAbstractMetricIO.java
```diff
@@ -183,7 +267,9 @@ public ResultSetFuture putAsync(Locator locator, long collectionTime, Rollup rol
      * @param range
      * @return
      */
-    protected <T extends Rollup> Table<Locator, Long, T> getRollupsForLocators(final List<Locator> locators, String columnFamily, Range range) {
+    protected <T extends Object> Table<Locator, Long, T> getRollupsForLocators(final List<Locator> locators,
```
shintasmith: I was confused at first why this was changed from ```<T extends Rollup>``` to ```<T extends Object>```. But now I see you are using it to also fetch basic metrics from metrics_full. Should we change the name to something else since it no longer fetches just Rollup objects? Maybe ```getDataForLocators``` or ```getMetricsForLocators``` ?

Updating the javadoc/comment would also help :)

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AbstractMetricsRW.java
ChandraAddala: My vote is to preserve the signature atleast for the ones where there is a common parent class. That would mean creating multiple methods.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DAbstractMetricIO.java
```diff
@@ -206,7 +292,7 @@ public ResultSetFuture putAsync(Locator locator, long collectionTime, Rollup rol
      * @param rollup
      * @return
      */
-    protected abstract <T extends Rollup> ByteBuffer toByteBuffer(T rollup);
+    protected abstract <T extends Object> ByteBuffer toByteBuffer(T rollup);
```
shintasmith: Parameter name should be something else if it's being used for things other than rollup.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AbstractMetricsRW.java
usnavi: I feel that sharing implementation is more important than having to do manual type checking.  Its the only way to ensure that the 2 implementations don't diverge.

The real fix is to fix the Object vs Rollup dichotomy when writing to metrics_full vs one of the other granularities.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/PreaggregatedRW.java
ChandraAddala: What do we loose if we drop this method since its only being used for testing?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AbstractMetricsRW.java
shintasmith: I am ok with multiple methods. I am ok with ```<T extends Object>``` as well. It is less than ideal, but to do it the ideal way, it is more refactoring that is outside the scope of this PR.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AbstractMetricsRW.java
shintasmith: @usnavi and I talked offline. I tracked down where the instanceof String check is happening. One place is in the DataType.getMetricsType(). I was trying to get rid of yet another instanceof String check. But at this point, it's more aesthetic comment than anything else. So I am ok with leaving it as is.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DAbstractMetricsRW.java
```diff
@@ -0,0 +1,324 @@
...
+            RollupType rType = RollupType.fromString( rollupType );
+
+            // get the right PreaggregatedIO class that can process
+            // this rollupType
+            io = rollupTypeToIO.get(rType);
+        }
+
+
+        if (io == null) {
+            throw new IllegalArgumentException(String.format("dont know how to handle locator=%s with rollupType=%s and type=%s", locator, rollupType, dataType));
```
shintasmith: Should we make the exception the same as in line 78? I don't care which one. I just think we should make them the same.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DAbstractMetricsRW.java
```diff
@@ -0,0 +1,324 @@
...
+                ResultSetFuture future = io.putAsync(locator, writeContext.getTimestamp(), rollup, writeContext.getGranularity(), ttl);
+                futureLocatorMap.put(future, writeContext);
+            }
+
+            for (ResultSetFuture future : futureLocatorMap.keySet()) {
+                try {
+                    future.getUninterruptibly().all();
+                } catch (Exception ex) {
+                    SingleRollupWriteContext writeContext = futureLocatorMap.get(future);
+                    LOG.error(String.format("error writing to locator %s, granularity %s", writeContext.getLocator(), writeContext.getGranularity()), ex);
```
ChandraAddala: I dont see much instrumentation going on for errors. Something like below.
```
Instrumentation.markWriteError(e);
```

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DAbstractMetricsRW.java
```diff
@@ -0,0 +1,324 @@
...
+        }
+
+
+        if (io == null) {
+            throw new IllegalArgumentException(String.format("dont know how to handle locator=%s with rollupType=%s and type=%s", locator, rollupType, dataType));
+        }
+
+        // put everything in a map of locator -> io so
+        // we can use em up later
+        locatorIOMap.put( locator, io );
```
shintasmith: Why having a ```Map<Locator, IO>``` here? Do we expect the loop in line 142 to have multiple occurrences of the same Locator?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DAbstractMetricsRW.java
usnavi: I can actually get rid of the dataType param.  I previously was using it to get the simpleIO object in its own map, but then it turned out there was only 1 value in the map.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DAbstractMetricsRW.java
```diff
@@ -0,0 +1,324 @@
...
+     */
+    @Override
+    public void insertRollups(List<SingleRollupWriteContext> writeContexts) throws IOException {
+
+        if (writeContexts.size() == 0) {
+            return;
+        }
+
+        Map<ResultSetFuture, SingleRollupWriteContext> futureLocatorMap = new HashMap<ResultSetFuture, SingleRollupWriteContext>();
+        Timer.Context ctx = Instrumentation.getWriteTimerContext( writeContexts.get( 0 ).getDestinationCF().getName() );
```
ChandraAddala: Should this be ``Instrumentation.getBatchWriteTimerContext``?

usnavi: BatchWriter is only for Astyanax, right?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DAbstractMetricsRW.java
```diff
@@ -0,0 +1,324 @@
...
+    /**
+     * This method inserts a collection of metrics in the
+     * {@link com.rackspacecloud.blueflood.service.SingleRollupWriteContext}
+     * objects to the appropriate Cassandra column family
+     *
+     * @param writeContexts
+     * @throws IOException
+     */
+    @Override
+    public void insertRollups(List<SingleRollupWriteContext> writeContexts) throws IOException {
```
ChandraAddala: What exactly throws ``IOException`` here?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DAbstractMetricsRW.java
ChandraAddala: We are just eating exception here.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/PreaggregatedRW.java
usnavi: The PreaggregatedMetrics tests were written using this method.  The BasicMetrics tests were written using ```SingleRollupWriteContext```.  PreaggregatedMetrics could use that as well, but figured no need to mess with it.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DBasicNumericIO.java
```diff
@@ -0,0 +1,33 @@
+package com.rackspacecloud.blueflood.io.datastax;
```
shintasmith: Can we name this class DBasicRollupIO to make it obvious that it operates with BasicRollup? I get confused this class versus DSimpleNumberIO.

usnavi: I guess...I was trying to follow your convention with no Rollups in the name. :-P

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DAbstractMetricsRW.java
usnavi: Yep, that's what we for all the metrics stuff.  We return with incomplete data and log the error.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DBasicNumericIO.java
shintasmith: OH! It's fine then, just add the comment to make it more clear the difference between DBasicNumericIO and DSimpleNumberIO.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DAbstractMetricsRW.java
shintasmith: Eating exceptions, yum!

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DAbstractMetricsRW.java
```diff
@@ -0,0 +1,324 @@
...
+        try {
+            // read the rollup object from the proper IO class
+            DAbstractMetricIO io;
+
+            if( columnFamilyName.equals( CassandraModel.CF_METRICS_FULL_NAME) ) {
+
+                io = simpleIO;
+            }
+            else
+                io = rollupTypeToIO.get(rollupType);
```
shintasmith: Why don't we use getIO() here?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DAbstractMetricsRW.java
ChandraAddala: side effect.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DAbstractMetricsRW.java
usnavi: This was pulled up from the previous implementation of ```DPreaggregatedMetricsRW```.  I assumed things would get wired together in the next task.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DAbstractMetricsRW.java
```diff
@@ -0,0 +1,324 @@
...
+     */
+    private DAbstractMetricIO getIO( Map<Locator, DAbstractMetricIO> locatorIOMap,
+                                     Locator locator,
+                                     String dataType,
+                                     String rollupType,
+                                     Granularity gran ) {
+
+        DAbstractMetricIO io;
+
+        if( rollupType == null && gran == Granularity.FULL ) {
```
shintasmith: Can we clean this up a bit? There is a RollupType.NOT_A_ROLLUP. Can we somehow assign when rollupType is null and granularity is FULL, then it's RollupType.NOT_A_ROLLUP. Then just add an entry in ```rollupTypeToIO``` map for the simpleIO.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DBasicMetricsRW.java
```diff
@@ -0,0 +1,350 @@
...
+                }
+
+                Locator locator = metric.getLocator();
+
+                if( !isLocatorCurrent( locator ) ) {
+
+                    setLocatorCurrent( locator );
+
+                    if( !DataType.isStringOrBoolean( metric.getMetricValue() ) )
+                        locatorIO.insertLocator( locator );
```
ChandraAddala: So, string/boolean metrics do not get rolled up?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DAbstractMetricsRW.java
shintasmith: But good catch, @ChandraAddala , we should add the ```Instrumentation.markWriteError()``` here.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DBasicMetricsRW.java
ChandraAddala: Also isnt ``shouldPersist(metric)`` already taking care of string/boolean. Why add condition here again?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DBasicMetricsRW.java
```diff
@@ -0,0 +1,350 @@
...
+     */
+    @Override
+    public Points getDataToRollup(final Locator locator,
+                                  RollupType rollupType,
+                                  Range range,
+                                  String columnFamilyName) throws IOException, CacheException {
+
+        if( columnFamilyName.equals ( CassandraModel.CF_METRICS_STRING_NAME ) ) {
+
+            String msg = String.format( String.format( "DBasicMetricsRW.getDataToRollup: %s Attempting to write String/Boolean metric for a method which should not get them.",
```
ChandraAddala: Isnt it "Attempting to read" ?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DAbstractMetricsRW.java
shintasmith: Yeah, let's move the put to outside the method, to the place where this method is being called. Then maybe you can make a generic getIO() method that can be used in multiple places

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DAbstractMetricsRW.java
shintasmith: I am not sure I understand your reply here. My comment was for the exception that is thrown. Here, when we get null io, we throw IllegalArgumentException. In line 78, we throw InvalidDataException. Should we be making them throw the same exception?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DRawIO.java
```diff
@@ -0,0 +1,145 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger( DRawIO.class );
+
+    public static final String KEY = "key";
+    public static final String COLUMN1 = "column1";
+    public static final String VALUE = "value";
+
+    private PreparedStatement putString;
+    private PreparedStatement putNumeric;
+    private PreparedStatement getString;
+    private PreparedStatement getLastString;
```
ChandraAddala: Can you make these 3 instance variables final?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DSimpleNumberIO.java
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * This class holds the utility methods to read/write simple numbers
+ * using Datastax driver.
+ */
+public class DSimpleNumberIO extends DAbstractMetricIO {
+
+    private RawSerDes serDes = new RawSerDes();
+
+    @Override
+    protected ByteBuffer toByteBuffer( Object rollup ) {
```
ChandraAddala: Should we have check for acceptable type for rollup object like we have everywhere?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DAbstractMetricsRW.java
shintasmith: Take a look at RollupRunnable where the old AstyanaxReader.getDataToRoll() was called:
https://github.com/rackerlabs/blueflood/blob/master/blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/RollupRunnable.java#L143

We are going to have to call either DBasicMetricsRW or DPreaggregatedMetricsRW depending on the metrics RollupType. It seems to be that this ```if ( columnFamilyName.equals(CassandraModel.CF_METRICS_FULL_NAME) )``` should be in the DBasicMetricsRW, no? 

One idea is to have an abstract ```getIO()``` method that returns you the right IO object. Move all the ```rollupTypeToIO`` map stuff to the DPreaggregatedMetricsRW (since it only deals with Rollup). Therefore, this DAbstractMetricsRW abstract class don't need to do all these logic to determine which IO object is appropriate.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DBasicMetricsRW.java
```diff
@@ -0,0 +1,350 @@
...
+
+        for( Map.Entry<Locator, ResultSetFuture> future : futures.entrySet() ) {
+
+            Points points = new Points<String>();
+
+            try {
+                for ( Row row : future.getValue().getUninterruptibly().all() ) {
+
+                    Object value = isBoolean ? Boolean.valueOf( row.getString( VALUE ) ) : row.getString( VALUE );
+                    points.add( new Points.Point( row.getLong( COLUMN1 ), value ) );
```
shintasmith: These seems to be in the wrong layer. The *RW classes are not supposed to interact with ColumnFamily's COLUMN and VALUE. Those are the job of the *IO classes. 

Is there any reasons why these are not put in the DStringIO or DBooleanIO, which could inherit from DAbstractMetricIO.selectForLocatorAndRange() ?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DBasicMetricsRW.java
```diff
@@ -0,0 +1,350 @@
...
+        Timer.Context ctx = Instrumentation.getWriteTimerContext( CassandraModel.CF_METRICS_FULL_NAME );
+
+        Map<Locator, ResultSetFuture> futures = new HashMap<Locator, ResultSetFuture>();
+
+        try {
+
+            for( IMetric metric : metrics ) {
+
+                if (!shouldPersist(metric)) {
+                    LOG.trace("Metric shouldn't be persisted, skipping insert", metric.getLocator().toString());
```
shintasmith: This one doesn't have a format. First argument should be a string format. 

Also, should we be logging INFO? Is that too noisy? If we are dropping metrics, it feels like it should be more than TRACE. But this one is a minor comment. :-) I'm ok either way.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DBasicMetricsRW.java
usnavi: ```shouldPersist()``` keep us from 
- writing strings when blueflood is not configured to write them
- writing strings when the last-written-value is the same as the to-be-written value

the if condition just keeps us from writing the string locator value.  

I believe these are 2 different conditions.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DBasicMetricsRW.java
```diff
@@ -0,0 +1,350 @@
...
+     * specified {@link com.rackspacecloud.blueflood.types.Locator} and
+     * {@link com.rackspacecloud.blueflood.types.Range} from the specified column family
+     *
+     * @param locator
+     * @param range
+     * @param gran
+     * @return
+     */
+    @Override
+    public MetricData getDatapointsForRange( final Locator locator, Range range, Granularity gran ) throws IOException {
```
shintasmith: Isn't this a dup of DAbstractMetricsRW.getDatapointsForRange() ? I couldn't tell the diff.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DBasicMetricsRW.java
usnavi: As for the rolled up question, yep, no rollups for strings (and therefore) booleans.   Every request, no matter what granularity, goes to the ```metrics_string``` table.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DBasicMetricsRW.java
```diff
@@ -0,0 +1,350 @@
...
+     *
+     * @param locators
+     * @param range
+     * @param gran
+     * @return
+     */
+    @Override
+    public Map<Locator, MetricData> getDatapointsForRange( List<Locator> locators, Range range, Granularity gran ) throws IOException {
+
+        Timer.Context ctx = Instrumentation.getReadTimerContext( "DBasicMetricsRW.getDatapointsForRange" );
```
shintasmith: Instrumentation.getReadTimerContext() is meant to take ColumnFamily name.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DBasicMetricsRW.java
```diff
@@ -0,0 +1,350 @@
...
+                    points.add( new Points.Point( row.getLong( COLUMN1 ), value ) );
+                }
+
+                MetricData.Type type = isBoolean ? MetricData.Type.BOOLEAN : MetricData.Type.BOOLEAN;
+
+                metrics.put( future.getKey(), new MetricData( points, getUnitString( future.getKey() ), type ) );
+            }
+            catch (Exception e ) {
+
+                LOG.error( String.format( "error writing metric %s, metrics_string", future.getKey() ), e );
```
ChandraAddala: Do you mean "error reading"? We need to add instrumentation here too.

ChandraAddala: Also this should throw ``IOException`` or something.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DSimpleNumberIO.java
usnavi: It's not a rollup (I'll change the parameter).  The serializer can be called on any ```Object```.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DBasicMetricsRW.java
```diff
@@ -0,0 +1,350 @@
...
+            for ( Locator locator : locators ) {
+
+                Object type;
+                try {
+
+                    type = metadataCache.get( locator, DATA_TYPE_CACHE_KEY );
+
+                } catch ( CacheException e ) {
+
+                    LOG.error( "Unable to read cache", e );
```
shintasmith: Log also the locator.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DBasicMetricsRW.java
usnavi: I think TRACE is appropriate as its expected behavior.  

I can add a string format, but its not required.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DBasicMetricsRW.java
ChandraAddala: I had the question about  ``shouldPersist()`` after looking at the existing code for Astyanax which only has this check at one place..

https://github.com/rackerlabs/blueflood/blob/9dd237ad81d5d7b3edc9c8903c227a431fd8b0d3/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/astyanax/AstyanaxWriter.java#L125

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DBasicMetricsRW.java
```diff
@@ -0,0 +1,350 @@
...
+     * @param gran
+     * @return
+     * @throws IOException
+     */
+    private Map<Locator, MetricData> getNumericOrStringDataForRange( List<Locator> locators,
+                                                                     Range range,
+                                                                     String columnFamily,
+                                                                     Granularity gran ) throws IOException {
+
+        Map<Locator, MetricData> metrics = new HashMap<Locator, MetricData>();
```
shintasmith: In AstyanaxReader, there is a call to ```Instrumentation.markScanAllColumnFamilies()``` when this method is called. Do we want to preserve that?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DBasicMetricsRW.java
ChandraAddala: Where in the log message ``"Metric shouldn't be persisted, skipping insert"`` will the ``metric.getLocator().toString()`` value be placed? Shouldnt the format be like this?

``"Metric shouldn't be persisted, skipping insert for {}"``

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DBasicMetricsRW.java
usnavi: This method handles writing to ```metrics_full``` and ```metrics_string```.   I believe ```getReadTimerContext``` can take any arbitrary string and creates a Timer, so I think its OK, but I understand if it breaks convention.

You think its worth it to only put individual timers around  following calls?
- metrics.putAll( getBooleanDataForRange( booleans, range ) );
- metrics.putAll( getStringDataForRange( strings, range ) );
- metrics.putAll( getNumericOrStringDataForRange( unknowns, range, columnFamily, gran )

shintasmith: I guess not... given that metrics_string is going away, I'd say it's not worth it. But is this method not called for when we read from metrics_5m, metrics_20m, etc?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DBasicMetricsRW.java
usnavi: @ChandraAddala  - I see the second check here as well:  https://github.com/rackerlabs/blueflood/blob/9dd237ad81d5d7b3edc9c8903c227a431fd8b0d3/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/astyanax/AstyanaxWriter.java#L135

am i missing something?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DBasicMetricsRW.java
usnavi: Blueflood doesn't throw exceptions on reads, it logs the error and silently continues.  I don't think we should change this now.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DBasicMetricsRW.java
```diff
@@ -0,0 +1,350 @@
...
+     * @param columnFamilyName
+     * @return
+     */
+    @Override
+    public Points getDataToRollup(final Locator locator,
+                                  RollupType rollupType,
+                                  Range range,
+                                  String columnFamilyName) throws IOException, CacheException {
+
+        if( columnFamilyName.equals ( CassandraModel.CF_METRICS_STRING_NAME ) ) {
```
shintasmith: I'm ok leaving this check here but I know this will never happen. RollupRunnable is the only caller of this method and it would never try to roll String metrics.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DBasicMetricsRW.java
usnavi: Oh yeah, its reading that as well.  But its the same question, right?  Despite the granularity you might have strings and you might have numerics.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DBasicMetricsRW.java
usnavi: Yeah, I put that here to be extra paranoid.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DBasicMetricsRW.java
shintasmith: But my point is that previously we have individual read metrics for each column family. Now you are grouping all reads into one read metric.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DBasicMetricsRW.java
ChandraAddala: I must be blind :)

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DAbstractMetricsRW.java
usnavi: Yeah, I like the abstract ```getIO()```.  I'll look into that.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DBasicMetricsRW.java
shintasmith: You can be less paranoid when we gut out String metrics :-P

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DEnumIO.java
```diff
@@ -165,7 +168,7 @@ protected ByteBuffer toByteBuffer(Rollup rollup) {
      * @return
      */
     @Override
-    protected BluefloodEnumRollup fromByteBuffer(ByteBuffer byteBuffer) {
+    protected Object fromByteBuffer(ByteBuffer byteBuffer) {
```
shintasmith: I think here we don't need to change ```BluefloodEnumRollup``` to ```Object```

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DRawIO.java
```diff
@@ -0,0 +1,145 @@
...
+import java.util.List;
+
+import static com.datastax.driver.core.querybuilder.QueryBuilder.*;
+
+
+/**
+ * This class deals with writing boolean, numeric and string values to the metrics_full & metrics_string column
+ * families using the Datastax driver.
+ */
+public class DRawIO {
```
shintasmith: This is not a big deal but is it worth splitting the DRawIO to 2 classes? One that deals with writing to metrics_full and another deals with writing to metrics_string?

blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/io/BasicMetricsRWAsytanaxWriteIntegrationTest.java
```diff
@@ -0,0 +1,616 @@
...
+import java.util.*;
+
+import static org.junit.Assert.*;
+
+/**
+ * Verify MetricsRW for basic metrics (SimpleNumber, String, Boolean) implementations, here when Astyanax is writing and
+ * Datastax is reading.
+ */
+@RunWith( JUnitParamsRunner.class )
+public class BasicMetricsRWAsytanaxWriteIntegrationTest extends BasicMetricsRWIntegrationTest {
```
shintasmith: You are not realizing your full potential. Should've name this ```BasicMetricsRWAstyanaxWriteDatastaxReadIntegrationTest``` 😂

blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/io/BasicMetricsRWAsytanaxWriteIntegrationTest.java
```diff
@@ -0,0 +1,616 @@
...
+                    expectedMetric.getMetricValue(),
+                    rollup.getMaxValue().toLong() );
+            assertEquals( String.format( "locator %s min value is the same", locator ),
+                    expectedMetric.getMetricValue(),
+                    rollup.getMinValue().toLong() );
+        }
+    }
+
+    // This is maintaining broken functionality from astyanax.  You can't put a rollup into metrics_full,
+    // but astyanax allows you to.  However, when you attempt to read it, you get 0 points.
```
shintasmith: Lol!

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DRawIO.java
usnavi: the "DRawIO.insertAsync()" is only called one place, ```DBasicMetricsRW.insertMetrics()```.  I think the numeric vs string check will have to happen somewhere, I think keeping the functionality encapsulated in this class makes sense.

blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/io/BasicMetricsRWAsytanaxWriteIntegrationTest.java
usnavi: @shintasmith  - You couldn't handle my full potential

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DAbstractMetricsRW.java
usnavi: The Asytanax side throws the IOException, but not here.  I'll remove it.

ChandraAddala: how about throwing it?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DRawIO.java
usnavi: My vote is to keep them the same as it encapsulates the Numeric vs String check in ```insertAsync```.  Otherwise I think we'd have to pull that check up, which doesn't really help with readability.

shintasmith: Ok. It's fine for this one. The reason I brought it up is because I like the classes to have specific responsibility. But since String is going away anyways.... I won't fuss too much. :-)

usnavi:
@shintasmith @ChandraAddala - updated PR as per comments.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/astyanax/AstyanaxWriter.java
```diff
@@ -166,12 +166,13 @@ private final void insertEnumValuesWithHashcodes(Locator locator, BluefloodEnumR
...
-        final boolean isString = metric.isString();
-        final boolean isBoolean = metric.isBoolean();
+    private void insertMetric(IMetric metric, MutationBatch mutationBatch) {
+        final boolean isString = DataType.isStringMetric( metric.getMetricValue() );
+        final boolean isBoolean = DataType.isBooleanMetric( metric.getMetricValue() );
 
         if (isString || isBoolean) {
-            metric.setTtl(STRING_TTL);
+            // they were already casting long to int in Metrics.setTtl()
+            metric.setTtlInSeconds( (int) STRING_TTL.toSeconds() );
```
shintasmith: Why change the call, tho? Shouldn't we just leave it to ```metric.setTtl()``` ?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DAbstractMetricsRW.java
```diff
@@ -0,0 +1,259 @@
...
+     * @return
+     */
+    protected Map<Locator, MetricData> getDatapointsForRange( List<Locator> locators,
+                                                              Range range,
+                                                              String columnFamily,
+                                                              Granularity granularity ) {
+        // in this loop, we will fire all the executeAsync() of
+        // various select statements, the collect all of the
+        // ResultSetFutures
+        Map<Locator, List<ResultSetFuture>> locatorToFuturesMap = new HashMap<Locator, List<ResultSetFuture>>();
```
shintasmith: I just noticed.... Are we missing ```Instrumentation.getReadTimerContext(columnFamilyName)``` here?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DAbstractMetricsRW.java
```diff
@@ -0,0 +1,259 @@
...
+
+            Table<Locator, Long, Object> locatorTimestampRollup = io.getRollupsForLocator(locator, columnFamilyName, range);
+
+            // transform them to Points
+            Points points = new Points();
+            for (Table.Cell<Locator, Long, Object> cell : locatorTimestampRollup.cellSet()) {
+                points.add( createPoint( cell.getColumnKey(), cell.getValue()));
+            }
+            return points;
+        } finally {
```
shintasmith: do we need to do a generic ```catch(Exception ex)``` so we can call ```Instrumentation.markReadError()``` ?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DBasicMetricsRW.java
```diff
@@ -0,0 +1,362 @@
...
+                catch ( Exception e ) {
+                    Instrumentation.markWriteError();
+                    LOG.error(String.format("error writing metric for locator %s",
+                            f.getKey()), e );
+                }
+            }
+        }
+        finally {
+
+            ctx.close();
```
shintasmith: Curious, why ```close()``` and not ```stop()```? Everywhere else we call ```stop()```, I think.... And yes, I know they do the same thing. :-)

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DAbstractMetricsRW.java
shintasmith: Never mind... it looks like the caller puts the Instrumentation... I wonder if we should change that...

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DBasicMetricsRW.java
```diff
@@ -0,0 +1,362 @@
...
+     * than the other granularities.
+     *
+     * @param rollupType
+     * @param granularity
+     * @return
+     */
+    @Override
+    public DAbstractMetricIO getIO( String rollupType, Granularity granularity ) {
+
+      if( granularity == Granularity.FULL )
```
shintasmith: Beautiful! 👍

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/astyanax/AstyanaxWriter.java
usnavi: I change the method to accept IMetric, and that has setTtlInSeconds().

shintasmith: ah ok, that's fine, then.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DAbstractMetricsRW.java
usnavi: Yeah, let's update that.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DAbstractMetricsRW.java
usnavi: Looks like we don't do it on the astyanax-side, but it wouldn't hurt.  I'll add it.

usnavi:
@shintasmith @ChandraAddala  - updated the branch

shintasmith:
lgtm 👍

ChandraAddala:
Looks good.

PR #693, opened 2016-05-10
Modify json parsing exceptions catching

VinnyQ:
- WHAT
Modify response body message when empty json payload is received.  This functionality was modified when the 207 response code was introduced in the previous commits, which change the location of json parsing but not where the exception is caught.

- HOW
Since json processing is changed from previous JSONMetricsContainer.toMetrics() call to just process when the class is initialized (i.e. in the constructor), all json parsing and exceptions catching should have moved with it to the property try-catch location.  I did so by merging the two separated blocks into one, and checking if the jsonMetrics list isEmpty, which was previously inferred by the toMetrics() call during json parsing.

### DIFF
--- a/blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpAggregatedMultiIngestionHandler.java
+++ b/blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpAggregatedMultiIngestionHandler.java
@@ -112,16 +112,19 @@ public class HttpAggregatedMultiIngestionHandler implements HttpRequestHandler {
                 if (errors.isEmpty()) {
                     // no validation error, response OK
                     DefaultHandler.sendResponse(ctx, request, null, HttpResponseStatus.OK);
+                    return;
                 }
                 else {
                     // has some validation errors, response MULTI_STATUS
                     DefaultHandler.sendResponse(ctx, request, null, HttpResponseStatus.MULTI_STATUS);
+                    return;
                 }
 
             }
             else {
-                // no aggregated metric bundles in body
-                DefaultHandler.sendResponse( ctx, request, null, HttpResponseStatus.BAD_REQUEST );
+                // no aggregated metric bundles in body, response OK
+                DefaultHandler.sendResponse(ctx, request, null, HttpResponseStatus.OK);
+                return;
             }
         } catch (JsonParseException ex) {
             log.debug(String.format("BAD JSON: %s", body));
--- a/blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionHandler.java
+++ b/blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionHandler.java
@@ -99,7 +99,9 @@ public class HttpMetricsIngestionHandler implements HttpRequestHandler {
 
             requestCount.inc();
             final String tenantId = request.getHeader("tenantId");
-            JSONMetricsContainer jsonMetricsContainer = null;
+            JSONMetricsContainer jsonMetricsContainer;
+            List<Metric> metrics;
+
             final Timer.Context jsonTimerContext = jsonTimer.time();
 
             final String body = request.getContent().toString(Constants.DEFAULT_CHARSET);
@@ -107,10 +109,21 @@ public class HttpMetricsIngestionHandler implements HttpRequestHandler {
                 jsonMetricsContainer = createContainer(body, tenantId);
 
                 if (jsonMetricsContainer == null) {
-                    log.warn(ctx.getChannel().getRemoteAddress() + " No valid metrics");
+                    log.warn(ctx.getChannel().getRemoteAddress() + " Failed to create jsonMetricsContainer.");
                     DefaultHandler.sendResponse(ctx, request, "No valid metrics", HttpResponseStatus.BAD_REQUEST);
                     return;
+                } else if (jsonMetricsContainer.getJsonMetrics().isEmpty()) {
+                    log.warn(ctx.getChannel().getRemoteAddress() + " Json metrics is empty, no valid json metrics to parse.");
+                    DefaultHandler.sendResponse(ctx, request, "Error converting JSON payload to metric objects", HttpResponseStatus.BAD_REQUEST);
+                    return;
                 }
+
+                if (jsonMetricsContainer.areDelayedMetricsPresent()) {
+                    Tracker.getInstance().trackDelayedMetricsTenant(tenantId, jsonMetricsContainer.getDelayedMetrics());
+                }
+
+                metrics = jsonMetricsContainer.getValidMetrics();
+                forceTTLsIfConfigured(metrics);
 
             } catch (JsonParseException e) {
                 log.warn("Exception parsing content", e);
@@ -120,6 +133,12 @@ public class HttpMetricsIngestionHandler implements HttpRequestHandler {
                 log.warn("Exception parsing content", e);
                 DefaultHandler.sendResponse(ctx, request, "Cannot parse content", HttpResponseStatus.BAD_REQUEST);
                 return;
+            } catch (InvalidDataException ex) {
+                // todo: we should measure these. if they spike, we track down the bad client.
+                // this is strictly a client problem. Someting wasn't right (data out of range, etc.)
+                log.warn(ctx.getChannel().getRemoteAddress() + " " + ex.getMessage());
+                DefaultHandler.sendResponse(ctx, request, "Invalid data " + ex.getMessage(), HttpResponseStatus.BAD_REQUEST);
+                return;
             } catch (IOException e) {
                 log.warn("IO Exception parsing content", e);
                 DefaultHandler.sendResponse(ctx, request, "Cannot parse content", HttpResponseStatus.BAD_REQUEST);
@@ -127,31 +146,6 @@ public class HttpMetricsIngestionHandler implements HttpRequestHandler {
             } catch (Exception e) {
                 log.warn("Other exception while trying to parse content", e);
                 DefaultHandler.sendResponse(ctx, request, "Failed parsing content", HttpResponseStatus.INTERNAL_SERVER_ERROR);
-                return;
-            }
-
-            List<Metric> metrics;
-            try {
-                metrics = jsonMetricsContainer.getValidMetrics();
-                forceTTLsIfConfigured(metrics);
-
-                if (jsonMetricsContainer.areDelayedMetricsPresent()) {
-                    Tracker.getInstance().trackDelayedMetricsTenant(tenantId, jsonMetricsContainer.getDelayedMetrics());
-                }
-            } catch (InvalidDataException ex) {
-                // todo: we should measure these. if they spike, we track down the bad client.
-                // this is strictly a client problem. Someting wasn't right (data out of range, etc.)
-                log.warn(ctx.getChannel().getRemoteAddress() + " " + ex.getMessage());
-                DefaultHandler.sendResponse(ctx, request, "Invalid data " + ex.getMessage(), HttpResponseStatus.BAD_REQUEST);
-                return;
-            } catch (Exception e) {
-                // todo: when you see these in logs, go and fix them (throw InvalidDataExceptions) so they can be reduced
-                // to single-line log statements.
-                log.warn("Exception converting JSON container to metric objects", e);
-                // This could happen if clients send BigIntegers as metric values. BF doesn't handle them. So let's send a
-                // BAD REQUEST message until we start handling BigIntegers.
-                DefaultHandler.sendResponse(ctx, request, "Error converting JSON payload to metric objects",
-                        HttpResponseStatus.BAD_REQUEST);
                 return;
             } finally {
                 jsonTimerContext.stop();
@@ -183,10 +177,12 @@ public class HttpMetricsIngestionHandler implements HttpRequestHandler {
                 if( !errors.isEmpty() ) {
                     // has some validation errors, return MULTI_STATUS
                     DefaultHandler.sendResponse(ctx, request, null, HttpResponseStatus.MULTI_STATUS);
+                    return;
                 }
                 else {
                     // no validation error, return OK
                     DefaultHandler.sendResponse(ctx, request, null, HttpResponseStatus.OK);
+                    return;
                 }
 
             } catch (TimeoutException e) {
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/formats/AggregatedPayload.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/formats/AggregatedPayload.java
@@ -19,7 +19,6 @@ package com.rackspacecloud.blueflood.inputs.formats;
 import com.rackspacecloud.blueflood.service.Configuration;
 import com.rackspacecloud.blueflood.service.CoreConfig;
 import com.rackspacecloud.blueflood.types.*;
-import org.codehaus.jackson.annotate.JsonIgnore;
 
 import java.util.*;
 
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/formats/JSONMetricsContainer.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/formats/JSONMetricsContainer.java
@@ -29,16 +29,13 @@ import java.util.List;
 import java.util.concurrent.TimeUnit;
 
 public class JSONMetricsContainer {
-    private final String tenantId;
-    private final List<JSONMetric> jsonMetrics;
 
     private static final long TRACKER_DELAYED_METRICS_MILLIS = Configuration.getInstance().getLongProperty(CoreConfig.TRACKER_DELAYED_METRICS_MILLIS);
     private static final long MAX_AGE_ALLOWED = Configuration.getInstance().getLongProperty(CoreConfig.ROLLUP_DELAY_MILLIS);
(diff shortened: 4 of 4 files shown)


blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpAggregatedIngestionHandler.java
```diff
@@ -99,10 +99,10 @@ public void handle(ChannelHandlerContext ctx, HttpRequest request) {
                         HttpResponseStatus.BAD_REQUEST );
             }
 
-        } catch (JsonParseException ex) {
+        } catch (JsonSyntaxException ex) {
```
VinnyQ: Note, for some reason, we are using Jackon for parsing json in *HttpMetricsIngestionHandler*, but Gson for parsing json in *HttpAggregatedIngestionHandler* and *HttpAggregatedMultiIngestionHandler*.

So this exception catching has been wrong because since the Gson class which throws JsonSyntaxException, not JsonParseException (thrown by codehaus.Jackson.TypeFactory).  We were never going to catch any JsonParseException here, so it's valid but a useless catch statement.

We now have better json parsing error message for aggregated metrics ingestion, and not just a generic "Failed parsing content" as the response from the catch-all exception.

VinnyQ:
@shintasmith @usnavi @ChandraAddala a small fix for exceptions catching in the handlers, please review.  And +1 if the tests/checks are good.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionHandler.java
```diff
@@ -120,38 +133,19 @@ public void handle(ChannelHandlerContext ctx, HttpRequest request) {
...
-            }
-
-            List<Metric> metrics;
-            try {
-                metrics = jsonMetricsContainer.getValidMetrics();
-                forceTTLsIfConfigured(metrics);
-
-                if (jsonMetricsContainer.areDelayedMetricsPresent()) {
-                    Tracker.getInstance().trackDelayedMetricsTenant(tenantId, jsonMetricsContainer.getDelayedMetrics());
-                }
```
VinnyQ: Note here that I have combined all the catch statements into try with all the catches, since we're processing the json as we create the container, and not have a separate json parsing and then a toMetrics() call to process like before.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionHandler.java
```diff
@@ -99,19 +99,32 @@ public void handle(ChannelHandlerContext ctx, HttpRequest request) {
...
+
             final Timer.Context jsonTimerContext = jsonTimer.time();
 
             final String body = request.getContent().toString(Constants.DEFAULT_CHARSET);
             try {
                 jsonMetricsContainer = createContainer(body, tenantId);
 
                 if (jsonMetricsContainer == null) {
-                    log.warn(ctx.getChannel().getRemoteAddress() + " No valid metrics");
+                    log.warn(ctx.getChannel().getRemoteAddress() + " No valid json for container.");
```
usnavi: I'd use "No valid json content"  I wouldn't use the term "container".

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionHandler.java
```diff
@@ -99,19 +99,32 @@ public void handle(ChannelHandlerContext ctx, HttpRequest request) {
...
                     DefaultHandler.sendResponse(ctx, request, "No valid metrics", HttpResponseStatus.BAD_REQUEST);
                     return;
+                } else if (jsonMetricsContainer.getJsonMetrics().isEmpty()) {
+                    log.warn(ctx.getChannel().getRemoteAddress() + " No valid json metrics to parse.");
+                    DefaultHandler.sendResponse(ctx, request, "Error converting JSON payload to metric objects", HttpResponseStatus.BAD_REQUEST);
+                    return;
```
usnavi: Just to clarify - I thought if they didn't send any metrics we were going to send a 200.  Was that for a different scenario?

VinnyQ: According to the current tests, for regular metrics if they valid but lame json, e.g. '[]' or '{}', it should return a 400 error code with that error message.

But for aggregated metrics, if they send empty array, then it's 200.  Do you want me to modify it so that they're both consistent one way or the other?

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionHandler.java
VinnyQ: Note, this is just a log entry for our own debugging, not a response.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionHandler.java
usnavi: I think we should maintain current behavior and create a story outlining the discrepancies between normal and preaggregated metrics.  Include it in the epic "API cleanup and validation".

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpAggregatedIngestionHandler.java
VinnyQ: Ok never mind, while the statement is true that we're using two different json parsing library (Jackson for regular metrics and Gson for aggregated metrics), it's ok to leave this as JsonParseException, since JsonSyntaxException extends it, so when it is throw by gson it still catches it here.

It is however still true that while Jackson are ok with the string '[]', which is valid and we only caught and return 400 because of having no metrics return from the mapping, Gson does not like seeing '[]' at all for non-array expecting metric (i.e. if it's trying to map to a single aggregated metric), and will throw the error "java.lang.IllegalStateException: Expected BEGIN_OBJECT but was BEGIN_ARRAY at line 1 column ".

I'll create a story as a tech debt to choose one or the other to use (Jackson or Gson) and refactor the code so it's more consistent with errors and validations.

VinnyQ:
Here's a complete list of the error code and response body for invalid or empty json BEFORE the 207 story changes.  Note the inconsistency between single and multi ingestion of the same metric types, and the error messages between the two different metrics types.  This PR will modify the current blueflood to match this exact behavior for now.

I'll include these code in the story to normalize the json parsing lib and error responses, and we can change them to be more consistent then.

```
# should NOT be able send empty array: POST /v2.0/{tenantId}/ingest/multi: #endpoint.description
curl -i -X POST 'http://localhost:19000/v2.0/836986/ingest' -d ''
>> 400 Cannot parse content

curl -i -X POST 'http://localhost:19000/v2.0/836986/ingest' -d '{'
>> 400 Cannot parse content

curl -i -X POST 'http://localhost:19000/v2.0/836986/ingest' -d '{}'
>> 400 Cannot parse content

curl -i -X POST 'http://localhost:19000/v2.0/836986/ingest' -d '[]'
>> 400 Error converting JSON payload to metric objects

curl -i -X POST 'http://localhost:19000/v2.0/836986/ingest' -d '[{}]'
>> 400 'null': 'collectionTime' '0' is more than '259200000' milliseconds into the past.

curl -i -X POST 'http://localhost:19000/v2.0/836986/ingest' -d '{"asdf}'
>> 400 Cannot parse content

curl -i -X POST 'http://localhost:19000/v2.0/836986/ingest/multi' -d ''
>> 400 Cannot parse content

curl -i -X POST 'http://localhost:19000/v2.0/836986/ingest/multi' -d '{'
>> 400 Cannot parse content

curl -i -X POST 'http://localhost:19000/v2.0/836986/ingest/multi' -d '{}'
>> 400 Cannot parse content

curl -i -X POST 'http://localhost:19000/v2.0/836986/ingest/multi' -d '[]'
>> 400 Error converting JSON payload to metric objects

curl -i -X POST 'http://localhost:19000/v2.0/836986/ingest/multi' -d '[{}]'
>> 400 
>> The following errors have been encountered:
>> 'null': 'collectionTime' '0' is more than '259200000' milliseconds into the past.
>> 'null': No tenantId is provided for the metric.

curl -i -X POST 'http://localhost:19000/v2.0/836986/ingest/multi' -d '[{"asdf}]'
>> 400 Cannot parse content


curl -i -X POST 'http://localhost:19000/v2.0/836986/ingest/aggregated' -d ''
>> 500 Failed parsing content

curl -i -X POST 'http://localhost:19000/v2.0/836986/ingest/aggregated' -d '{'
>> 400 java.io.EOFException: End of input at line 1 column 2

curl -i -X POST 'http://localhost:19000/v2.0/836986/ingest/aggregated' -d '{}'
>> 400
>> The following errors have been encountered:
>> 'timestamp' '0' is more than '259200000' milliseconds into the past.

curl -i -X POST 'http://localhost:19000/v2.0/836986/ingest/aggregated' -d '[]'
>> 400 java.lang.IllegalStateException: Expected BEGIN_OBJECT but was BEGIN_ARRAY at line 1 column 259200000

curl -i -X POST 'http://localhost:19000/v2.0/836986/ingest/aggregated' -d '[{}]'
>> 400 java.lang.IllegalStateException: Expected BEGIN_OBJECT but was BEGIN_ARRAY at line 1 column 2

curl -i -X POST 'http://localhost:19000/v2.0/836986/ingest/aggregated' -d '{"asdf}'
>> 400 com.google.gson.stream.MalformedJsonException: Unterminated string at line 1 column 8

curl -i -X POST 'http://localhost:19000/v2.0/836986/ingest/aggregated/multi' -d ''
>> 500 Failed parsing content

curl -i -X POST 'http://localhost:19000/v2.0/836986/ingest/aggregated/multi' -d '{'
>> 500 Failed parsing content

curl -i -X POST 'http://localhost:19000/v2.0/836986/ingest/aggregated/multi' -d '{}'
>> 500 Failed parsing content

curl -i -X POST 'http://localhost:19000/v2.0/836986/ingest/aggregated/multi' -d '[]'
>> 200 OK

curl -i -X POST 'http://localhost:19000/v2.0/836986/ingest/aggregated/multi' -d '[{}]'
>> 400
>> The following errors have been encountered:
>> 'timestamp' '0' is more than '259200000' milliseconds into the past.

curl -i -X POST 'http://localhost:19000/v2.0/836986/ingest/aggregated/multi' -d '[{"asdf}]'
>> 400
>> com.google.gson.stream.MalformedJsonException: Unterminated string at line 1 column 10
```

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpAggregatedIngestionHandler.java
shintasmith: Are you going to put back the ```catch (JsonParseException ex)``` ?

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpAggregatedIngestionHandler.java
```diff
@@ -99,10 +99,10 @@ public void handle(ChannelHandlerContext ctx, HttpRequest request) {
...
             log.debug(String.format("BAD JSON: %s", body));
             log.error(ex.getMessage(), ex);
-            DefaultHandler.sendResponse(ctx, request, ex.getMessage(), HttpResponseStatus.BAD_REQUEST);
+            DefaultHandler.sendResponse(ctx, request, "Failed parsing Json content: " + ex.getMessage(), HttpResponseStatus.BAD_REQUEST);
```
shintasmith: We have to be careful here that the exception message does not make the response an invalid Json. For example, if the exception message contains quotes, it will have to be escaped.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpAggregatedIngestionHandler.java
shintasmith: I think we should go back to catching ```JsonParseException``` because there are 2 exceptions subclassing ```JsonParseException```: one is ```JsonSyntaxException```, the other is ```JsonIOException```. We want to be able to catch both.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpAggregatedIngestionHandler.java
```diff
@@ -111,14 +111,14 @@ public void handle(ChannelHandlerContext ctx, HttpRequest request) {
...
-            DefaultHandler.sendResponse(ctx, request, "Failed parsing content", HttpResponseStatus.INTERNAL_SERVER_ERROR);
+            DefaultHandler.sendResponse(ctx, request, "Failed parsing content: " + ex.getMessage(), HttpResponseStatus.INTERNAL_SERVER_ERROR);
         } finally {
             requestCount.dec();
             timerContext.stop();
         }
     }
     
-    public static AggregatedPayload createPayload(String json) {
+    public static AggregatedPayload createPayload(String json) throws JsonSyntaxException {
```
shintasmith: We should use ```JsonParseException``` here if you want to declare the exception.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpAggregatedMultiIngestionHandler.java
```diff
@@ -121,12 +121,12 @@ public void handle(ChannelHandlerContext ctx, HttpRequest request) {
             }
             else {
                 // no aggregated metric bundles in body
-                DefaultHandler.sendResponse( ctx, request, null, HttpResponseStatus.BAD_REQUEST );
+                DefaultHandler.sendResponse( ctx, request, "Error converting JSON payload to metric objects", HttpResponseStatus.BAD_REQUEST );
             }
-        } catch (JsonParseException ex) {
+        } catch (JsonSyntaxException ex) {
```
shintasmith: I vote for catching ```JsonParseException``` here

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpAggregatedMultiIngestionHandler.java
```diff
@@ -135,14 +135,14 @@ public void handle(ChannelHandlerContext ctx, HttpRequest request) {
...
-            DefaultHandler.sendResponse(ctx, request, "Failed parsing content", HttpResponseStatus.INTERNAL_SERVER_ERROR);
+            DefaultHandler.sendResponse(ctx, request, "Failed parsing content: " + ex.getMessage(), HttpResponseStatus.INTERNAL_SERVER_ERROR);
         } finally {
             requestCount.dec();
             timerContext.stop();
         }
     }
 
-    public static List<AggregatedPayload> createBundleList(String json) {
+    public static List<AggregatedPayload> createBundleList(String json) throws JsonSyntaxException {
```
shintasmith: JsonParseException

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionHandler.java
```diff
@@ -99,19 +99,32 @@ public void handle(ChannelHandlerContext ctx, HttpRequest request) {
...
             final Timer.Context jsonTimerContext = jsonTimer.time();
 
             final String body = request.getContent().toString(Constants.DEFAULT_CHARSET);
             try {
                 jsonMetricsContainer = createContainer(body, tenantId);
 
                 if (jsonMetricsContainer == null) {
-                    log.warn(ctx.getChannel().getRemoteAddress() + " No valid metrics");
+                    log.warn(ctx.getChannel().getRemoteAddress() + " No valid json content.");
                     DefaultHandler.sendResponse(ctx, request, "No valid metrics", HttpResponseStatus.BAD_REQUEST);
```
shintasmith: If you are changing the log.warn(), why don't we change the response as well, to match?

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionHandler.java
shintasmith: I'm fine with the combined catch statements. But will this fix the failing system regression tests, I wonder?

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionHandler.java
VinnyQ: It was to distinguish between the two errors in the log and to make the first log reflect more accurately why it was an error, but i didn't want to change the response message and keep all responses the same for now.  The else handles a new state we didn't have before here because we now process json on container creation and not with a separate method call later, so we check for empty json metrics here.

I changed both logs now to reflect the errors it much better.

VinnyQ:
@shintasmith @usnavi Ok I reverted most of the code so that it response exactly as it does now regarding invalid and empty json and other errors.  There are still code changes require to make the responses the same for some of the cases because we moved where we are processing json and converting it to metrics, so the changes here are exactly for that, plus a few minor clean up of used code.

shintasmith:
lgtm 👍 Have you run the failing system test against this code? Is it passing with this code changes?

VinnyQ:
yes i have, and the tests passes!

PR #694, opened 2016-05-20
Fix intermittent errors with ROR tests

VinnyQ:
- WHAT

Uncomment and fix intermittent assert errors in RollupHandlerIntegrationTest.

- HOW
 * Fix bug with calculate full mean using long value sum and division, which were dropping of whole decimal places during calculation.

 * Fix bug in testMplotRollupsOnReadGenerationRight() using the wrong start value (should be ``startRollupMS`` instead of ``startMs``)

 * Modify all FULL datapoints query for mean calculation to assert with to use MIN_5. snapMillis to match the RollupHandler range calculation during rollup.  Data loss still present when mean is calculated during rollup.

### DIFF
--- a/blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/outputs/RollupHandlerIntegrationTest.java
+++ b/blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/outputs/RollupHandlerIntegrationTest.java
@@ -64,6 +64,11 @@ public class RollupHandlerIntegrationTest extends IntegrationTestBase {
     private long startRollupMS = startMS + (1000 * 60 * 60 * (hours/2 - 5));
     private long endRollupMS = startMS + (1000 * 60 * 60 * (hours/2));
 
+    // max possible value for random generator of metric values
+    private final int MAX_METRIC_VALUE = 100;
+    // epsilon for assert of differences between mean of full points and mean of coarser granularity (rolled) points
+    private final double MEAN_EPSILON = 0.0025 * MAX_METRIC_VALUE;
+
     @Before
     public void initData() throws Exception {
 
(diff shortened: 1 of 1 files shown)


VinnyQ:
@usnavi ok let's start a discussion.  After a few tweaks I am able to set the epsilon @ 0.0025 (a quarter of 1%) for the possible MAX_VALUE of the random values being generated (with it being 100 currently).

Let me know what you think!

usnavi:
@VinnyQ  - yeah, looks like the snapMillis() is what was required.  lgtm.  thanks!

VinnyQ:
@usnavi well, both the snapMillis and converting the mean calculation to using double values instead of long, which was causing a lot of data lost over time for the full mean calculation (I caught it because it always returned a whole number for the mean every time the test is run, even with 2880 points of random values in the range of 0-100).  Converting the sum from long to double in ``meanOfPointCollectionRoll()`` fixed that.  

There was also one test that was using different start time for the rolled up points retrieval and full points retrieval (``testMplotRollupsOnReadGenerationRight()``).  That was causing radically huge discrepancy randomly.

blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/outputs/RollupHandlerIntegrationTest.java
```diff
@@ -64,6 +64,11 @@
     private long startRollupMS = startMS + (1000 * 60 * 60 * (hours/2 - 5));
     private long endRollupMS = startMS + (1000 * 60 * 60 * (hours/2));
 
+    // max possible value for random generator of metric values
+    private final int MAX_METRIC_VALUE = 100;
+    // epsilon for assert of differences between mean of full points and mean of coarser granularity (rolled) points
+    private final double MEAN_EPSILON = 0.0025 * MAX_METRIC_VALUE;
```
ChandraAddala: How did you arrive at 0.0025? Can you add some documentation?

ChandraAddala:
Other than the comment I made, everything looks good.

PR #695, opened 2016-05-25
Docker compose fix

goru97:
Fixed CQL version mismatch.

### DIFF
--- a/contrib/blueflood-docker/docker-compose.yml
+++ b/contrib/blueflood-docker/docker-compose.yml
@@ -9,14 +9,14 @@ blueflood:
         - "9180:9180"
 
 cassandra:
-  image: cassandra:2.2.6
+  image: cassandra:2.1
   ports:
         - "9042:9042"
         - "9160:9160"
         - "7199:7199"
 
 cassandran:
-  image: cassandra:2.2.6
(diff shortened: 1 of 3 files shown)


goru97:
@shintasmith Take a look!

contrib/blueflood-docker/docker-compose.yml
```diff
@@ -1,5 +1,5 @@
 blueflood:
-  build: .
+  image: goru97/blueflood
```
usnavi: Should we have user names in our example?

goru97: @usnavi @shintasmith had the same concern in https://github.com/rackerlabs/blueflood/pull/696. I reached out to #rackspace #respose channels on IRC, perused through one wiki and also e-mailed dev and devops list but nobody got the credentials for rackerlabs docker-hub account. usernames are part of the docker images if they are hosted on your docker hub account. They are not required if they are hosted by docker official account e.g. cassandra, elasticsearch.

usnavi: OK, let's discuss more when you return.  Have a good break!

usnavi:
Included in #696

PR #697, opened 2016-05-25
Added configuration to enable sub metrics

wolfdancer:
This will enable uptime reporting related
dashboard through Grafana.

### DIFF
--- a/contrib/Docker_Grafana_Blueflood/docker-entrypoint.sh
+++ b/contrib/Docker_Grafana_Blueflood/docker-entrypoint.sh
@@ -37,6 +37,8 @@ blueflood:
   apikey: $RAX_APIKEY                     
   authentication_module: rax_auth
   authentication_class: BluefloodAuth
+  enable_submetrics: true
+  submetric_aliases: { "_sum": 'sum' }
   urls:
     - $BLUEFLOOD_QUERY_URL
 EOL


contrib/Docker_Grafana_Blueflood/docker-entrypoint.sh
```diff
@@ -37,6 +37,8 @@ blueflood:
   apikey: $RAX_APIKEY                     
   authentication_module: rax_auth
   authentication_class: BluefloodAuth
+  enable_submetrics: true
+  submetric_aliases: { "_sum": 'sum' }
```
shintasmith: Should we enable the other submetrics too? Or are we only interested in solving the API Availability use case?

wolfdancer: I am only thinking about the uptime reporting.  I am still testing it out to see how it works out.

We can add others when we find good use case for them.

shintasmith:
lgtm 👍

contrib/Docker_Grafana_Blueflood/docker-entrypoint.sh
ChandraAddala: @wolfdancer  If you only enable "sum", then thats the only submetric possible.  If someone wants to do "average" which has been the default so far, I dont believe its possible unless its added here.

wolfdancer: @ChandraAddala right. This PR was more narrowly focused on the uptime reporting. I think with Vinny's research on StatsD metrics support we can turn on the rest of the submetrics in another PR.

ChandraAddala: @wolfdancer I am just concerned that if someone spins up blueflood and starts playing with it, they would have difficulty with graphs as we dont have any documentation about submetrics and this is not regular behavior of grafana.

wolfdancer: @ChandraAddala I see.  I didn't realize that it is not regular behavior of Grafana.  For this PR, the goal is to provide a replacement for the Enum support that we are going to take out.  So does this mean that we need to document the submetrics (and make releases if needed) before we can promote a better way for uptime reporting?

PR #702, opened 2016-06-15
Docker revamp

goru97:
# What:
Revamping Blueflood Docker Image.

# Why:
To give it a standardize form, make it more pliable and production ready (can be used as rollup/ingest/query node) so that it can be officially released on docker-hub as a part of rackerlabs account.
Also, keeping my promise to @usnavi


usnavi:
Need this merged to generate an image for rackerlabs docker hub.

PR #703, opened 2016-06-16
Change logging

shintasmith:
Change a few logging:
* when we are ingesting metrics and it takes more than 5 seconds, we log ERROR even though it's not an error. Metrics are still being ingested, they just take longer than 5 seconds. Change this logging to INFO
* added more logging when we return 500 to clients, we log which metrics failed


ChandraAddala:
@shintasmith I think its 5 seconds, not 5 mins.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/processors/BatchWriter.java
```diff
@@ -122,10 +122,13 @@ public Boolean call() throws Exception {
...
-                        if (System.currentTimeMillis() - writeStartTime > timeout.toMillis()) {
+                        long now = System.currentTimeMillis();
+                        if ( now - writeStartTime > timeout.toMillis()) {
                             exceededScribeProcessingTime.mark();
-                            getLogger().error("Exceeded timeout " + timeout.toString() + " before persisting " +
-                                    "all metrics for batch " + batchId);
+                            getLogger().info(
+                                    String.format("Batch write time %d (ms) exceeded timeout %s before persisting " +
+                                            "all metrics for batch %d",
+                                            now - writeStartTime, timeout.toString(), batchId));
```
ChandraAddala: I agree that its not an error. But somebody thought its really bad that we are exceeding that timeout. Any idea what this timeout actually is?

shintasmith: I have no idea... It seems arbitrary. It is hard coded inside HttpMetricsIngestionServer

ChandraAddala: I did look into it a little bit. I vote for changing it to debug. There is meter associated with it. I plotted it and saved on ingest dashboard. Incase we wanted to find out if ingestion is taking more than 5 seconds, we can look at that dashboard.

http://graphite00.iad.prod.bf.k1k.me/grafana/#/dashboard/db/ingest-dashboard?panelId=31&fullscreen&edit

shintasmith: I am ok with debug. I will change it.

ChandraAddala:
Looks good.

shintasmith:
@ChandraAddala , I changed the logging in BatchWriter to debug()

@usnavi , @VinnyQ , @izrik , pls review

ChandraAddala:
Looks good.

PR #704, opened 2016-06-16
Refactor TTL providers

DanielNi:
WHAT: This PR combines the TTL providers into a single class `CombinedTtlProvider.java` and changes the `TenantTtlProvider.java` interface slightly so that it is less complicated.

WHY: Cleans up the TTL provider code a bit and adds some testing.

HOW: 
- Added the `CombinedTtlProvider.java` class
- `TenantTtlProvider.java` now returns `Optional`, which is easier to reason with than exceptions
- Removed `set()` method in the providers since it is not currently used and throws an exception
- Moved the 5x TTL logic from `SafetyTtlProvider.java` to `CassandraModel.java` to make it less confusing
- Made the TTL provider tests a bit more robust
- Removed the Hist TTL configs since they are no longer being used

### DIFF
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/cache/CombinedTtlProvider.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/cache/CombinedTtlProvider.java
@@ -48,14 +48,14 @@ public class CombinedTtlProvider implements TenantTtlProvider {
     public Optional<TimeValue> getTTL(String tenantId, Granularity gran, RollupType rollupType) {
         Optional<TimeValue> primaryValue = primary.getTTL(tenantId, gran, rollupType);
         Optional<TimeValue> safetyValue = safety.getTTL(tenantId, gran, rollupType);
-        return getMaximum(primaryValue, safetyValue);
+        return getTimeValue(primaryValue, safetyValue);
     }
 
     @Override
     public Optional<TimeValue> getTTLForStrings(String tenantId) {
         Optional<TimeValue> primaryValue = primary.getTTLForStrings(tenantId);
         Optional<TimeValue> safetyValue = safety.getTTLForStrings(tenantId);
-        return getMaximum(primaryValue, safetyValue);
+        return getTimeValue(primaryValue, safetyValue);
     }
 
     public long getFinalTTL(String tenantid, Granularity gran) {
@@ -68,7 +68,7 @@ public class CombinedTtlProvider implements TenantTtlProvider {
         return ttl;
     }
 
-    private static Optional<TimeValue> getMaximum(Optional<TimeValue> primaryValue, Optional<TimeValue> safetyValue) {
+    private static Optional<TimeValue> getTimeValue(Optional<TimeValue> primaryValue, Optional<TimeValue> safetyValue) {
         if (!primaryValue.isPresent()) {
             return safetyValue;
         }
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/cache/CombinedTtlProvider.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/cache/CombinedTtlProvider.java
@@ -1,0 +1,72 @@
+/*
+ * Copyright 2016 Rackspace
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+package com.rackspacecloud.blueflood.cache;
+
+import com.google.common.base.Optional;
+import com.rackspacecloud.blueflood.rollup.Granularity;
+import com.rackspacecloud.blueflood.types.RollupType;
+import com.rackspacecloud.blueflood.utils.TimeValue;
+
+/**
+ * TTL Provider which obtains the TTL from the primary or secondary provider.
+ */
+public class CombinedTtlProvider implements TenantTtlProvider {
+    private final ConfigTtlProvider primary = ConfigTtlProvider.getInstance();
+    private final SafetyTtlProvider safety = SafetyTtlProvider.getInstance();
+
+    private static final CombinedTtlProvider INSTANCE = new CombinedTtlProvider();
+
+    public static CombinedTtlProvider getInstance() {
+        return INSTANCE;
+    }
+
+    @Override
+    public Optional<TimeValue> getTTL(String tenantId, Granularity gran, RollupType rollupType) {
+        Optional<TimeValue> primaryValue = primary.getTTL(tenantId, gran, rollupType);
+        Optional<TimeValue> safetyValue = safety.getTTL(tenantId, gran, rollupType);
+        return getMaximum(primaryValue, safetyValue);
+    }
+
+    @Override
+    public Optional<TimeValue> getTTLForStrings(String tenantId) {
+        Optional<TimeValue> primaryValue = primary.getTTLForStrings(tenantId);
+        Optional<TimeValue> safetyValue = safety.getTTLForStrings(tenantId);
+        return getMaximum(primaryValue, safetyValue);
+    }
+
+    @Override
+    public Optional<TimeValue> getConfigTTLForIngestion() {
+        Optional<TimeValue> primaryValue = primary.getConfigTTLForIngestion();
+        Optional<TimeValue> safetyValue = safety.getConfigTTLForIngestion();
+        return getMaximum(primaryValue, safetyValue);
+    }
+
+    public long getFinalTTL(String tenantid, Granularity gran) {
+        long ttl;
+        if (gran == Granularity.FULL && primary.areTTLsForced()) {
+            ttl = primary.getConfigTTLForIngestion().get().toMillis();
+        } else {
+            ttl = getTTL(tenantid, gran, RollupType.BF_BASIC).get().toMillis();
+        }
+        return ttl;
+    }
+
+    private static Optional<TimeValue> getMaximum(Optional<TimeValue> primaryValue, Optional<TimeValue> safetyValue) {
+        if (!primaryValue.isPresent()) { return safetyValue; }
+        return primaryValue;
+    }
+}
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/cache/CombinedTtlProvider.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/cache/CombinedTtlProvider.java
@@ -16,6 +16,7 @@
 
 package com.rackspacecloud.blueflood.cache;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Optional;
 import com.rackspacecloud.blueflood.rollup.Granularity;
 import com.rackspacecloud.blueflood.types.RollupType;
@@ -25,13 +26,22 @@ import com.rackspacecloud.blueflood.utils.TimeValue;
  * TTL Provider which obtains the TTL from the primary or secondary provider.
  */
 public class CombinedTtlProvider implements TenantTtlProvider {
-    private final ConfigTtlProvider primary = ConfigTtlProvider.getInstance();
-    private final SafetyTtlProvider safety = SafetyTtlProvider.getInstance();
+    private final ConfigTtlProvider primary;
+    private final SafetyTtlProvider safety;
 
-    private static final CombinedTtlProvider INSTANCE = new CombinedTtlProvider();
+    private static final CombinedTtlProvider INSTANCE = new CombinedTtlProvider(
+            ConfigTtlProvider.getInstance(),
+            SafetyTtlProvider.getInstance()
+    );
 
     public static CombinedTtlProvider getInstance() {
         return INSTANCE;
+    }
+
+    @VisibleForTesting
+    CombinedTtlProvider(ConfigTtlProvider primary, SafetyTtlProvider safety) {
+        this.primary = primary;
+        this.safety = safety;
     }
 
     @Override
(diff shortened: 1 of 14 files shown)


DanielNi:
Just realized that this changes the behavior/intent of the providers in a way that might not be ideal, so I'm going to tweak it a bit.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/cache/CombinedTtlProvider.java
```diff
@@ -0,0 +1,77 @@
...
+        long ttl;
+        if (gran == Granularity.FULL && primary.areTTLsForced()) {
+            ttl = primary.getConfigTTLForIngestion().toMillis();
+        } else {
+            ttl = safety.getTTL(tenantid, gran, RollupType.BF_BASIC).get().toMillis();
+        }
+        return ttl;
+    }
+
+    private static Optional<TimeValue> getMaximum(Optional<TimeValue> primaryValue, Optional<TimeValue> safetyValue) {
```
usnavi: do we want to change the name on this?  To me, getMaximum() implies a comparison, which isn't really happening.  getTTL() ?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/cache/ConfigTtlProvider.java
```diff
@@ -41,98 +43,103 @@ public static ConfigTtlProvider getInstance() {
...
         // String rollups
-        stringTTL = new TimeValue(config.getIntegerProperty(TtlConfig.STRING_METRICS_TTL), TimeUnit.DAYS);
+        TimeValue stringTTL = null;
+        try {
+            int value = config.getIntegerProperty(TtlConfig.STRING_METRICS_TTL);
+            if (value > 0) {
+                stringTTL = new TimeValue(config.getIntegerProperty(TtlConfig.STRING_METRICS_TTL), TimeUnit.DAYS);
+            }
+        } catch (NumberFormatException ignored) {
+        }
```
usnavi: would we want to log this exception?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/cache/ConfigTtlProvider.java
```diff
@@ -41,98 +43,103 @@ public static ConfigTtlProvider getInstance() {
...
+        Configuration config,
+        Granularity granularity,
+        RollupType rollupType,
+        TtlConfig configKey) {
+        int value;
+        try {
+            value = config.getIntegerProperty(configKey);
+            if (value < 0) return false;
+        } catch (NumberFormatException e) {
+            return false;
```
usnavi: log this?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/CassandraModel.java
```diff
@@ -44,30 +44,20 @@
...
+    public static final MetricColumnFamily CF_METRICS_STRING = new MetricColumnFamily(CF_METRICS_STRING_NAME, new TimeValue(365 * 3 * 5, TimeUnit.DAYS));
+
+    public static final MetricColumnFamily CF_METRICS_PREAGGREGATED_FULL = new MetricColumnFamily(CF_METRICS_PREAGGREGATED_FULL_NAME, new TimeValue(5, TimeUnit.DAYS));
+    public static final MetricColumnFamily CF_METRICS_PREAGGREGATED_5M = new MetricColumnFamily(CF_METRICS_PREAGGREGATED_5M_NAME, new TimeValue(10, TimeUnit.DAYS));
+    public static final MetricColumnFamily CF_METRICS_PREAGGREGATED_20M = new MetricColumnFamily(CF_METRICS_PREAGGREGATED_20M_NAME, new TimeValue(20, TimeUnit.DAYS));
+    public static final MetricColumnFamily CF_METRICS_PREAGGREGATED_60M = new MetricColumnFamily(CF_METRICS_PREAGGREGATED_60M_NAME, new TimeValue(31 * 5, TimeUnit.DAYS));
+    public static final MetricColumnFamily CF_METRICS_PREAGGREGATED_240M = new MetricColumnFamily(CF_METRICS_PREAGGREGATED_240M_NAME, new TimeValue(60 * 5, TimeUnit.DAYS));
+    public static final MetricColumnFamily CF_METRICS_PREAGGREGATED_1440M = new MetricColumnFamily(CF_METRICS_PREAGGREGATED_1440M_NAME, new TimeValue(365 * 5, TimeUnit.DAYS));
```
usnavi: :+1 - i'm not verify familiar with this, but much cleaner.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/cache/CombinedTtlProvider.java
DanielNi: I think the original idea was that the safety values would be a lower bound, but I agree that this name is confusing. Besides, I don't think the lower bound is enforced anywhere (probably something worth doing).

blueflood-core/src/main/java/com/rackspacecloud/blueflood/cache/ConfigTtlProvider.java
```diff
@@ -41,98 +43,105 @@ public static ConfigTtlProvider getInstance() {
...
         // String rollups
-        stringTTL = new TimeValue(config.getIntegerProperty(TtlConfig.STRING_METRICS_TTL), TimeUnit.DAYS);
+        TimeValue stringTTL = null;
+        try {
+            int value = config.getIntegerProperty(TtlConfig.STRING_METRICS_TTL);
+            if (value > 0) {
+                stringTTL = new TimeValue(config.getIntegerProperty(TtlConfig.STRING_METRICS_TTL), TimeUnit.DAYS);
+            }
+        } catch (NumberFormatException ex) {
+            log.warn("No valid String TTL in config.");
```
usnavi: I'd include the exception as well.  E.g., http://www.slf4j.org/api/org/slf4j/Logger.html#warn(java.lang.String, java.lang.Throwable)

blueflood-core/src/main/java/com/rackspacecloud/blueflood/cache/ConfigTtlProvider.java
```diff
@@ -41,98 +43,105 @@ public static ConfigTtlProvider getInstance() {
...
+        Configuration config,
+        Granularity gran,
+        RollupType rollupType,
+        TtlConfig configKey) {
+        int value;
+        try {
+            value = config.getIntegerProperty(configKey);
+            if (value < 0) return false;
+        } catch (NumberFormatException e) {
+            log.info("No valid TTL config set for granularity: {}, rollup type: {}", gran.name(), rollupType.name());
```
usnavi: Please include the exception in the log statement.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/cache/SafetyTtlProvider.java
```diff
@@ -27,79 +29,51 @@
...
-                try {
-                    MetricColumnFamily metricCF = CassandraModel.getColumnFamily(RollupType.classOf(type, granularity), granularity);
-                    TimeValue ttl = new TimeValue(metricCF.getDefaultTTL().getValue() * 5, metricCF.getDefaultTTL().getUnit());
-                    ttlMapBuilder.put(granularity, type, ttl);
-                } catch (IllegalArgumentException ex) {
-
+                if (type == RollupType.NOT_A_ROLLUP) {
+                    continue;
                 }
+                MetricColumnFamily metricCF = CassandraModel.getColumnFamily(RollupType.classOf(type, granularity), granularity);
```
usnavi: @DanielNi  - I missed this before, but FYI, we are currently moving to the Datastax driver, which will allow us to upgrade cassandra.  MetricColumnFamily sub-classes an Astyanax-specific class.  Long-term we'll want to wrap driver-specific types in a class, e.g., https://github.com/rackerlabs/blueflood/blob/master/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DRawIO.java

However, this will probably happen after we accept this PR, so we can handle this later.

cc - @ChandraAddala @shintasmith @VinnyQ

blueflood-core/src/main/java/com/rackspacecloud/blueflood/cache/CombinedTtlProvider.java
```diff
@@ -0,0 +1,77 @@
...
+        long ttl;
+        if (gran == Granularity.FULL && primary.areTTLsForced()) {
+            ttl = primary.getConfigTTLForIngestion().toMillis();
+        } else {
+            ttl = safety.getTTL(tenantid, gran, RollupType.BF_BASIC).get().toMillis();
+        }
+        return ttl;
+    }
+
+    private static Optional<TimeValue> getTimeValue(Optional<TimeValue> primaryValue, Optional<TimeValue> safetyValue) {
```
shintasmith: Why does this need to be ```static```?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/cache/CombinedTtlProvider.java
```diff
@@ -0,0 +1,77 @@
...
+            ConfigTtlProvider.getInstance(),
+            SafetyTtlProvider.getInstance()
+    );
+
+    public static CombinedTtlProvider getInstance() {
+        return INSTANCE;
+    }
+
+    @VisibleForTesting
+    CombinedTtlProvider(ConfigTtlProvider primary, SafetyTtlProvider safety) {
```
shintasmith: 👍

blueflood-core/src/main/java/com/rackspacecloud/blueflood/cache/ConfigTtlProvider.java
```diff
@@ -41,98 +43,106 @@ public static ConfigTtlProvider getInstance() {
...
+    /**
+     * Helper function to build the ttl mapping. Only insert to the mapping if the value is a valid date.
+     * @param ttlMapBuilder
+     * @param config
+     * @param gran
+     * @param rollupType
+     * @param configKey
+     * @return true if the insertion is successful, false otherwise.
+     */
+    static boolean put(
```
shintasmith: Does this method needs to be ```static``` ?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/CassandraModel.java
```diff
@@ -44,30 +44,20 @@
...
-
-    public static final MetricColumnFamily CF_METRICS_PREAGGREGATED_FULL = new MetricColumnFamily(CF_METRICS_PREAGGREGATED_FULL_NAME, new TimeValue(1, TimeUnit.DAYS));
-    public static final MetricColumnFamily CF_METRICS_PREAGGREGATED_5M = new MetricColumnFamily(CF_METRICS_PREAGGREGATED_5M_NAME, new TimeValue(2, TimeUnit.DAYS));
-    public static final MetricColumnFamily CF_METRICS_PREAGGREGATED_20M = new MetricColumnFamily(CF_METRICS_PREAGGREGATED_20M_NAME, new TimeValue(4, TimeUnit.DAYS));
-    public static final MetricColumnFamily CF_METRICS_PREAGGREGATED_60M = new MetricColumnFamily(CF_METRICS_PREAGGREGATED_60M_NAME, new TimeValue(31, TimeUnit.DAYS));
-    public static final MetricColumnFamily CF_METRICS_PREAGGREGATED_240M = new MetricColumnFamily(CF_METRICS_PREAGGREGATED_240M_NAME, new TimeValue(60, TimeUnit.DAYS));
-    public static final MetricColumnFamily CF_METRICS_PREAGGREGATED_1440M = new MetricColumnFamily(CF_METRICS_PREAGGREGATED_1440M_NAME, new TimeValue(365, TimeUnit.DAYS));
+    public static final MetricColumnFamily CF_METRICS_FULL = new MetricColumnFamily(CF_METRICS_FULL_NAME, new TimeValue(5, TimeUnit.DAYS));
```
shintasmith: Why was this changed from 1 day to 5 days?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/CassandraModel.java
shintasmith: Actually, all of the above have changed values. Can we leave them as is? You can always configure them in config file, yes?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/Configuration.java
```diff
@@ -17,25 +17,41 @@
...
+     */
+    public static Configuration getTestInstance(Properties defaultProps) {
+        Configuration config = new Configuration(false);
+        for (Map.Entry<Object, Object> entry : defaultProps.entrySet()) {
+            config.defaultProps.put(entry.getKey(), entry.getValue());
+        }
+        return config;
+    }
+
+    private Configuration(boolean shouldInitialize) {
```
shintasmith: To be honest, I'm not exactly thrilled on the addition of parameter ```shouldInitialize``` to the constructor. It creates a different behavior of the class that skips the ```init()```, for the sake of testing. However, I don't have a better suggestion other than to mock Configuration class for testing. This has been done in some places in blueflood.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/CassandraModel.java
shintasmith: Sorry, I just realized these values were moved from TtlConfig. I guess we never used the default TTLs that were specified here?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/Configuration.java
```diff
@@ -17,25 +17,41 @@
 package com.rackspacecloud.blueflood.service;
 
 import com.google.common.collect.Lists;
+import com.rackspacecloud.blueflood.cache.ConfigTtlProvider;
```
shintasmith: Do you need this import?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/CassandraModel.java
DanielNi: Yeah, for some reason they were being multiplied by 5 there...

blueflood-core/src/main/java/com/rackspacecloud/blueflood/cache/CombinedTtlProvider.java
DanielNi: It doesn't have to be, but I suppose it's a matter of style/philosophy. I can change it if you prefer.

DanielNi:
@shintasmith @usnavi Does this look okay? If so, can we merge?

shintasmith:
@DanielNi , this looks good. Merging.

PR #705, opened 2016-06-16
Added 'per shard' HTTP query handler

DanielNi:
WHAT: This PR adds a “per shard” HTTP query handler and endpoint.

WHY: We use this at Square to populate our graphite frontend with all of our metrics. We do it “per shard” instead of querying for “all” at once is because a large request for all metrics at once gave us all sorts of problems with GC, request timeouts, etc.

HOW: Added a `HttpShardedMetricsHandler.java` class that takes in a shard number and a corresponding endpoint in `HttpMetricsIngestionHandler.java`.

Not sure if this would be useful in general, but I thought I'd open this so you guys can take a look. Let me know what you think!

### DIFF
--- a/blueflood-http/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/HttpMetricDataQueryServer.java
+++ b/blueflood-http/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/HttpMetricDataQueryServer.java
@@ -70,7 +70,7 @@ public class HttpMetricDataQueryServer {
         router.get("/v2.0/:tenantId/metric_name/search", new HttpMetricTokensHandler());
         router.get("/v2.0/:tenantId/events/getEvents", new HttpEventsQueryHandler(getEventsIO()));
         router.options("/v2.0/:tenantId/events/getEvents", new HttpEventsQueryHandler(getEventsIO()));
-        router.get("/v2.0/:tenantId/shard/:shard", new HttpShardedMetricsHandler());
+        router.get("/v2.0/:tenantId/metrics/shard/:shard", new HttpShardedMetricsHandler());
 
         log.info("Starting metric data query server (HTTP) on port {}", this.httpQueryPort);
         ServerBootstrap server = new ServerBootstrap(
--- a/blueflood-http/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/HttpMetricDataQueryServer.java
+++ b/blueflood-http/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/HttpMetricDataQueryServer.java
@@ -70,6 +70,7 @@ public class HttpMetricDataQueryServer {
         router.get("/v2.0/:tenantId/metric_name/search", new HttpMetricTokensHandler());
         router.get("/v2.0/:tenantId/events/getEvents", new HttpEventsQueryHandler(getEventsIO()));
         router.options("/v2.0/:tenantId/events/getEvents", new HttpEventsQueryHandler(getEventsIO()));
+        router.get("/v2.0/:tenantId/shard/:shard", new HttpShardedMetricsHandler());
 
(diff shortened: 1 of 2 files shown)


blueflood-http/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/HttpMetricDataQueryServer.java
```diff
@@ -70,6 +70,7 @@ public void startServer() {
         router.get("/v2.0/:tenantId/metric_name/search", new HttpMetricTokensHandler());
         router.get("/v2.0/:tenantId/events/getEvents", new HttpEventsQueryHandler(getEventsIO()));
         router.options("/v2.0/:tenantId/events/getEvents", new HttpEventsQueryHandler(getEventsIO()));
+        router.get("/v2.0/:tenantId/shard/:shard", new HttpShardedMetricsHandler());
```
ChandraAddala: The url doesnt give a sense that the request is to get "metrics". May be something like this?

`/v2.0/:tenantId/metrics/shard/:shard`

ChandraAddala:
Other than the comment I had, everything else looks good.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/HttpShardedMetricsHandler.java
```diff
@@ -0,0 +1,106 @@
...
+    private static final Logger log = LoggerFactory.getLogger(HttpShardedMetricsHandler.class);
+    private final LocatorIO io = IOContainer.fromConfig().getLocatorIO();
+    private final Timer httpShardedMetricsFetchTimer = Metrics.timer(HttpShardedMetricsHandler.class,
+            "Handle HTTP request for metrics by shard");
+
+    @Override
+    public void handle(ChannelHandlerContext ctx, HttpRequest request) {
+        Tracker.getInstance().track(request);
+        final String tenantId = request.getHeader("tenantId");
+        final String shardId = request.getHeader("shard");
```
usnavi: @shintasmith @VinnyQ @ChandraAddala - if we accept this, would reading generic-sounding header names be an issue?  Should they be more specific?  Also, we'll need to see how the "tenantId" would match up with how we run repose and tenanted access.

shintasmith:
I am not sure I like exposing the notion of "shard" to the Blueflood API. To me, shard is an internal grouping of work for blueflood nodes. But I will discuss this with my team and get back to you.

DanielNi:
That makes sense; our intent with this endpoint was to quickly query for all metrics in Blueflood without the associated problems with such a large request. If there is a better way to do this without exposing the concept of "shard" I would love to hear it!

ChandraAddala:
There is also `/metrics/search` API which is powered by elastic search, using which you can query for all metrics belonging to a tenant. It supports glob search.

shintasmith:
@DanielNi , that's funny.... because yesterday while discussing this with my team, I came across a similar thought... You could write a Java standalone program that queries Cassandra directly, maybe? I don't know how exactly you would be consuming the output of this.

I take it you guys don't use ElasticSearch ? We have a python project:
https://github.com/rackerlabs/blueflood-graphite-finder

that maybe is doing the same thing as what you are doing. We use this to serve Grafana a drop down list of metric names. Internally, it uses ```/v2.0/{tenantId}/metric_name``` API. I am failing to find the docs for this tho... I'll get back to you on this.

DanielNi:
We don't use ElasticSearch, but we have a Java layer in between Blueflood and Graphite (that queries this endpoint) which seems to do something similar to `blueflood-graphite-finder`. I'll have to look into the API some more, but I think modifying this layer to talk to Cassandra directly might be the way to go.

Thanks for the suggestions! Feel free to close this PR if it's no longer relevant.

shintasmith:
Closing PR as we have discussed this feature is best implemented outside Blueflood.

PR #706, opened 2016-06-17
Added back Zookeeper classes, tests and configs

DanielNi:
**WHAT**: This is the ZooKeeper code we’re running internally to manage shard ownership. It’s a slightly tweaked version of what was removed in #613.

**WHY**: We've been running this in production for a while and it seems to work for us.

**HOW**: Adding our `ZKShardLockManager.java` and tests as well as adding back ZooKeeper related configs

I’m still not completely sure why ZK was removed in the first place, but we certainly find it useful so it might be worth adding back.


usnavi:
@DanielNi this is great!  We removed ZK because we weren't using it, supposedly because it wasn't working.  You all running it in prod, gives us a lot higher confidence.  thanks!

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/CoreConfig.java
```diff
@@ -90,12 +90,16 @@
     SHARD_PUSH_PERIOD("2000"),
     SHARD_PULL_PERIOD("20000"),
 
+    // blueflood uses zookeeper to acquire locks before working on shards
+    ZOOKEEPER_CLUSTER("127.0.0.1:22181"),
```
shintasmith: Would it be too much trouble to default this to NONE here? The reason is... existing blueflood consumers (who are not using ZK) would need to know to deploy a config change that explicitly sets this to NONE, when they deploy this code change.

shintasmith:
Other than my comment on the default value of the Config, everything looks good. Thanks, @DanielNi !

DanielNi:
@shintasmith @usnavi Same here. Is this one okay to be merged?

usnavi:
lgtm.  @DanielNi - this branch has merge conflicts.  Would you mind rebasing this with master?

DanielNi:
@usnavi Rebased.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/ScheduleContext.java
```diff
@@ -152,9 +153,20 @@ public ScheduleContext(long currentTimeMillis, Collection<Integer> managedShards
         registerMBean();
     }
 
+    public ScheduleContext(long currentTimeMillis, Collection<Integer> managedShards, String zookeeperCluster) {
+        this.scheduleTime = currentTimeMillis;
+        this.shardStateManager = new ShardStateManager(managedShards, asMillisecondsSinceEpochTicker());
+        ZKShardLockManager lockManager = new ZKShardLockManager(zookeeperCluster, new HashSet<Integer>(shardStateManager.getManagedShards()));
+        lockManager.init(new TimeValue(5, TimeUnit.SECONDS));
```
ChandraAddala: Can you create a constant for this 5 seconds? Also explaining what this value is.

PR #707, opened 2016-06-17
"sum" sub-metric for BasicRollup

usnavi:
Adding "sum" sub-metric for BasicRollup.
Created BaseRollup class which is parent to BluefloodGaugeRollup & BasicRollup and pulled up common sub-metrics.
Removed com.rackspacecloud.blueflood.io.serializers.json.* and associated unit tests as they are not used anywehre in the codebase.


blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/serializers/metrics/BaseRollupSerDes.java
```diff
@@ -0,0 +1,105 @@
...
+import com.google.protobuf.CodedOutputStream;
+import com.rackspacecloud.blueflood.exceptions.SerializationException;
+import com.rackspacecloud.blueflood.io.Constants;
+import com.rackspacecloud.blueflood.types.BaseRollup;
+import com.rackspacecloud.blueflood.utils.Metrics;
+
+import java.io.IOException;
+
+/**
+ * The serialization and deserialization methods for sub-metrics used by BasicMetricRollup and GaugeMetricRollup.
```
shintasmith: Add comment on what is common between BasicRollupSerDes and GaugeSerDes, i.e: the ser-des of substats average, variance, min, max.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/serializers/metrics/BaseRollupSerDes.java
```diff
@@ -0,0 +1,105 @@
...
+ */
+public class BaseRollupSerDes extends AbstractSerDes {
```
shintasmith: Minor comment: I wonder if we should pick a more meaningful name than ```BaseRollupSerDes```? Maybe ```SubStatRollupSerDes``` ? Or ```AvgVarMinMaxRollupSerDes``` ?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/serializers/metrics/BasicRollupSerDes.java
```diff
@@ -54,74 +48,48 @@ public BasicRollup deserialize(ByteBuffer byteBuffer) {
...
-        serializeRollup(basicRollup, protobufOut);
-    }
+        protobufOut.writeRawByte(Constants.VERSION_2_ROLLUP);
 
-    protected void serializeRollup(BasicRollup basicRollup, CodedOutputStream protobufOut) throws IOException {
-        protobufOut.writeRawByte(Constants.VERSION_1_ROLLUP);
-        protobufOut.writeRawVarint64(basicRollup.getCount());          // stat count
+        serializeBaseRollupHelper( basicRollup, protobufOut );
 
         if (basicRollup.getCount() > 0) {
```
shintasmith: Should this be a different check? What if basicRollup.getCount() is only 4 (only has avg, variance, min, max) ?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/types/BaseRollup.java
```diff
@@ -0,0 +1,177 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.util.Map;
+
+/**
+ * The sub-metrics used by BasicMetricRollup and GaugeMetricRollup.
+ */
+public abstract class BaseRollup implements Rollup {
```
shintasmith: Why don't we name it SubMetricsRollup ?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/serializers/metrics/BaseRollupSerDes.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    /**
+     * Serialize without sum attribute.  This is used by {@link GaugeSerDes}
+     *
+     * @param baseRollup
+     * @param protobufOut
+     * @throws IOException
+     */
+    protected void serializeRollupV1( BaseRollup baseRollup, CodedOutputStream protobufOut ) throws IOException {
```
shintasmith: Don't you need to call ```rollupSize.update(buf.length)``` somewhere here?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/serializers/metrics/BaseRollupSerDes.java
usnavi: I'm up for a different name, but the two suggestions don't really seem that better, IMHO.  AvgVarMinMaxRollupSerDes is too specific as the list of stats might change.  SubStatRollupSerDes doesn't seem any different than BaseRollupSerDes.  Let me think about this.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/serializers/metrics/BasicRollupSerDes.java
usnavi: getCount() returns number of points.  If there are no points, then there are no stats.  E.g., can't have a maxValue with no data.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/outputs/serializers/BasicRollupsOutputSerializer.java
```diff
@@ -101,8 +101,8 @@ Object convertRawSampleToObject(Object rawSample) {
         NUM_POINTS("numPoints") {
             @Override
             Object convertRollupToObject(Rollup rollup) throws Exception {
-                if (rollup instanceof BasicRollup)
-                    return ((BasicRollup) rollup).getCount();
+                if (rollup instanceof BaseRollup)
+                    return ((BaseRollup) rollup).getCount();
```
shintasmith: Are you sure Gauge supports returning the number of points?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/serializers/metrics/BaseRollupSerDes.java
```diff
@@ -0,0 +1,112 @@
...
+import com.rackspacecloud.blueflood.exceptions.SerializationException;
+import com.rackspacecloud.blueflood.io.Constants;
+import com.rackspacecloud.blueflood.types.BaseRollup;
+import com.rackspacecloud.blueflood.utils.Metrics;
+
+import java.io.IOException;
+
+/**
+ * The serialization and deserialization methods for sub-metrics used by BasicMetricRollup and GaugeMetricRollup.
+ * Comment sub-metrics are:
```
shintasmith: Common sub-metrics, you mean?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/serializers/metrics/BasicRollupSerDes.java
shintasmith: Gotcha

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/serializers/metrics/BaseRollupSerDes.java
```diff
@@ -0,0 +1,112 @@
...
+ *     <li>min</li>
+ *     <li>sum</li>
+ * </ul>
+ */
+public class BaseRollupSerDes extends AbstractSerDes {
+
+    /**
+     * Our internal metric to count the number of rollups
+     */
+    protected static Histogram rollupSize = Metrics.histogram( BasicRollupSerDes.class, "Rollup Metric Size" );
```
shintasmith: Comment seems wrong. It's not a count of number of rollups. It seems to be tracking the buffer size. But that could be existing comment. 

I wonder if this metric is useful at all. Doesn't the buffer size the same for all? I mean maybe there will be a difference between Gauge rollup buffer size and Basic metric rollup buffer size, now that we add "sum" for Basic, but not Gauge. 

You don't need to change anything in the code... I'm just commenting :-)

blueflood-core/src/test/java/com/rackspacecloud/blueflood/io/serializers/json/IMetricSerializerTest.java
```diff
@@ -1,138 +0,0 @@
...
-import com.rackspacecloud.blueflood.utils.TimeValue;
-import junit.framework.Assert;
-import org.codehaus.jackson.map.ObjectMapper;
-import org.junit.Test;
-
-import java.io.IOException;
-import java.math.BigInteger;
-import java.util.concurrent.TimeUnit;
-
-public class IMetricSerializerTest {
```
shintasmith: Why was this removed?

usnavi: This class tested the com.rackspacecloud.blueflood.io.serializers.json.* classes, which I removed as they were no longer used in the codebase.

blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/outputs/handlers/HttpRollupsQueryHandlerIntegrationTest.java
```diff
@@ -84,7 +84,7 @@ public void testSingleplotQueryWithEnum() throws Exception {
         final String metric_name = "enum_metric_test" + postfix;
 
         // post multi metrics for ingestion and verify
-        HttpResponse response = postMetric(tenant_id, postAggregatedPath, "sample_enums_payload.json", now, postfix );
+        HttpResponse response = postMetric(tenant_id, postAggregatedPath, "sample_single_enum_payload.json", now, postfix );
```
shintasmith: I know this test has been breaking intermittently. But curious, why did you change it to just post one Enum? What's wrong with posting multiple Enums?

shintasmith:
Sorry if I miss this, but should we be adding tests that cover these scenarios:
* get the sum of Basic metrics for a time frame when they were rolled up with no sum (i.e: v1)
* get the sum of Basic metrics for a time frame when some points were rolled up without sum, some points were rolled up with sum
* get the sum of Basic metrics for a time frame when all points were rolled up with sum
* do the same for Gauge
* one of the code change essentially enables "numPoints" for Gauge (it wasn't enabled before). So we probably should add a test for this. And document.

blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/outputs/handlers/HttpRollupsQueryHandlerIntegrationTest.java
usnavi: This test was posting several enum values to 1 timestamp (values of v1, v2, v3).  Only 1 of them is recorded on our local boxes as well as on jdk7 on travis, the v3 value.  However, on travis only v2 is returned under jdk 8.  I changed this so we can verify the default behavior on travis jdk8.  I didn't mess with this issue as we will eventually be getting rid of enums.  I can document this in the code.

Now, the multiple eums values are recorded in elasticsearch as the enum values are turned into metric names, so they can be queried. This is tested in HttpEnumIntegrationTest.testMetricIngestionWithEnum().  So yeah, enums are whack.  Only one value is saved for a timestamp, but all submitted values, regardless if they are the same timestamp are recorded in elasticsearch.

shintasmith: That's whack!!

blueflood-http/src/main/java/com/rackspacecloud/blueflood/outputs/serializers/BasicRollupsOutputSerializer.java
usnavi: BluefloodGaugeRollup always sub-classed BasicRollup, so it would have been caught in the pre-existing  if ( rollup instanceof BasicRollup) check, right?  If so, then numPoints already existed for Gauge.

blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/outputs/handlers/HttpRollupsQueryHandlerIntegrationTest.java
shintasmith: But maybe change the test so that it posts the different Enum values with different timestamps?

blueflood-http/src/main/java/com/rackspacecloud/blueflood/outputs/serializers/BasicRollupsOutputSerializer.java
shintasmith: Never mind... I think it does today.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/types/BaseRollup.java
```diff
@@ -0,0 +1,177 @@
...
+package com.rackspacecloud.blueflood.types;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.util.Map;
+
+/**
+ * The sub-metrics used by BasicMetricRollup and GaugeMetricRollup.
```
ChandraAddala: Are BasicMetricRollup and GaugeMetricRollup class names? The look they are class names.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/types/BaseRollup.java
```diff
@@ -0,0 +1,177 @@
...
+ */
+public abstract class BaseRollup implements Rollup {
+    private static final Logger log = LoggerFactory.getLogger( BaseRollup.class );
+    public static final int NUM_STATS = 4;
```
ChandraAddala: Seeing the number '4' hardcoded bothers me a little bit. Is there a way to derive this number?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/serializers/metrics/BaseRollupSerDes.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    /**
+     * Serialize without sum attribute.  This is used by {@link GaugeSerDes}
+     *
+     * @param baseRollup
+     * @param protobufOut
+     * @throws IOException
+     */
+    protected void serializeRollupV1( BaseRollup baseRollup, CodedOutputStream protobufOut ) throws IOException {
```
ChandraAddala: `deserialize` takes version as parameter. Should we make `serialize` method also generic and take version as parameter?

ChandraAddala:
Other than the minor comments I had, everything else looks good.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/serializers/metrics/BaseRollupSerDes.java
usnavi: I was thinking that the BaseRollupSerDes is only aware of VERSION_1_ROLLUP serialization, the VERSION_2_ROLLUP is BasicRollup specific (but leverages some functionality of BaseRollupSerDes).  Didn't think it was worth passing in the version because of this.

The only reason the version is passed in for deserializeBaseRollup() is for error reporting.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/types/BaseRollup.java
usnavi: Yeah, I agree with you.  Not sure how to calculate that number in a meaningful way though.  Let's talk offline.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/io/serializers/SerializationTest.java
```diff
@@ -135,7 +139,35 @@ public void testFullResSerializationAndDeserializationV1() throws IOException {
...
+            Assert.assertEquals( "count is same after serialization for V1", TO_SERIALIZE_BASIC_ROLLUP[i].getCount(), basicRollup.getCount() );
+            Assert.assertEquals( "max is same after serialization for V1", TO_SERIALIZE_BASIC_ROLLUP[i].getMaxValue(), basicRollup.getMaxValue() );
+            Assert.assertEquals( "min is same after serialization for V1", TO_SERIALIZE_BASIC_ROLLUP[i].getMinValue(), basicRollup.getMinValue() );
+            Assert.assertEquals( "var is same after serialization for V1", TO_SERIALIZE_BASIC_ROLLUP[i].getVariance(), basicRollup.getVariance() );
+            Assert.assertEquals( "sum is 0 after serialization for V1", 0, basicRollup.getSum(), EPSILON );
+        }
+    }
+
+    @Test
+    public void testBasicRollupSerializationAndDeserialization() throws IOException {
```
usnavi: @shintasmith - I added a test to verify that sum = 0 for all BasicRollups which are serialized as V1.  This should answer the comment for verifying that sum == 0 when V1 is deserialized.

shintasmith:
I don't have anything else to add. lgtm

PR #713, opened 2016-06-28
refactor to ensure datastax PreparedStatements are created only once

shintasmith:
This PR fixes some of the warning logs coming out from Datastax, indicating that most of our PreparedStatements are created more than once. The PreparedStatements are mostly created inside of the *IO classes constructors. In particular, the PreparedStatements were created more than once for the following cases:

* Both IOContainer and DAbstractMetricsRW classes instantiate new instance of DLocatorIO. This is fixed by making DAbstractMetricsRW constructor takes a LocatorIO.
* IOContainer holds instance of *BasicMetricsRW and *PreagggregatedMetricsRW. The first time IOContainer.fromConfig() is called, it will instantiate all the *IO classes and the *RW classes. The two *RW classes hold an instance of MetadataCache. But MetadataCache needs an instance of MetadataIO class, so it calls IOContainer.fromConfig().getMetadataIO(). The call to fromConfig() causes IOContainer to create another set of *IO classes and *RW classes. This is fixed by delaying the call to MetadataCache.getInstance() inside the methods that need it.
* A lot of PreparedStatements live inside DAbstractMetricsIO, which many other *IO classes extend from. Everytime we create DEnumIO, DLocatorIO, etc, new instances of these PreparedStatements are created. This is fixed by moving all the PreparedStatements to a singleton class DMetricsCFPreparedStatement

### DIFF
--- a/blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionServer.java
+++ b/blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionServer.java
@@ -28,7 +28,6 @@ import com.rackspacecloud.blueflood.inputs.processors.DiscoveryWriter;
 import com.rackspacecloud.blueflood.inputs.processors.BatchWriter;
 import com.rackspacecloud.blueflood.inputs.processors.RollupTypeCacher;
 import com.rackspacecloud.blueflood.inputs.processors.TypeAndUnitProcessor;
-import com.rackspacecloud.blueflood.io.AbstractMetricsRW;
 import com.rackspacecloud.blueflood.io.EventsIO;
 import com.rackspacecloud.blueflood.service.*;
 import com.rackspacecloud.blueflood.tracker.Tracker;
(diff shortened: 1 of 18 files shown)


usnavi:
lgtm

ChandraAddala:
Looks good.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionServer.java
```diff
@@ -129,9 +128,22 @@ public ChannelPipeline getPipeline() throws Exception {
...
+                        } else {
+                            log.debug("channel " + ctx.getChannel().toString() + " is no longer writeable, not sending 400 response back to client");
+                        }
+                    } catch (Exception ex) {
+                        // If we are getting exception trying to write,
+                        // don't propagate to caller. It may cause this
+                        // method to be called again and will produce
+                        // stack overflow. So just log it here.
+                        log.debug("Can't write to channel " + ctx.getChannel().toString(), ex);
+                    }
```
shintasmith: @usnavi @ChandraAddala I added a fix to the stack overflow issue I saw when I was using grinder against the Ingest nodes directly (no LB VIP involved). Pls review.

ChandraAddala: :thumbsup:

izrik:
lgtm

PR #715, opened 2016-06-29
Isolating-Docker-Docker-Compose

goru97:
## What:
Isolate Docker and Docker-Compose

## Why:
Users may want to run Blueflood's docker container with previously running Instances of C* and ES or if they want to do some customization with the BF image (although it's already very flexible). Theuy can do so by using the docker setup.
If a user wants to run the datastores (C* and ES) in containers, he/she can use docker-compose setup (docker-compose.yml).
If a user also wants to run graphite-api container to see the metrics using Grafana, it can be done by using another docker-compose file i.e. docker-compose-graphite-api.yml


usnavi:
@goru97 - would you please add a description to this PR?  Its helpful for archival purposes as well.

contrib/blueflood-docker-compose/docker-compose.yml
```diff
@@ -1,5 +1,5 @@
 blueflood:
-  build: .
+  image: rackerlabs/blueflood
```
usnavi: 👍  oh yeah!

goru97:
@usnavi Added.

usnavi:
lgtm

Going forward, do we want to put the whole docker setup in its own repo?  I'd imagine most people who are running the docker instance don't care about the source code.  Something to think about as we work on the grafana 3 stuff.

goru97:
@usnavi We have to keep it somewhere as we'll be needing this to create images in future. I didn't understand when you said: "Something to think about as we work on the grafana 3 stuff."

usnavi:
@goru97 - I merely was thinking out loud.  My thought:  would our docker-related stuff be more presentable in its own project, rather than with the source-code.

goru97:
@usnavi Can you land this PR for now. You can host the dockerfile and docker-enterypoint.sh separately afterwards if you'd like.

PR #716, opened 2016-06-29
Fixing blueflood counter rollup tests.

ChandraAddala:
For Counters we provide sum and the rate. 
1) During rollup, we just add all the counters (we dont average them, since our counters do not have avg)
2) We compute rate while rolling up. For computing rate from raw samples, we divide the total count by the total time span. For computing rate from rollups of lower granularity, we compute the the rate by dividing the total count by sum of all the time spans.

As per what I see, blueflood is doing it the right way.


ChandraAddala:
@izrik Can you take a look at the changes I had made and let me know what u think?

shintasmith:
I also think the current implementation is right. lgtm 👍

izrik:
@ChandraAddala lgtm.

PR #717, opened 2016-07-06
dont print stack trace on bad inputs

shintasmith:
In Prod, we have a few clients sending lots of Query requests with ```select=average``` on lots of Gauge metrics. The warning log message and its corresponding stack trace ended up filling up disk space. 

This PR changes Blueflood so that it doesn't log the stack trace on what is essentially bad input from clients.

### DIFF
--- a/blueflood-http/src/main/java/com/rackspacecloud/blueflood/outputs/serializers/JSONBasicRollupsOutputSerializer.java
+++ b/blueflood-http/src/main/java/com/rackspacecloud/blueflood/outputs/serializers/JSONBasicRollupsOutputSerializer.java
@@ -169,8 +169,8 @@ public class JSONBasicRollupsOutputSerializer implements BasicRollupsOutputSeria
                 } else {
                     filteredObject.put(stat.toString(), filteredValue);
                 }
-            } catch (Exception ex) {
-                log.warn(ex.getMessage(), ex);
+            } catch (UnsupportedOperationException ex) {
+                log.warn(ex.getMessage());
             }
         }
         return filteredObject;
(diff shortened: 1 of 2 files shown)


shintasmith:
@usnavi @ChandraAddala @VinnyQ @izrik @drenalin23 , please review.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/outputs/serializers/JSONBasicRollupsOutputSerializer.java
```diff
@@ -169,8 +169,8 @@ private JSONObject getFilteredStatsForRollup(Rollup rollup, Set<MetricStat> filt
                 } else {
                     filteredObject.put(stat.toString(), filteredValue);
                 }
-            } catch (Exception ex) {
-                log.warn(ex.getMessage(), ex);
+            } catch (UnsupportedOperationException ex) {
+                log.warn(ex.getMessage());
```
usnavi: Should we log this as DEBUG?  Do we care if they ask for unsupported metrics?  that's user error, right?

Also, we can capture the 4XX in our repose access logs.

shintasmith: @usnavi , I don't think we are returning 400 in this case. They are fetching multiple metrics. They just passed in ```select=average``` for all of them. I think we just send back what submetrics are supported for each type of metrics, and ignore the rest.

I am fine with debug as well. Let's see what @ChandraAddala thinks.

ChandraAddala: I will go with debug.

usnavi:
lgtm

shintasmith:
@usnavi @ChandraAddala , I updated to log debug. Also change the thing to ```ex.toString()``` because that will print also the exception class. Not that it matters too much in this case, but I like it better :)

PR #719, opened 2016-07-19
Perf improvement for aggregation query from /metric_name/search API

ChandraAddala:
I made a change in the aggregated query that we use. Added another filter to filter data before data is used for aggregation.


shintasmith:
not knowing anything about ES, lgtm :-)

PR #720, opened 2016-07-21
New perf script

izrik:
This PR cleans up the python code. It almost completely whitespace changes. It makes no changes to the functionality. It reduces the neumber of `pep8` violations from 500 down to 47. Some violations were not removed:
* Line-too-long violations due to long string literals
* "module level import not at top of file" violations in `tests.py`, due to having to modify the package path before importing
* One instance of `x == None` instead of `x is None`, which if fixed could change behavior in some rare circumstances

### DIFF
--- a/contrib/grinder/scripts/annotationsingest.py
+++ b/contrib/grinder/scripts/annotationsingest.py
@@ -1,78 +1,80 @@
 import random
-try: 
-  from com.xhaus.jyson import JysonCodec as json
+
+try:
+    from com.xhaus.jyson import JysonCodec as json
 except ImportError:
-  import json
+    import json
 from utils import *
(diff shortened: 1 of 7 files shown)


izrik:
@VinnyQ @ChandraAddala @usnavi @shintasmith Please review.

shintasmith:
lgtm 👍

usnavi:
lgtm.  I assume this was tabs -> whitespace?

izrik:
@usnavi Not tab characters, specifically, but still related to indenting. Most of the code was indented at intervals of two spaces, whereas `pep8` prefers four. Also, a lot of lines were running over the 80-character limit, and those got split up properly. I don't think there were any tabs.

PR #721, opened 2016-07-25
Add CORS support for query and options requests

VinnyQ:
## WHAT
Add Cross-Origin Resource Sharing (CORS aka Cross Domains) support in blueflood.

## WHY
To allow graphite-api, finders, and grafana on a separate domain to obtain metrics from blueflood

## HOW
- refactor the test bases for https and outputs handlers.
- rename HttpAnnotationsIntegrationTest to HttpEventsQueryHandlerIntegrationTest
- add cors support for options request
- modify query responses to return accept headers origin
- make cors origin, methods, headers, and max age configurable

## TEST
After deployment, run the following curl commands against the deployed query nodes:
```
curl -i -X OPTIONS 'http://localhost:2500/v2.0/333333/metrics/search'
curl -i -X OPTIONS 'http://localhost:2500/v2.0/333333/metric_name/search'
curl -i -X OPTIONS 'http://localhost:2500/v2.0/333333/views'
curl -i -X OPTIONS 'http://localhost:2500/v2.0/333333/views/metric.name'
```

And verify that they all return the following response code and headers
```
HTTP/1.1 204 No Content
Access-Control-Allow-Headers: X-Auth-Token, Accept
Access-Control-Allow-Origin: *
Access-Control-Max-Age: 1728000
Access-Control-Allow-Methods: GET
Content-Length: 0
```


VinnyQ:
@usnavi @shintasmith @ChandraAddala @izrik please review, thanks!

blueflood-http/src/main/java/com/rackspacecloud/blueflood/http/DefaultHandler.java
```diff
@@ -46,18 +46,18 @@ public static void sendResponse(ChannelHandlerContext channel, HttpRequest reque
...
 
+        final Timer.Context sendResponseTimerContext = sendResponseTimer.time();
         try {
             if (messageBody != null && !messageBody.isEmpty()) {
                 response.setContent(ChannelBuffers.copiedBuffer(messageBody, Constants.DEFAULT_CHARSET));
             }
 
-            Tracker.getInstance().trackResponse(request, response);
             HttpResponder.respond(channel, request, response);
+            Tracker.getInstance().trackResponse(request, response);
```
shintasmith: Just curious, why was the order changed?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/CoreConfig.java
```diff
@@ -174,8 +174,13 @@
     // 3 days - this matches the TTL for our metrics_full table, we don't accept anything older than the TTL.
     BEFORE_CURRENT_COLLECTIONTIME_MS("259200000"),
     // 10 minutes
-    AFTER_CURRENT_COLLECTIONTIME_MS("600000");
+    AFTER_CURRENT_COLLECTIONTIME_MS("600000"),
 
+    // Cross-Origin Resource Sharing
+    CORS_ALLOWED_ORIGINS(""),
```
shintasmith: Is empty String a valid value for this header? This w3C page doesn't say it's valid:
https://www.w3.org/TR/2008/WD-access-control-20080912/#access-control-allow-origin

blueflood-http/src/main/java/com/rackspacecloud/blueflood/http/HttpResponder.java
```diff
@@ -34,6 +36,10 @@ public static void respond(ChannelHandlerContext ctx, HttpRequest req, HttpRespo
     }
 
     public static void respond(ChannelHandlerContext ctx, HttpRequest req, HttpResponse res) {
+
+        // set response headers
+        res.setHeader("Access-Control-Allow-Origin", Configuration.getInstance().getStringProperty(CoreConfig.CORS_ALLOWED_ORIGINS));
```
shintasmith: If it's invalid, per spec, to send empty Access-Control-Allow-Origin, should we put this in an if statement? I.e: only sends back the header if the config has non empty configuration?

shintasmith: Also, do we need to get this property on every request? Can we get the value of it in the constructor and reuse?

blueflood-http/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/HttpEventsQueryHandler.java
```diff
@@ -29,47 +30,37 @@ public HttpEventsQueryHandler(EventsIO searchIO) {
     @Override
     public void handle(ChannelHandlerContext ctx, HttpRequest request) {
 
-        if(request.getMethod().toString().equalsIgnoreCase("GET")) {
```
shintasmith: Why was this check for GET removed? Don't we want to only serve GET requests for the URI /v2.0/tenantId/events/getEvents ?

izrik:
I'm looking over the changes to the test classes. I'll let you know if I find anythimg.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/http/DefaultHandler.java
VinnyQ: I would like to be able to track the response headers too, so have to move it after the respond call where the headers are set.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/CoreConfig.java
VinnyQ: @shintasmith 
Good points... I'll modify the code to not create any CORS headers in the response if the origins are set to empty string.  It'll be the main switch for turning CORS support or and off.  

But since we're here, should we default this value to empty string in blueflood?  or should it be "*"?  With wiring the getEvents code to this configuration with this PR, it really should be "*" (which I was going to create a blueflood-chef PR to add to the blueflood.conf on the query nodes)... but I was afraid that "*" as the default might be a security risk for us and any installer of bluefloods if they're not aware of this setting.   What do you think the default should be?

blueflood-http/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/HttpEventsQueryHandler.java
VinnyQ: @shintasmith 
I moved *ALL* OPTIONS requests to be handled by the new class HttpOptionsHandler in .../blueflood-http/.../outputs/handlers/HttpOptionsHandler.java.

See the code in class HttpMetricDataQueryServer for the handler changes to 
```
router.options("/v2.0/:tenantId/events/getEvents", new HttpOptionsHandler());
```

i.e.
Previously, this class was routing both GET and OPTIONS request for this path to HttpEventsQueryHandler, hence the need to check the request methods.  But now only GET requests are routed to this the HttpEventsQueryHandler so this class is now dedicated to only have to serve GET requests (similar to other handlers in the outputs.handlers package with the exception of the new HttpOptionsHandler class).

blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/http/HttpIntegrationTestBase.java
```diff
@@ -259,7 +330,7 @@ public URIBuilder getMetricsURIBuilder() throws URISyntaxException {
 
     public URIBuilder getMetricDataQueryURIBuilder() throws URISyntaxException {
         return new URIBuilder().setScheme("http").setHost("127.0.0.1")
-                .setPort(httpPortQuery).setPath("/v2.0/tenantId/metrics/search")
+                .setPort(httpPortQuery).setPath(getSearchPath)
```
izrik: @VinnyQ  Won't this leave a `%s` in the path? Should that be replaced with an actual tenant id?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/CoreConfig.java
ChandraAddala: Yeah, by default, there should not be any CORS support in blueflood.

shintasmith: @VinnyQ so I am ok with keeping the config here. The default of the configuration property can be "". But in the code that actually sends back the headers, we should check this config. If the config is "", then don't send back any CORS headers. Sending ```Access-Control-Allow-Origin: ``` with empty value is illegal.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/http/HttpResponder.java
shintasmith: @VinnyQ , here, instead of always calling ```res.setHeader()```, we should check the value of the Config. Only call ```res.setHeader()``` if the value of the CORS_ALLOWED_ORIGINS is not empty

blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/http/HttpIntegrationTestBase.java
VinnyQ: @izrik this is just a default path to show the format of the path (note the hard coed "tenantId" string in the previous version.  If you look at where getMetricDataQueryURIBuilder() are actually use, they would have a setPath call there to replace it with the correct path (i.e. see queryMetricIncludeEnum()).

This is similar to how the other other getXXXURIBuilder methods in the test class are written -- all i am doing is making the string a variable so it can be reused in getQuerySearchURI.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/HttpMetricDataQueryServer.java
```diff
@@ -61,15 +61,21 @@ public void startServer() {
...
         router.get("/v2.0/:tenantId/metrics/search", new HttpMetricsIndexHandler());
         router.get("/v2.0/:tenantId/metric_name/search", new HttpMetricTokensHandler());
         router.get("/v2.0/:tenantId/events/getEvents", new HttpEventsQueryHandler(getEventsIO()));
-        router.options("/v2.0/:tenantId/events/getEvents", new HttpEventsQueryHandler(getEventsIO()));
+
+        router.options("/v2.0/:tenantId/views/:metricName", new HttpOptionsHandler());
+        router.options("/v2.0/:tenantId/views", new HttpOptionsHandler());
+        router.options("/v2.0/:tenantId/metrics/search", new HttpOptionsHandler());
+        router.options("/v2.0/:tenantId/metric_name/search", new HttpOptionsHandler());
+        router.options("/v2.0/:tenantId/events/getEvents", new HttpOptionsHandler());
```
VinnyQ: @shintasmith @ChandraAddala @usnavi Hi guys, note here that I hard coded all these OPTIONS request paths to be individual lines all with the ame HttpOptionsHandler.  However, in theory we can just make this all one line with a regex that supports any OPTIONS requests with the path pattern matching "/v2.0/:tenantId/.*" which should covers any path starting with that pattern, and return the CORS response headers.  

Do you guys prefer it this way (with the very specific paths that allows for OPTIONS request method)?  Or would you prefer opening it to any paths with some sort of common regex pattern?

blueflood-http/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/HttpMetricTokensHandler.java
```diff
@@ -40,10 +40,10 @@ public HttpMetricTokensHandler(DiscoveryIO discoveryHandle) {
     @Override
     public void handle(ChannelHandlerContext ctx, HttpRequest request) {
 
-        final Timer.Context httpMetricNameTokensHandlerTimerContext = HttpMetricNameTokensHandlerTimer.time();
-
         Tracker.getInstance().track(request);
```
VinnyQ: note:  this was moved because I don't think the timer metrics for this handler should be affect by the tracking call if it's turned on.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/HttpOptionsHandler.java
```diff
@@ -0,0 +1,41 @@
...
+    private final String ALLOWED_ORIGINS = Configuration.getInstance().getStringProperty(CoreConfig.CORS_ALLOWED_ORIGINS);
+    private final String ALLOWED_METHODS = Configuration.getInstance().getStringProperty(CoreConfig.CORS_ALLOWED_METHODS);
+    private final String ALLOWED_HEADERS = Configuration.getInstance().getStringProperty(CoreConfig.CORS_ALLOWED_HEADERS);
+    private final String ALLOWED_MAX_AGE = Configuration.getInstance().getStringProperty(CoreConfig.CORS_ALLOWED_MAX_AGE);
+
+    @Override
+    public void handle(ChannelHandlerContext ctx, HttpRequest request) {
+        // log the request if tracking enabled
+        Tracker.getInstance().track(request);
+
```
VinnyQ: @shintasmith @ChandraAddala how does the change below looks?

blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/http/HttpIntegrationTestBase.java
```diff
@@ -65,26 +81,75 @@
...
+    }
+
+    @AfterClass
+    public static void shutdown() {
+        Configuration.getInstance().setProperty(CoreConfig.DISCOVERY_MODULES.name(), "");
+        Configuration.getInstance().setProperty(CoreConfig.ENUMS_DISCOVERY_MODULES.name(), "");
+        Configuration.getInstance().setProperty(CoreConfig.EVENTS_MODULES.name(), "");
+        System.clearProperty(CoreConfig.DISCOVERY_MODULES.name());
+        System.clearProperty(CoreConfig.ENUMS_DISCOVERY_MODULES.name());
+        System.clearProperty(CoreConfig.EVENTS_MODULES.name());
```
izrik: Calling `Configuration.getInstance().init()` might be simpler instead. It has the added benefit that, if someone in the future changes a different property during a test, they won't have to remember to also reset it. See [here](https://github.com/rackerlabs/blueflood/blob/master/blueflood-core/src/test/java/com/rackspacecloud/blueflood/service/LocatorFetchRunnableTest.java#L83-L86). This makes tests much more independent.

blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/http/HttpIntegrationTestBase.java
izrik: @VinnyQ Sounds good to me.

izrik:
Other than that, looks ok to me.

blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/outputs/handlers/HttpMetricTokensHandlerIntegrationTest.java
```diff
@@ -0,0 +1,110 @@
...
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Integration Tests for GET .../metric_name/search
+ */
+public class HttpMetricTokensHandlerIntegrationTest extends HttpIntegrationTestBase {
```
shintasmith: Minor nit pick, should this be named HttpMetricTokenSearchHandlerIntegrationTest ?

blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/outputs/handlers/HttpMetricsIndexHandlerIntegrationTest.java
```diff
@@ -0,0 +1,120 @@
...
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Integration Tests for GET .../metrics/search
+ */
+public class HttpMetricsIndexHandlerIntegrationTest extends HttpIntegrationTestBase {
```
shintasmith: What is the difference between this and the TokensHandler one above?

blueflood-http/src/main/java/com/rackspacecloud/blueflood/http/HttpResponder.java
```diff
@@ -29,11 +31,20 @@
...
     public static void respond(ChannelHandlerContext ctx, HttpRequest req, HttpResponseStatus status) {
         respond(ctx, req, new DefaultHttpResponse(HTTP_1_1, status));
     }
 
     public static void respond(ChannelHandlerContext ctx, HttpRequest req, HttpResponse res) {
+
+        // set response headers
+        if (ALLOWED_ORIGINS != null && !ALLOWED_ORIGINS.isEmpty()) {
+            res.setHeader("Access-Control-Allow-Origin", ALLOWED_ORIGINS);
+        }
```
shintasmith: 👍

blueflood-http/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/HttpEventsQueryHandler.java
shintasmith: This is actually much cleaner. I like it 👍

blueflood-http/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/HttpMetricDataQueryServer.java
shintasmith: I think I prefer specific paths, although if there's a strong benefit to using regex, I'm open to it.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/CoreConfig.java
goru97: @VinnyQ This setting should also support reqular expressions. eg if a url is prepended with ```/```, it would be treated as a regular expression. Default should be *. To enable CORS there should be an explicit boolean property  ```CORS_ENABLED``` For more reference see how ES has these settings here: https://www.elastic.co/guide/en/elasticsearch/reference/1.4/modules-http.html

VinnyQ: @goru97 do we take the regular expression from the config and translate to it to the response header values somehow to be returned?  If so, can you provide me with an example?  Or would it be that if a regular expression is provided in the config, that same regular expression would be returned in the response header (i.e. for Access-Control-Allow-Origin).

blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/outputs/handlers/HttpMetricTokensHandlerIntegrationTest.java
VinnyQ: No.  All the output.handlers integration test classes name in this package are in the name format XXXHandlerClassNameIntegrationTest.  I organized it as integration tests for each specific handler, so for this one it's integration tests for the HttpMetricTokensHandler class.

shintasmith: Got it. Thanks!

blueflood-http/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/HttpOptionsHandler.java
```diff
@@ -0,0 +1,42 @@
...
+        // set CORS headers in the response
+        Map<String, String> headers = new HashMap<String, String>();
+        if (CORS_ENABLED) {
+            headers.put("Access-Control-Allow-Origin", CORS_ALLOWED_ORIGINS);
```
goru97: @VinnyQ Never mind my previous comment. It will make it complicated and we can postpone that feature. But, here we are missing the essence of CORS. eg if someone sets ```CORS_ALLOWED_ORIGINS``` property to ```http://*.rackspace.com``` With the above code, if a request comes from anywhere, we are returning ```Access-Control-Allow-Origin: http://*.rackspace.com``` as response to every single request. 
There should be a method here that should essentially validate the requester's origin and see if it matches with the ```CORS_ALLOWED_ORIGINS``` property.
eg if a request comes from ```http://foo.rackspace.com``` and the ```CORS_ALLOWED_ORIGINS``` property is set to ```http://*.rackspace.com``` then the method should return ```Access-Control-Allow-Origin: http://foo.rackspace.com``` as response headers, along with other headers. Same is the case for ```http://bar.rackspace.com```
This is how you can get the the request origin ```String host = request.getRemoteHost(); ```
Do the pattern matching on that and send the response.
.
.
.
Now this sounds like a lot of work, isn't it and will involve a lot of testing for all the cases.
How about we leverage the features provided by netty server. Unfortunately we are running an older version of netty. Probably 3.x. If we upgrade our server to Version 4.10 That comes with CORS support and you won't have to do all this manual work :blush:

VinnyQ: Story created to use netty CorsHandler:  https://jira.rax.io/browse/CMD-1249

PR #722, opened 2016-07-27
adding more metrics for rollup

shintasmith:
The main purpose of this PR is to add a few more metrics to Rollup process. Changes include:
* add the following metric:
  * meter for number of Rollup writes (to record num of Rollup writes per min rate)
* minor clarifications on some log statements

### DIFF
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DAbstractMetricsRW.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DAbstractMetricsRW.java
@@ -48,12 +48,6 @@ public abstract class DAbstractMetricsRW extends AbstractMetricsRW {
     protected abstract DAbstractMetricIO getIO( String rollupType, Granularity gran );
 
     /**
-     * Returns the Timer.Context object for timing the waits of the write results
-     * @return
-     */
-    protected abstract Timer.Context startWaitResultsTimer();
-
-    /**
      * This method inserts a collection of metrics in the
      * {@link com.rackspacecloud.blueflood.service.SingleRollupWriteContext}
      * objects to the appropriate Cassandra column family
@@ -85,18 +79,15 @@ public abstract class DAbstractMetricsRW extends AbstractMetricsRW {
             }
 
             for (final ResultSetFuture future : futureLocatorMap.keySet()) {
-                final SingleRollupWriteContext writeContext = futureLocatorMap.get(future);
-                Timer.Context resultsWaitTimer = startWaitResultsTimer();
                 try {
                     future.getUninterruptibly().all();
                 } catch (Exception ex) {
                     Instrumentation.markWriteError();
+                    final SingleRollupWriteContext writeContext = futureLocatorMap.get(future);
                     LOG.error(String.format("error writing locator %s, granularity %s, timestamp %d",
                                         writeContext.getLocator(), writeContext.getGranularity(),
                                         writeContext.getTimestamp()),
                                 ex);
-                } finally {
-                    resultsWaitTimer.stop();
                 }
             }
         } finally {
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DAbstractMetricsRW.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DAbstractMetricsRW.java
@@ -5,7 +5,6 @@ import com.datastax.driver.core.ResultSetFuture;
 import com.google.common.collect.Table;
 import com.rackspacecloud.blueflood.cache.MetadataCache;
 import com.rackspacecloud.blueflood.exceptions.CacheException;
-import com.rackspacecloud.blueflood.exceptions.InvalidDataException;
 import com.rackspacecloud.blueflood.io.*;
 import com.rackspacecloud.blueflood.outputs.formats.MetricData;
 import com.rackspacecloud.blueflood.rollup.Granularity;
@@ -18,7 +17,7 @@ import java.io.IOException;
 import java.util.*;
 
 /**
- * This class deals with aspects of reading/writing metrics which are common accross all column families
+ * This class deals with aspects of reading/writing metrics which are common across all column families
  * using Datastax driver
  */
 public abstract class DAbstractMetricsRW extends AbstractMetricsRW {
@@ -26,7 +25,6 @@ public abstract class DAbstractMetricsRW extends AbstractMetricsRW {
     private static final Logger LOG = LoggerFactory.getLogger( DAbstractMetricsRW.class );
 
     protected final LocatorIO locatorIO;
-
 
     /**
      * Constructor
@@ -50,6 +48,18 @@ public abstract class DAbstractMetricsRW extends AbstractMetricsRW {
     protected abstract DAbstractMetricIO getIO( String rollupType, Granularity gran );
 
     /**
+     * Returns the Timer.Context object for timing the Datastax write
+     * @return
+     */
+    protected abstract Timer.Context startWriteTimer();
+
+    /**
+     * Returns the Timer.Context object for timing the waits of the write results
+     * @return
+     */
+    protected abstract Timer.Context startWaitResultsTimer();
+
+    /**
      * This method inserts a collection of metrics in the
      * {@link com.rackspacecloud.blueflood.service.SingleRollupWriteContext}
      * objects to the appropriate Cassandra column family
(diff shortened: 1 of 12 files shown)


blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/RollupRunnable.java
```diff
@@ -165,6 +165,7 @@ public void run() {
                     singleRollupReadContext.getRange().toString(),
                     srcGran.name(),
                     e});
+            executionContext.markUnsuccessful(e);
```
shintasmith: @izrik , @VinnyQ , do you think this would've fixed CMD-1149 ?

shintasmith: The above addition of the ```executionContext.markUnsuccessfull()``` had bad effects, lots of tests are failing. I'm removing it.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DAbstractMetricsRW.java
```diff
@@ -66,28 +76,43 @@ public void insertRollups(List<SingleRollupWriteContext> writeContexts) {
...
+                        LOG.error(String.format("error writing locator %s, granularity %s, timestamp %d",
+                                        writeContext.getLocator(), writeContext.getGranularity(),
+                                        writeContext.getTimestamp()),
+                                ex);
+                    } finally {
+                        resultsWaitTimer.stop();
+                    }
                 }
+            } finally {
+
```
usnavi: is this try/finally block necessary?

ChandraAddala: Whats with the empty finally block?

contrib/grinder/scripts/ingest.py
```diff
@@ -93,7 +93,8 @@ def make_request(self, logger):
                                         self.slice[self.position])
         self.position += 1
         result = self.request.POST(self.ingest_url(), payload)
+        if result.getStatusCode() == 500:
```
ChandraAddala: @shintasmith @usnavi This code is moved to a new repo now right? We have to remove the grinder folder from blueflood or retrofit new changes to new repo till we remove this folder from blueflood.

ChandraAddala:
Other than the couple of comments I had, it looks good to me.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DAbstractMetricsRW.java
```diff
@@ -50,6 +48,18 @@ protected DAbstractMetricsRW(LocatorIO locatorIO) {
...
+     */
+    protected abstract Timer.Context startWriteTimer();
+
+    /**
+     * Returns the Timer.Context object for timing the waits of the write results
+     * @return
+     */
+    protected abstract Timer.Context startWaitResultsTimer();
+
+    /**
```
ChandraAddala: By the way they look, the look kind of similar but there is quite a bit of distinction. `startWriteTimer` is for batch of records whereas `startWaitResultsTimer` is for individual write times. 

Also `startWriteTimer` is not always for a batch of same size. The batch size depends on number of threads available, ROLLUP_BATCH_MIN_SIZE, ROLLUP_BATCH_MAX_SIZE. If the batch size is not consistent, I doubt if this timer would be of any use. Atleast I am concerned if anyone can keep all these into account when looking at graphs.

Overall, I am not a big fan of `startWriteTimer` timer.

ChandraAddala:
I just thought about something and added few more comments. I am taking back my +1 :).

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DAbstractMetricsRW.java
```diff
@@ -50,6 +48,18 @@ protected DAbstractMetricsRW(LocatorIO locatorIO) {
     protected abstract DAbstractMetricIO getIO( String rollupType, Granularity gran );
 
     /**
+     * Returns the Timer.Context object for timing the Datastax write
+     * @return
```
ChandraAddala: How relevant are these timers with Astyanax?

contrib/grinder/scripts/ingest.py
usnavi: yeah, we'll need to sync up with @izrik on this.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DAbstractMetricsRW.java
shintasmith: You are right that there's a distinction here. I initially made ```startWaitResultsTimer``` to count the time it took for the whole loop to complete. But then I realized I'm just recording the same thing as the ```Batch Write Timer``` that's already existing. So I decided to record individual call to ```future.getUninterruptibly().all()```.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DAbstractMetricsRW.java
shintasmith: Not relevant. The model for Astyanax is not the same as Datastax. In Datastax, we fire a bunch of Insert statements, and loop over individual ```ResultSetFuture```.

In Astyanax, we prepare a MutationBatch with a bunch of Insert statements, and then we call ```mutationBatch.execute()```. 

The programming model is different.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DAbstractMetricsRW.java
shintasmith: I love empty lines of code????

I'll remove it

contrib/grinder/scripts/ingest.py
shintasmith: So you guys think I should just remove this change?

shintasmith:
@ChandraAddala @usnavi , incorporated all your feedback:
* remove the write-only timer in DAbstractMetricsRW
* remove empty finally block
* remove changes to grinder (will make it when @izrik publishes the new location)
* explicitly setConsistencyLevel() to ONE on all Insert statements to make maven-cassandra plugin happy

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DMetricsCFPreparedStatements.java
```diff
@@ -97,22 +98,28 @@ private DMetricsCFPreparedStatements() {
         //
         insertToMetricsPreaggrFullStatement = session.prepare(
                 String.format(INSERT_KEY_COLUMN_VALUE_TTL_FORMAT,
-                        CassandraModel.CF_METRICS_PREAGGREGATED_FULL_NAME));
+                        CassandraModel.CF_METRICS_PREAGGREGATED_FULL_NAME))
+                .setConsistencyLevel(ConsistencyLevel.ONE); // needed by maven cassandra plugin
```
ChandraAddala: Isn't `LOCAL_ONE` the default consistency level for datastax? I understand that the plugin requires it for tests but what happens when we deploy it?

shintasmith: For our purpose, LOCAL_ONE and ONE is the same. LOCAL_ONE goes only to local DC. ONE can go to non-local DC. We only have one DC. So it should be the same.

shintasmith: But that reminds me. We need to upgrade our cassandra maven plugin at some point, so we can remove this setConsistencyLevel() call

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/RollupBatchWriteRunnable.java
```diff
@@ -51,11 +53,12 @@ public void run() {
         } catch (Exception e) {
             LOG.warn("not able to insert rollups", e);
             executionContext.markUnsuccessful(e);
-
+        } finally {
+            executionContext.decrementWriteCounter(writeContexts.size());
+            rollupsPerBatch.update(writeContexts.size());
+            rollupsWriteRate.mark(writeContexts.size());
```
ChandraAddala: I wanted to ask you this earlier. It feels redundant to have timer, meter and histogram around this piece of code. What do you think?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DMetricsCFPreparedStatements.java
ChandraAddala: I am worried no one is going to remember to go back and change this piece of code when we go multi-DC.

usnavi: How about we put the same comment as it found here:  https://github.com/rackerlabs/blueflood/blob/master/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DRawIO.java#L62

We had to do this for multiple DS insert statements.  Having the same TODO will ensure that when we change one, we'll see this as well.

shintasmith: I can add the TODO

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/RollupBatchWriteRunnable.java
shintasmith: @ChandraAddala , I think they serve different purpose. 

The new Meter I added: rollupsWriteRate, measures the rate of locators write per minute. In the dashboard, I am graphing the m1_rate. This is essentially our Rollup Write throughput. 
http://graphite00.dfw.perf01.bf.k1k.me/grafana/#/dashboard/db/rollup-dashboard?panelId=14&fullscreen&from=now-4d&to=now

The old Histogram rollupsPerBatch, does not let me get the m1_rate, since it's a Histogram. We can plot mean, max, p95, p99 from this metric:
http://graphite00.dfw.perf01.bf.k1k.me/grafana/#/dashboard/db/rollup-dashboard?panelId=38&fullscreen&from=now-4d&to=now

The Timer measures the time of this block. It does let you graph m1_rate, but it's the rate of the method call. There is no way for me to add the size of the list. 

Out of the 3 metrics, I'd say I need the Meter and the Timer the most. The Histogram doesn't provide me with too much info, although I think it's useful too.

Do you think I should remove the Histogram?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DMetricsCFPreparedStatements.java
shintasmith: @usnavi I added the TODOs

shintasmith:
@ChandraAddala @usnavi I removed the timer around ```future.getUninterruptibly().all()```

ChandraAddala:
+1

PR #723, opened 2016-08-05
Removing old finder code

ChandraAddala:
We had placed the finder code into its own repository(https://github.com/rackerlabs/blueflood-graphite-finder) a while back and the current heat template uses the new repository. 

Removing the old code to avoid confusion.


usnavi:
removing grinder or finder?  (see title) :-P

ChandraAddala:
grinder, finder they are all same...

VinnyQ:
👍  .. looks good, but are we certain no one is referencing this in their template/blog/scripts at the moment?

shintasmith:
@ChandraAddala do we still need this change? If so, you'll need to rebase.

PR #725, opened 2016-08-11
change graphite-api source for pip install in Dockerfile

VinnyQ:
# What
Modify the url for installing graphite-api in the Docker scripts.

# Why
we reforked from the official 1.1.3 version of brutasse's graphite-api and added our own changes and tagged it to 1.1.3-rax.1

### DIFF
--- a/contrib/graphite/scripts/bootstrap.sh
+++ b/contrib/graphite/scripts/bootstrap.sh
@@ -95,7 +95,7 @@ ln -s /etc/nginx/sites-available/grafana /etc/nginx/sites-enabled/grafana
 /etc/init.d/nginx restart
 
 pip install gunicorn
-pip install --upgrade "git+http://github.com/rackerlabs/graphite-api.git@george/fetch_multi_with_patches"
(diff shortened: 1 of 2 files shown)


shintasmith:
lgtm, but let's have @ChandraAddala or @usnavi check it too

contrib/graphite/scripts/bootstrap.sh
```diff
@@ -95,7 +95,7 @@ ln -s /etc/nginx/sites-available/grafana /etc/nginx/sites-enabled/grafana
 /etc/init.d/nginx restart
```
ChandraAddala: I had never used this file. Do you what what it does?

VinnyQ: I have no idea either, I just seach for and replace the reference.  My guess is that this is an old version of dockerfile for blueflood... i think this is how the original docker setup got started maybe?  Similar to how graphite-api and finder was started?  We can probably delete this out of blueflood/contrib eventually (along with the contrib/graphite).

PR #726, opened 2016-08-12
upgrade to netty 4.x

shintasmith:
Update code base to consume the proper netty 4.0.x classes. This upgrades blueflood code base to use netty 4.0.40.FINAL, which is the latest 4.0.x netty up to date. There is also netty 4.1.4.FINAL but since datastax driver 3.0.0 is pulling in netty 4.0.x, I figure it's safer to use netty 4.0.x.

The main goal of this upgrade is so that datastax can use the Native epoll transport, which is supposed to perform better.


blueflood-http/src/main/java/com/rackspacecloud/blueflood/http/QueryStringDecoderAndRouter.java
```diff
@@ -16,51 +16,54 @@
...
+        router.route(ctx, HttpRequestWithDecodedQueryParams.create(msg));
     }
 
     @Override
-    public void exceptionCaught(ChannelHandlerContext ctx, ExceptionEvent e) {
-        if (e.getCause() instanceof IllegalArgumentException) {
-            if ("empty text".equals(e.getCause().getMessage())) {
+    public void exceptionCaught(ChannelHandlerContext ctx, Throwable thr) {
+        if (thr.getCause() instanceof IllegalArgumentException) {
+            if ("empty text".equals(thr.getCause().getMessage())) {
```
usnavi: wonderful! :-P  I think its fine since it was already there, but if the message changes (like between versions) it break.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/Constants.java
```diff
@@ -66,7 +65,7 @@
             throw new RuntimeException("Invalid FullResSerializer.CUR_VERSION. Please increment until this exception does not happen.");
     }
 
-    public static final Charset DEFAULT_CHARSET = CharsetUtil.UTF_8;
+    public static final Charset DEFAULT_CHARSET = Charset.forName("UTF-8");
```
usnavi: I think this still exists in 4.0 https://netty.io/4.0/api/io/netty/util/CharsetUtil.html  Looks like you use it later https://github.com/rackerlabs/blueflood/pull/726/files#diff-02100b1d3f0794dc57e4870d8bb80873R63

Was there a different reason to remove it?

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionServer.java
```diff
@@ -95,72 +110,64 @@ public void startServer() {
...
-            pipeline.addLast("encoder2", new HttpResponseDecoder());
-            pipeline.addLast("handler", new QueryStringDecoderAndRouter(router));
-
-            return pipeline;
-        }
+            }
+        });
+        pipeline.addLast("inflater", new HttpContentDecompressor());
+        pipeline.addLast("chunkaggregator", new HttpObjectAggregator(Constants.MAX_CONTENT_LENGTH));
+        pipeline.addLast("encoder2", new HttpResponseDecoder());
```
usnavi: can we change this to just "encoder"?

usnavi: nm

blueflood-http/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/HttpMetricDataQueryServer.java
```diff
@@ -77,36 +76,52 @@ public void startServer() {
...
+                    // If we are getting exception trying to write,
+                    // don't propagate to caller. It may cause this
+                    // method to be called again and will produce
+                    // stack overflow. So just log it here.
+                    log.debug("Can't write to channel " + ctx.channel().toString(), ex);
+                }
+            }
+        });
+        pipeline.addLast("chunkaggregator", new HttpObjectAggregator(Constants.MAX_CONTENT_LENGTH));
+        pipeline.addLast("handler", new QueryStringDecoderAndRouter(router));
```
usnavi: As a future story, would we want to pull up the common code across the different servers into a parent class?  Seems like a fair amount of duplication.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/http/QueryStringDecoderAndRouter.java
shintasmith: Yeah... That if statement is probably old. ```IllegalArgumentException``` probably would've come from our code, but I don't see any of our code throwing such message. Should I remove it?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/Constants.java
shintasmith: I dunno why but I missed it. I will update to use Netty's

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionServer.java
shintasmith: I can change it to be respdecoder. There is another encoder and decoder up on top.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/http/QueryStringDecoderAndRouter.java
usnavi: Yeah, if you think its OK, let's do it.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionServer.java
usnavi: this is fine.  I eventually noticed the other one on above.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/HttpMetricDataQueryServer.java
shintasmith: There's a few things we can refactor in this area, but the pipeline probably doesn't need to.

We basically have 2 types of servers: 
* the Ingest server (HttpMetricsIngestServer.java)
* the Query server (HttpMetricDataQueryServer.java)

The pipeline for those actually look slightly differently. The Ingest deals with HttpContentDecompressor (for handling gzip request payload). The Query doesn't have to deal with that, because most Query requests don't have big request payload.

The part that needs refactoring is the "handler", the QueryStringDecoderAndRouter. Both Ingest and Query servers use it, but probably only Query makes sense. We just use it for the Ingest server for the "router" part, not the "queryStringDecoder" part. That class breaks the single responsibility principle. 

But I feel this part is not our main component. I kinda think we shouldn't be doing our own routing anyways. So I leave it alone. If someone is so bored, they can refactor this part. Otherwise, I think we have bigger fish to fry.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/http/Constants.java
```diff
@@ -0,0 +1,12 @@
...
+
+/**
+ * A class to hold constants used by Blueflood HTTP subsystem
+ */
+public class Constants {
+
+    /**
+     * The maximum length of a request's HTTP body
+     */
+    public static final int MAX_CONTENT_LENGTH = 1048576; // 1 MB
```
ChandraAddala: Should we make this configurable? I mean keep it in CoreConfig?

blueflood-http/src/main/java/com/rackspacecloud/blueflood/http/HttpRequestWithDecodedQueryParams.java
```diff
@@ -0,0 +1,163 @@
...
+/**
+ * This class is a special type of {@link io.netty.handler.codec.http.HttpRequest}
+ * that knows how to decode the query parameters.
+ */
+public class HttpRequestWithDecodedQueryParams implements FullHttpRequest {
+
+    private Map<String, List<String>> queryParams;
+    protected FullHttpRequest request;
+
+    protected HttpRequestWithDecodedQueryParams(FullHttpRequest request, Map<String, List<String>> queryParams) {
```
ChandraAddala: I am little confused here. If I create an instance of `HttpRequestWithDecodedQueryParams` and then using that instance, if I do `getQueryParams`, would I get decoded query params?

ChandraAddala:
Whats the main reason to switch from `HttpRequest` to `FullHttpRequest`?

blueflood-http/src/main/java/com/rackspacecloud/blueflood/outputs/handlers/HttpMetricDataQueryServer.java
```diff
@@ -77,36 +76,52 @@ public void startServer() {
...
+                                            .addListener(ChannelFutureListener.CLOSE);
+                    } else {
+                        log.debug("channel " + ctx.channel().toString() + " is no longer writeable, not sending 400 response back to client");
+                    }
+                } catch (Exception ex) {
+                    // If we are getting exception trying to write,
+                    // don't propagate to caller. It may cause this
+                    // method to be called again and will produce
+                    // stack overflow. So just log it here.
+                    log.debug("Can't write to channel " + ctx.channel().toString(), ex);
```
ChandraAddala: Should this be error?

shintasmith: This just means the client has closed the connection before we get a chance to reply. If this happens a lot, which I think it did, it would fill up logs.

shintasmith: The main reason for switching from ```HttpRequest``` to ```FullHttpRequest``` is the ```HttpRequest``` interface has changed from Netty 3.x to Netty 4.x. 

In Netty 3.x, ```HttpRequest``` and ```HttpResponse``` are the only interfaces to use when dealing with Http requests and responses. From those interfaces, you can get the headers, the body, all necessary operations of Http, in those 2 interfaces.

In Netty 4.x, they have changed the interfaces to be smaller and more granular. ```HttpRequest``` only has methods to deal with headers -- that's all. You can't access the request body. To do that, you'll have to use ```FullHttpRequest``` interface.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/http/Constants.java
shintasmith: Yeah, I should probably do that. Let me move it

shintasmith:
@ChandraAddala I moved the max content length in HttpConfig

ChandraAddala:
Looks good.

PR #727, opened 2016-08-15
Datastax rollup writes changed from using async to batch statements

ChandraAddala:
We had noticed that switching from datastax to astyanax has increased the rollup write time by 5x. The one major difference between datastax and astyanax with respect to rollup writes are in datastax, each write is a separate async call whereas in astyanax, we do a synchronous batch(MutationBatch) request. 

Based on several blogs(see References section) which talk about this, have found that using "unlogged batch statements" with datastax tend to have better performance than using async even though the documention clearly says to use "unlogged batch statements" only when we can batch them by partition key to see any benefits.

### DIFF
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DAbstractMetricIO.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DAbstractMetricIO.java
@@ -77,6 +77,28 @@ public abstract class DAbstractMetricIO {
         }
     }
 
+    public Statement createStatement(Locator locator, long collectionTime, Rollup rollup, Granularity granularity, int ttl) {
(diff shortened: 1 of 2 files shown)


usnavi:
no write up?  booo!!!!!!!!!

shintasmith:
👍 Other than the failing test, changes look good.

usnavi:
once there's a description, lgtm.

ChandraAddala:
I thought since discussed so much about it and we also have wiki page, I didnt bother adding writeup. Here is the writeup https://one.rackspace.com/pages/viewpage.action?title=Datastax+async+vs+unlogged+batches&spaceKey=cloudmetrics

usnavi:
mostly its for archival purposes.  If someone comes back to this PR a year from now, they might understand the context.

PR #728, opened 2016-08-18
track delayed metrics when they are sent to /aggregated and /aggregated/multi

shintasmith:
This PR adds the ability to track/log delayed metrics when they are sent to /ingest/aggregated and /ingest/aggregated/multi API. The same functionality exists for /ingest and /ingest/multi.

### DIFF
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/formats/AggregatedPayload.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/formats/AggregatedPayload.java
@@ -16,6 +16,8 @@
 
 package com.rackspacecloud.blueflood.inputs.formats;
 
+import com.google.gson.Gson;
+import com.rackspacecloud.blueflood.io.Instrumentation;
 import com.rackspacecloud.blueflood.service.Configuration;
 import com.rackspacecloud.blueflood.service.CoreConfig;
 import com.rackspacecloud.blueflood.types.*;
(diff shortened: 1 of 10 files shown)


shintasmith:
@usnavi @ChandraAddala @VinnyQ @izrik pls review. This is to add the logging/tracking delayed metrics for /ingest/aggregated and /ingest/aggregated/multi.

shintasmith:
I just realized this PR is only doing the Tracker part. But does not update the "short delay" and "long delay" graphs. I will have to add those on Monday.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/formats/AggregatedPayload.java
```diff
@@ -74,6 +75,52 @@ public String toString() {
...
+        }
+    }
+
+    public long getDelayTime(long ingestTime) {
+        return ingestTime - timestamp;
+    }
+
+    public List<String> getAllMetricNames() {
+        List<String> metricNames = new java.util.ArrayList<String>();
+        if ( gauges != null && gauges.length > 0) {
```
usnavi: I wonder if we should start thinking functionally for some of this stuff since we can now use java 8.  No need to change.

usnavi:
lgtm

blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/formats/AggregatedPayload.java
shintasmith: I was very tempted to use java 8 stuff. It would've been much cleaner. But that's a decision that I think have external impacts. We need to communicate that better. Can't just cut the cord off, I think...

I even have to change ```String.join()``` to something else, because it's only available on Java 8.

ChandraAddala:
Is it possible to populate the "Metrics with short/long delay received" graphs also? https://github.com/rackerlabs/blueflood/blob/e6de32ff53ed32cc348adbff8ae3692415820e1d/blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/formats/JSONMetricsContainer.java#L99

ChandraAddala:
oh, I just saw ur comment about these graphs.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpAggregatedIngestionHandler.java
```diff
@@ -69,6 +69,14 @@ public void handle(ChannelHandlerContext ctx, FullHttpRequest request) {
 
             AggregatedPayload payload = createPayload( body );
 
+            long ingestTime = System.currentTimeMillis();
```
ChandraAddala: Do u think you can use Clock to get current time?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/formats/AggregatedPayload.java
```diff
@@ -74,6 +75,52 @@ public String toString() {
         return errors;
     }
 
+    public boolean hasDelayedMetrics(long ingestTime) {
+        if ( getDelayTime(ingestTime) > TRACKER_DELAYED_METRICS_MILLIS ) {
```
ChandraAddala: Its probably a matter of preference, but you can just do `return getDelayTime(ingestTime) > TRACKER_DELAYED_METRICS_MILLIS`

shintasmith:
@ChandraAddala , I added the marking of the meter for short and long delay graphs and incorporated your feedback. Pls take a look.

ChandraAddala:
Looks good.

PR #732, opened 2016-09-06
Check if in rollup mode before starting Zookeeper

DanielNi:
Simple sanity check so a node doesn't try to grab ZK locks if it's not actually rolling up.

### DIFF
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/BluefloodServiceStarter.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/BluefloodServiceStarter.java
@@ -306,9 +306,10 @@ public class BluefloodServiceStarter {
         final Collection<Integer> shards = Collections.unmodifiableCollection(
                 Util.parseShards(config.getStringProperty(CoreConfig.SHARDS)));
         final String zkCluster = config.getStringProperty(CoreConfig.ZOOKEEPER_CLUSTER);
-        final ScheduleContext rollupContext = "NONE".equals(zkCluster) ?
-                new ScheduleContext(System.currentTimeMillis(), shards) :
-                new ScheduleContext(System.currentTimeMillis(), shards, zkCluster);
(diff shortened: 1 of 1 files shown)


blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/BluefloodServiceStarter.java
```diff
@@ -306,9 +306,10 @@ public void run() {
         final Collection<Integer> shards = Collections.unmodifiableCollection(
                 Util.parseShards(config.getStringProperty(CoreConfig.SHARDS)));
         final String zkCluster = config.getStringProperty(CoreConfig.ZOOKEEPER_CLUSTER);
-        final ScheduleContext rollupContext = "NONE".equals(zkCluster) ?
-                new ScheduleContext(System.currentTimeMillis(), shards) :
-                new ScheduleContext(System.currentTimeMillis(), shards, zkCluster);
+        final boolean rollupMode = config.getBooleanProperty(CoreConfig.ROLLUP_MODE);
+        final ScheduleContext rollupContext = !"NONE".equals(zkCluster) && rollupMode ?
```
shintasmith: Why was the logic changed to check for zkCluster != "NONE" ?

DanielNi: The logic is the same (the if and else cases are switched). There's just an additional check to make sure Blueflood is in rollup mode.

DanielNi:
@shintasmith ping. We found this useful to make sure non-rollup nodes didn't grab all of the ZK locks, preventing rollups.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/BluefloodServiceStarter.java
ChandraAddala: @shintasmith  and me discussed about it. It looks good.

PR #733, opened 2016-09-13
add locators per shard histogram

shintasmith:
This is to track how many locators we have per shard, over time.

### DIFF
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/LocatorFetchRunnable.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/LocatorFetchRunnable.java
@@ -16,6 +16,7 @@
 
 package com.rackspacecloud.blueflood.service;
 
+import com.codahale.metrics.Histogram;
 import com.codahale.metrics.Timer;
 import com.google.common.annotations.VisibleForTesting;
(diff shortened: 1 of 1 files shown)


shintasmith:
@VinnyQ @ChandraAddala @usnavi @izrik @drenalin23 , do you all think this is useful???

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/LocatorFetchRunnable.java
```diff
@@ -48,6 +49,7 @@
     private ScheduleContext scheduleCtx;
     private long serverTime;
     private static final Timer rollupLocatorExecuteTimer = Metrics.timer(RollupService.class, "Locate and Schedule Rollups for Slot");
+    private static final Histogram locatorsPerShard = Metrics.histogram(RollupService.class, "Locators Per Shard");
```
ChandraAddala: Should the class be `RollupService.class`? I guess we can keep it that way to be consistent.

shintasmith: This class seems to be used by all rollup related metrics. So I was going to keep it that.

ChandraAddala:
Looks good.

usnavi:
lgtm

PR #741, opened 2016-09-26
add querylogger, locator metric, minor log changes

shintasmith:
A few minor changes I'd like to add, as a result of the latest performance testing:
* add metrics to track the rate and timing of locator inserts
We currently do not track this. As I found in the performance tests, how many locators we are inserting impacts the throughput. This will be a useful metric to track
* add QueryLogger (which is not the same thing as QueryTrace). See also https://datastax.github.io/java-driver/3.0.3/manual/logging/
This will enable us to turn on logging at client/driver side, which queries/statements are in ERROR, SLOW, or NORMAL. These 3 categories of logging can be turned on independently from log4j, by having this in your log4j.properties:

```
log4j.logger.com.datastax.driver.core.QueryLogger.ERROR=DEBUG
log4j.logger.com.datastax.driver.core.QueryLogger.SLOW=DEBUG
```

* fix minor out of order logging statement (in RollupRunnable.java)

### DIFF
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/astyanax/ALocatorIO.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/astyanax/ALocatorIO.java
@@ -46,12 +46,17 @@ public class ALocatorIO implements LocatorIO {
      */
     @Override
     public void insertLocator(Locator locator) throws IOException {
+        Timer.Context timer = Instrumentation.getWriteTimerContext(CassandraModel.CF_METRICS_LOCATOR_NAME);
         try {
             MutationBatch mutationBatch = AstyanaxIO.getKeyspace().prepareMutationBatch();
             AstyanaxWriter.getInstance().insertLocator(locator, mutationBatch);
             mutationBatch.execute();
         } catch (Exception e) {
             throw new IOException(e);
+        } finally {
+            if ( timer != null ) {
+                timer.stop();
+            }
         }
     }
 
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/astyanax/ALocatorIO.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/astyanax/ALocatorIO.java
@@ -54,9 +54,7 @@ public class ALocatorIO implements LocatorIO {
         } catch (Exception e) {
             throw new IOException(e);
         } finally {
-            if ( timer != null ) {
-                timer.stop();
-            }
+            timer.stop();
         }
     }
 
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DLocatorIO.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DLocatorIO.java
@@ -82,7 +82,7 @@ public class DLocatorIO implements LocatorIO {
                 .value(VALUE, bindMarker());
         putValue = DatastaxIO.getSession()
                 .prepare(insert)
-                .setConsistencyLevel( ConsistencyLevel.LOCAL_ONE );  // TODO: remove later; required by the cassandra-maven-plugin 2.0.0-1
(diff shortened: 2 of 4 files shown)


shintasmith:
@usnavi @ChandraAddala @VinnyQ @drenalin23 @izrik pls review

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/astyanax/ALocatorIO.java
```diff
@@ -46,12 +46,17 @@
...
     public void insertLocator(Locator locator) throws IOException {
+        Timer.Context timer = Instrumentation.getWriteTimerContext(CassandraModel.CF_METRICS_LOCATOR_NAME);
         try {
             MutationBatch mutationBatch = AstyanaxIO.getKeyspace().prepareMutationBatch();
             AstyanaxWriter.getInstance().insertLocator(locator, mutationBatch);
             mutationBatch.execute();
         } catch (Exception e) {
             throw new IOException(e);
+        } finally {
+            if ( timer != null ) {
```
ChandraAddala: why do you need this null check?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxIO.java
```diff
@@ -57,13 +57,19 @@ private static void connect() {
         CodecRegistry codecRegistry = new CodecRegistry();
 
         cluster = Cluster.builder()
-                .withLoadBalancingPolicy(new TokenAwarePolicy(DCAwareRoundRobinPolicy.builder().withLocalDc("datacenter1").build()))
+                .withLoadBalancingPolicy(new TokenAwarePolicy(DCAwareRoundRobinPolicy.builder().withLocalDc("datacenter1").build(), false))
```
ChandraAddala: What is this false for?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DLocatorIO.java
```diff
@@ -101,7 +101,12 @@ public void insertLocator(Locator locator) throws IOException {
 
         // bound values and execute
         BoundStatement bs = putValue.bind(shard, locator.toString(), "");
-        session.execute(bs);
+        Timer.Context timer = Instrumentation.getWriteTimerContext(CassandraModel.CF_METRICS_LOCATOR_NAME);
+        try {
+            session.execute(bs);
```
ChandraAddala: May be the ``BoundStatement`` part also should be part of the timer execution?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/astyanax/ALocatorIO.java
shintasmith: Oh, I moved the code that instantiates the timer. I don't need it anymore. I will remove.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxIO.java
shintasmith: This is to turn off shuffleReplica flag, per documentation here:
http://docs.datastax.com/en/drivers/java/3.0/com/datastax/driver/core/policies/TokenAwarePolicy.html

This is recommended by the guys from Last Pickle (see slide 25)
http://www.slideshare.net/planetcassandra/successful-software-development-with-apache-cassandra-54922569

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DLocatorIO.java
shintasmith: I will change that

shintasmith:
@ChandraAddala I addressed your comments.

PR #746, opened 2016-10-06
Delayed Metrics tracking

ChandraAddala:
- Delayed Metrics ingest 
 - Created new column family for tracking delayed metrics by shard, slot and configurable granularity
 - Configurable way to turn tracking delayed metrics on/off
 - Adding this feature for both datastax and astyanax drivers
 - Tracking delayed metrics for both basic and pre-aggregated metrics
 - Caching of delayed metrics in memory to avoid repeated updates for frequently ingested locators
 - Separate IO classes to read/write from metrics_delayed_locator column family;
 - Created separate histogram for locator's rolling up during re-rolls

- Delayed Metrics rollup
     - Code changes to only rollup delayed locators when re-rolling a slot.
     - Created/Refactored isReroll method in ScheduleContext

### DIFF
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/cache/LocatorCache.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/cache/LocatorCache.java
@@ -14,17 +14,14 @@ public class LocatorCache {
 
     // this collection is used to reduce the number of locators that get written.
     // Simply, if a locator has been seen within the last 10 minutes, don't bother.
-    private final Cache<String, Boolean> insertedLocators =
-            CacheBuilder.newBuilder().expireAfterAccess(10,
-                    TimeUnit.MINUTES).concurrencyLevel(16).build();
+    private final Cache<String, Boolean> insertedLocators;
 
     // this collection is used to reduce the number of delayed locators that get
     // written per slot. Simply, if a locator has been seen for a slot, don't bother.
-    private final Cache<String, Boolean> insertedDelayedLocators =
-            CacheBuilder.newBuilder().expireAfterAccess(10,
-                    TimeUnit.MINUTES).concurrencyLevel(16).build();
+    private final Cache<String, Boolean> insertedDelayedLocators;
 
-    private static LocatorCache instance = new LocatorCache();
+    private static LocatorCache instance = new LocatorCache(10, TimeUnit.MINUTES);
+
 
     static {
         Metrics.getRegistry().register(MetricRegistry.name(LocatorCache.class, "Current Locators Count"),
@@ -44,8 +41,25 @@ public class LocatorCache {
                 });
     }
 
+
     public static LocatorCache getInstance() {
         return instance;
+    }
+
+    protected LocatorCache(long expireAfterAccessDuration, TimeUnit expireAfterAccessTimeUnit) {
+
+        insertedLocators =
+                CacheBuilder.newBuilder().expireAfterAccess(expireAfterAccessDuration,
+                        expireAfterAccessTimeUnit).concurrencyLevel(16).build();
+
+        insertedDelayedLocators =
+                CacheBuilder.newBuilder().expireAfterAccess(expireAfterAccessDuration,
+                        expireAfterAccessTimeUnit).concurrencyLevel(16).build();
+    }
+
+    @VisibleForTesting
+    public static LocatorCache getInstance(Long duration, TimeUnit timeUnit) {
+        return new LocatorCache(duration, timeUnit);
     }
 
     public long getCurrentLocatorCount() {
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/cache/LocatorCache.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/cache/LocatorCache.java
@@ -1,0 +1,107 @@
+package com.rackspacecloud.blueflood.cache;
+
+import com.codahale.metrics.Gauge;
+import com.codahale.metrics.MetricRegistry;
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
+import com.rackspacecloud.blueflood.types.Locator;
+import com.rackspacecloud.blueflood.utils.Metrics;
+
+import java.util.concurrent.TimeUnit;
+
+public class LocatorCache {
+
+    // this collection is used to reduce the number of locators that get written.
+    // Simply, if a locator has been seen within the last 10 minutes, don't bother.
+    private final Cache<String, Boolean> insertedLocators =
+            CacheBuilder.newBuilder().expireAfterAccess(10,
+                    TimeUnit.MINUTES).concurrencyLevel(16).build();
+
+    // this collection is used to reduce the number of delayed locators that get
+    // written per slot. Simply, if a locator has been seen for a slot, don't bother.
+    private final Cache<String, Boolean> insertedDelayedLocators =
+            CacheBuilder.newBuilder().expireAfterAccess(10,
+                    TimeUnit.MINUTES).concurrencyLevel(16).build();
+
+    private static LocatorCache instance = new LocatorCache();
+
+    static {
+        Metrics.getRegistry().register(MetricRegistry.name(LocatorCache.class, "Current Locators Count"),
+                new Gauge<Long>() {
+                    @Override
+                    public Long getValue() {
+                        return instance.getCurrentLocatorCount();
+                    }
+                });
+
+        Metrics.getRegistry().register(MetricRegistry.name(LocatorCache.class, "Current Delayed Locators Count"),
+                new Gauge<Long>() {
+                    @Override
+                    public Long getValue() {
+                        return instance.getCurrentDelayedLocatorCount();
+                    }
+                });
+    }
+
+    public static LocatorCache getInstance() {
+        return instance;
+    }
+
+    public long getCurrentLocatorCount() {
+        return insertedLocators.size();
+    }
+
+    public long getCurrentDelayedLocatorCount() {
+        return insertedDelayedLocators.size();
+    }
+
+    /**
+     * Checks if Locator is recently inserted
+     *
+     * @param loc
+     * @return
+     */
+    public synchronized boolean isLocatorCurrent(Locator loc) {
+        return insertedLocators.getIfPresent(loc.toString()) != null;
+    }
+
+    /**
+     * Check if the delayed locator is recently inserted for a given slot
+     *
+     * @param slot
+     * @param locator
+     * @return
+     */
+    public synchronized boolean isDelayedLocatorForASlotCurrent(int slot, Locator locator) {
+        return insertedDelayedLocators.getIfPresent(getLocatorSlotKey(slot, locator)) != null;
+    }
+
+    private String getLocatorSlotKey(int slot, Locator locator) {
+        return slot + "," + locator.toString();
+    }
+
+    /**
+     * Marks the Locator as recently inserted
+     * @param loc
+     */
+    public synchronized void setLocatorCurrent(Locator loc) {
+        insertedLocators.put(loc.toString(), Boolean.TRUE);
+    }
+
+    /**
+     * Marks the delayed locator as recently inserted for a given slot
+     * @param slot
+     * @param locator
+     */
+    public synchronized void setDelayedLocatorForASlotCurrent(int slot, Locator locator) {
+        insertedDelayedLocators.put(getLocatorSlotKey(slot, locator), Boolean.TRUE);
+    }
+
+    @VisibleForTesting
+    public synchronized void resetCache() {
+        insertedLocators.invalidateAll();
+        insertedDelayedLocators.invalidateAll();
+    }
+
+}
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AbstractMetricsRW.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AbstractMetricsRW.java
@@ -16,24 +16,18 @@
 
 package com.rackspacecloud.blueflood.io;
 
-import com.google.common.cache.Cache;
-import com.google.common.cache.CacheBuilder;
 import com.google.common.collect.LinkedListMultimap;
 import com.google.common.collect.Multimap;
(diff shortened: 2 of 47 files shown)


blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/astyanax/AstyanaxWriter.java
```diff
@@ -281,6 +327,10 @@ public void insertMetrics(Collection<IMetric> metrics, ColumnFamily cf) throws C
                             Instrumentation.markFullResPreaggregatedMetricWritten();
                         }
                     }
+
+                    if (locatorInsertOk) {
+                        insertDelayedLocator(metric, batch, isTrackingDelayedMetrics, clock);
```
shintasmith: We only call the regular insertLocator() only if it hasn't been inserted before. At least I think that's what the cache that's used in isLocatorCurrent() is for. Do we want to do the same here for delayed locator?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/astyanax/AstyanaxWriter.java
```diff
@@ -150,10 +169,37 @@ public void insertFull(Collection<? extends IMetric> metrics) throws ConnectionE
...
     // numeric only!
     public final void insertLocator(Locator locator, MutationBatch mutationBatch) {
-                mutationBatch.withRow(CassandraModel.CF_METRICS_LOCATOR, (long) Util.getShard(locator.toString()))
-                        .putEmptyColumn(locator, TenantTtlProvider.LOCATOR_TTL);
+        mutationBatch.withRow(CassandraModel.CF_METRICS_LOCATOR, (long) Util.getShard(locator.toString()))
+                .putEmptyColumn(locator, TenantTtlProvider.LOCATOR_TTL);
+    }
+
+    // numeric only!
+    public final void insertDelayedLocator(Granularity g, int slot, Locator locator, MutationBatch mutationBatch) {
```
shintasmith: You have two methods with the name "insertDelayedLocator", one public, one private, but one behaves quite differently than the other. Can we clarify the difference in the method name?

And why is it that the public one does not check if the delayed locator is current or not?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/CoreConfig.java
```diff
@@ -156,6 +156,10 @@
     TENANTIDS_TO_KEEP(""),
     TRACKER_DELAYED_METRICS_MILLIS("300000"),
 
+    //the granulariy for which we store delayed metrics. Allowed values are 5m, 20m, 60m, 240m, 1440m
```
shintasmith: "the granularity for which we store delayed metrics"
Does this mean we only store delay metrics for granularity 20m ?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/CoreConfig.java
```diff
@@ -156,6 +156,10 @@
...
+    DELAYED_METRICS_STORAGE_GRANULARITY("20m"),
+    ENABLE_TRACKING_DELAYED_METRICS("true"),
```
shintasmith: Perhaps change to ENABLE_PER_SLOT_LOCATOR_ROLLUP_DELAYED_METRICS :-)

blueflood-core/src/test/java/com/rackspacecloud/blueflood/io/IOContainerTest.java
```diff
@@ -59,11 +61,28 @@ public void setup() throws Exception {
...
         when( mockSession.prepare( any( RegularStatement.class ) ) ).thenReturn( mockPreparedStatement );
-        when( mockSession.prepare( anyString() ) ).thenReturn( mockPreparedStatement );
-        when( mockPreparedStatement.setConsistencyLevel( any(ConsistencyLevel.class) ) ).thenReturn( mockPreparedStatement );
+        when( mockSession.prepare( anyString() ) ).thenReturn(mockPreparedStatement);
+        when( mockPreparedStatement.setConsistencyLevel(any(ConsistencyLevel.class)) ).thenReturn( mockPreparedStatement );
+    }
+
+
+    @Test
+    public void testDatastaxDriverConfig() {
```
shintasmith: lol

I guess I will debug this a bit later

src/cassandra/cli/load.cdl
```diff
@@ -15,6 +15,13 @@ CREATE TABLE IF NOT EXISTS "DATA".metrics_locator (
     PRIMARY KEY (key, column1)
 ) WITH COMPACT STORAGE AND speculative_retry = 'NONE';
 
+CREATE TABLE IF NOT EXISTS "DATA".metrics_delayed_locator (
+    key text,
+    column1 text,
+    value text,
+    PRIMARY KEY (key, column1)
+) WITH COMPACT STORAGE AND speculative_retry = 'NONE';
```
shintasmith: This is fine but there's probably a copy of what we actually use in our environments in the blueflood-chef repo.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/astyanax/AstyanaxWriter.java
```diff
@@ -134,7 +148,12 @@ public void insertFull(Collection<? extends IMetric> metrics) throws ConnectionE
                     AstyanaxWriter.setLocatorCurrent(locator);
                 }
 
+                if (!isString && !isBoolean && mutationBatch != null) {
```
usnavi: If there isn't a `mutationbatch` should we create one?  Looks like there isn't a check around `insertMetric(metric, mutationBatch)` below.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/astyanax/AstyanaxWriter.java
```diff
@@ -62,11 +65,22 @@ public static AstyanaxWriter getInstance() {
...
     // this collection is used to reduce the number of locators that get written.  Simply, if a locator has been
     // seen within the last 10 minutes, don't bother.
     private static final Cache<String, Boolean> insertedLocators = CacheBuilder.newBuilder().expireAfterAccess(10,
             TimeUnit.MINUTES).concurrencyLevel(16).build();
 
+    // this collection is used to reduce the number of delayed locators that get written per slot. Simply, if a locator
+    // has been seen for a slot, don't bother.
+    private static final Cache<String, Boolean> insertedDelayedLocators = CacheBuilder.newBuilder().expireAfterAccess(10,
+            TimeUnit.MINUTES).concurrencyLevel(16).build();
+
```
usnavi: If this cache will be used, does it need to be static since we are accessing AstyanaxWriter as singleton?

src/cassandra/cli/load.cdl
usnavi: I think we should have more descriptive names in our table.   I think matching previously poorly-chosen names isn't that helpful.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AbstractMetricsRW.java
```diff
@@ -52,8 +53,16 @@
             CacheBuilder.newBuilder().expireAfterAccess(10,
                         TimeUnit.MINUTES).concurrencyLevel(16).build();
 
+    // this collection is used to reduce the number of delayed locators that get written per slot. Simply, if a locator
+    // has been seen for a slot, don't bother.
+    private static final Cache<String, Boolean> insertedDelayedLocators = CacheBuilder.newBuilder().expireAfterAccess(10,
```
ChandraAddala: `insertedDelayedLocators` is only present in `AbstractMetricsRW` and `AstyanaxWriter` for datastax and astyanax respectively. I dont see it in 'DAbstractMetricsRW'.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AbstractMetricsRW.java
```diff
@@ -52,8 +53,16 @@
...
+            TimeUnit.MINUTES).concurrencyLevel(16).build();
+
     private static final Logger LOG = LoggerFactory.getLogger(AbstractMetricsRW.class);
 
+    protected boolean isTrackingDelayedMetrics;
```
ChandraAddala: When I named it some kinda tracking seemed appropriate. I thought this would track it in the database whereas the one which is already existing would track it in the logs. Also the actual config is `ENABLE_TRACKING_DELAYED_METRICS`

The existing config is `TRACKER_DELAYED_METRICS_MILLIS`.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AbstractMetricsRW.java
```diff
@@ -74,6 +91,10 @@ protected synchronized void setLocatorCurrent(Locator loc) {
         insertedLocators.put(loc.toString(), Boolean.TRUE);
     }
 
+    protected synchronized void setDelayedLocatorForASlotCurrent(int slot, Locator locator) {
```
ChandraAddala: The existing method name to cache locators is `setLocatorCurrent`. I named this method similar to that.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/astyanax/AstyanaxWriter.java
ChandraAddala: I just modelled it after the existing cache `insertedLocators`.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/astyanax/AstyanaxWriter.java
ChandraAddala: The logic to update `metric_delayed_locator` is same as the one above which is for `metric_locator`. Didnt change much there. I felt like string and boolean metrics arent rolled up. Thats why they are not updated in metrics_locator. I am not 100% sure if that is the intention.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/astyanax/AstyanaxWriter.java
ChandraAddala: I will rename one of the methods.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/astyanax/AstyanaxWriter.java
ChandraAddala: We do use cache for inserting delayed locator also similar to the one we have to insert locators. That code is inside that method.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/service/CoreConfig.java
ChandraAddala: Thats correct. We only track at 20m granularity.

ChandraAddala:
@shintasmith If u get a chance can u look at IOContainerTest? Also I had addressed the review comment you had earlier. Pls take a look.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AbstractMetricsRW.java
shintasmith: Maybe I was mistaken for 'DAbstractMetricsRW'... I thought I saw it there too.

But does that mean if we are running Astyanax, then we have 2 caches created? One in AbstractMetricsRW and another in AstyanaxWriter ? Is there a way to avoid it?

blueflood-core/src/test/java/com/rackspacecloud/blueflood/io/IOContainerTest.java
shintasmith: Can you put a TODO here , please, on why you have to move this up? I'll debug it the next time I have to touch this file.

shintasmith:
@ChandraAddala also Travis is failing

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AbstractMetricsRW.java
ChandraAddala: We already have the cache for metrics locator in two places, one for astyanax and one for datastax. I added one more for delayed locators.

ChandraAddala:
@shintasmith @usnavi Can you guys take another look?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AbstractMetricsRW.java
shintasmith: That's not what I meant. Let's vidyo when you get a chance

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/astyanax/AstyanaxWriter.java
```diff
@@ -305,10 +364,22 @@ public static boolean isLocatorCurrent(Locator loc) {
         return insertedLocators.getIfPresent(loc.toString()) != null;
     }
 
+    public static boolean isDelayedLocatorForASlotCurrent(int slot, Locator locator) {
```
usnavi: I think you are correct with `DiscoveryWriter`--it is a bug.  I think it will probably be easier to override `ABasicMetricsRW.isLocatorCurrent()` and `APreaggregatedMetricsRW.isLocatorCurrent()` to use the cache in `AstyanaxWriter`.  

When we finally move to DataStax, we can delete all the Astyanax stuff.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/cache/LocatorCache.java
```diff
@@ -0,0 +1,107 @@
+package com.rackspacecloud.blueflood.cache;
```
shintasmith: No LocatorCacheTest?

ChandraAddala: @shintasmith Added a test.

Issue #748, opened 2016-10-26
Blueflood rollups issue


ViswanathSRIRAM:
Hi Team,

We observed that we are missing roll up data in our test environment. Further digging led to below conclusion. Want to share this to validate my findings. 

Few things about our environment: We ingest data every 10 sec and the default TTL is set to 2 days.

The metrics were missing in metrics_locator so the corresponding rollups were not  happening. 

My observation of this behavior: 

_**AstyanaxWriter.java:**_
    _private static final Cache<String, Boolean> insertedLocators = CacheBuilder.newBuilder().expireAfterAccess(10,
            TimeUnit.MINUTES).concurrencyLevel(16).build();_

According to the above code, the cache built by blueflood ingest  is expiring in 10 minutes from the time of access. Since we write the metric for every 10 secs the cache renews every 10secs for those metrics and never expires. 

Metrics_locator CF has TTL of 7 days set due to the following code when writing to this CF.
_public interface TenantTtlProvider {
    public static final int LOCATOR_TTL = 604800;  
..
}_

For rollup metrics_locator will play a keyrole.
After 7 days from the time of inserting the metric to metrics_locator the locator data expires but the cache never expires(since we are accessing the metrics for 10 secs).
So the subsequent rollups were missing from the 8th day.

Until we further restart the application the data wouldn't come in to locator again in our case.

We fixed this by changing the **expireAfterAccess** to **expireAfterWrite** with TTL of 2 days and monitored for few days. We see this is working for our case now.

We are validating now the same with the below files.
_blueflood-core/src/main/java/com/rackspacecloud/blueflood/utils/Util.java 
blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/AbstractMetricsRW.java_ 

Please share your feedback if you see any concerns in the above approach.

ChandraAddala:
@ViswanathSRIRAM  Thx for creating this issue. We will take a look and get back to you.

ViswanathSRIRAM:
Sure Chandra.. Let me know if you have any questions in understanding.. I can comment or we can have a separate discussion.

ChandraAddala:
@ViswanathSRIRAM Thanks for the detailed write up. I was able to look into the issue. I submitted the below PR to take care of the issue you had identified.

https://github.com/rackerlabs/blueflood/pull/750

[...]

PR #750, opened 2016-11-01
Expiring locator cache after 3 days of creation of locator

ChandraAddala:
This is related to the issue that was raised in github regarding locator cache.

https://github.com/rackerlabs/blueflood/issues/748


usnavi:
Is there a JIRA for this?  If not, would you please create one, add the label "unplanned" and it to the next sprint?

shintasmith:
lgtm

PR #751, opened 2016-11-08
add response body for partial 207 errors for aggregated payload

VinnyQ:
# What

Add response body for aggregated/multi payload ingestion with partial invalid metrics (207 response code).
Fix missing metricName for aggregated payload validation response.
Fix and refactor integration tests.

# Why

Currently partial failure during ingestion for aggregated metrics return a 207 response code with an empty body.  A response body is returned to indicate which metrics failed and the reason for the failure.

# How

Use the new sendErrorResponse() method in the HttpAggregatedMultiIngestionHandler.java class.

### DIFF
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/formats/AggregatedPayload.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/formats/AggregatedPayload.java
@@ -113,12 +113,8 @@ public class AggregatedPayload {
                 metricName = ((BluefloodSet) leafBean).getName();
             }
 
-            if ((metricName.isEmpty()) && (payload.getAllMetricNames().size() > 0)) {
-                metricName = payload.getAllMetricNames().get(0);
-            }
-
             payload.validationErrors.add(new ErrorResponse.ErrorData(payload.getTenantId(), metricName,
-                    source, constraintViolation.getMessage()));
+                    source, constraintViolation.getMessage(), payload.getTimestamp()));
         }
     }
 
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/formats/AggregatedPayload.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/formats/AggregatedPayload.java
@@ -113,6 +113,10 @@ public class AggregatedPayload {
                 metricName = ((BluefloodSet) leafBean).getName();
             }
 
+            if ((metricName.isEmpty()) && (payload.getAllMetricNames().size() > 0)) {
+                metricName = payload.getAllMetricNames().get(0);
+            }
+
             payload.validationErrors.add(new ErrorResponse.ErrorData(payload.getTenantId(), metricName,
                     source, constraintViolation.getMessage()));
         }
@@ -158,7 +162,7 @@ public class AggregatedPayload {
     /**
      * This method is invoked by the validator automatically
      */
-    @AssertTrue(message="Atleast one of the aggregated metrics(gauges, counters, timers, sets) are expected")
+    @AssertTrue(message="At least one of the aggregated metrics(gauges, counters, timers, sets) are expected")
     private boolean isValid() {
         boolean isGaugePresent = gauges != null && gauges.length > 0;
         boolean isCounterPresent = counters != null && counters.length > 0;
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/outputs/formats/ErrorResponse.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/outputs/formats/ErrorResponse.java
@@ -31,6 +31,7 @@ public class ErrorResponse {
     public static class ErrorData {
 
         private String tenantId;
+        private Long timestamp;
         private String metricName;
         private String source;
         private String message;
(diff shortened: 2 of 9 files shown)


VinnyQ:
@ChandraAddala @usnavi @shintasmith please review, thanks!

blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/formats/AggregatedPayload.java
```diff
@@ -113,6 +113,10 @@ private static void validate(AggregatedPayload payload) {
                 metricName = ((BluefloodSet) leafBean).getName();
             }
 
+            if ((metricName.isEmpty()) && (payload.getAllMetricNames().size() > 0)) {
+                metricName = payload.getAllMetricNames().get(0);
+            }
```
usnavi: does the first metricName in the list always relate to the current violation?  If there isn't a metricName, does that mean that it couldn't be processed, and should be empty?

blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/inputs/handlers/HttpHandlerIntegrationTest.java
```diff
@@ -139,7 +139,11 @@ public void testHttpIngestionPartialInvalidCollectionTime() throws Exception {
 
         try {
             assertEquals("Should get status 207 from " + String.format(postMultiPath, TENANT_ID), 207, response.getStatusLine().getStatusCode() );
-            assertTrue("", errorResponse.getErrors().size() > 0);
+            for (int i = 0; i < 4; i++) {
+                assertEquals("Invalid error source", "collectionTime", errorResponse.getErrors().get(i).getSource());
+                assertEquals("Invalid error message", "Out of bounds. Cannot be more than " + BEFORE_CURRENT_COLLECTIONTIME_MS + " milliseconds into the past." +
+                        " Cannot be more than " + AFTER_CURRENT_COLLECTIONTIME_MS + " milliseconds into the future", errorResponse.getErrors().get(i).getMessage());
+            }
```
usnavi: 👍

blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/formats/AggregatedPayload.java
VinnyQ: Yes, I tested this.  Each "single" aggregatedPayload will have the metricName of the offending metric as the first element.  And actually, it'll always be the case that each instance of AggregatedPayload will only contain one leaf "metric".  Here's an example of the part of a payload this class represent:

```
{
      "tenantId":"333333",
      "timestamp":1478711535000,
      "flushInterval": 15000,
      "counters":[
          {
              "name":"agg.count.ten",
              "value":1,
              "rate":1
          }
      ]
    }
```

Even if the metric above is part of a multi-metrics payload like this:
```
curl -i -H "content-type: application/json" -X POST 'http://localhost:2440/v2.0/836986/ingest/aggregated/multi' -d \
  '[
    {
      "tenantId":"333333",
      "timestamp":1478711535000,
      "flushInterval": 15000,
      "counters":[
          {
              "name":"agg.count.nine",
              "value":1,
              "rate":1
          }
      ]
    },
    {
      "tenantId":"333333",
      "timestamp":1478711535000,
      "flushInterval": 15000,
      "counters":[
          {
              "name":"agg.count.ten",
              "value":2,
              "rate":1
          }
      ]
    },
    {
      "tenantId":"333333",
      "timestamp":1478711535000,
      "flushInterval": 15000,
      "counters":[
          {
              "name":"agg.count.eleven",
              "value":1,
              "rate":1
          },
          {
              "name":"agg.count.twelve",
              "value":1,
              "rate":1
          }
      ]
    }
  ]'
```

This is because if you look in the HttpAggregatedMultiIngestionHandler class, it first parse the entire json payload to each individual element (stored in bundleList: 
https://github.com/rackerlabs/blueflood/blob/master/blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpAggregatedMultiIngestionHandler.java#L76
and
https://github.com/rackerlabs/blueflood/blob/master/blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpAggregatedMultiIngestionHandler.java#L176)

Then for each of the element in bundleList (an AggregatedPayload), it does the validation and create the error payload if there's violation:  https://github.com/rackerlabs/blueflood/blob/master/blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpAggregatedMultiIngestionHandler.java#L85


In the majority of the cases, the metricName is pulled from the previous block in the AggregatedPayload.validate() here:  https://github.com/rackerlabs/blueflood/blob/master/blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/formats/AggregatedPayload.java#L105-L114 

with the assumption that leafBean was created correctly from the json parser.  

However, there were cases where the parser itself didn't create the AggregatedPayload element correctly (e.g. a missing "value" field that is required), resulting in the type of the element not being set and an metricName.  In this case, the only way to retrieve the metricName is through the single json element payload itself. 

Hope that make sense.  But do know that I tried testing different cases including one with multiple aggregated payloads, with the first few being valid and the last one being invalid, and the error response has the correct metricName for the one failing the validation (and not ingested).  You'll see a few of this cases in the system integration tests itself.

VinnyQ:
nevermind you're right, good catch @usnavi , we really shouldn't return any metricNames for an entire payload if the collectionTime is invalid, since it's invalid for all of them, just not the first one.  

I added the timestamp to the error response, which is more helpful for the user to pinpoint which metrics failed validation and ingestion.

@usnavi @shintasmith @ChandraAddala @izrik please review latest when you have a chance.

usnavi:
Thanks @VinnyQ!

blueflood-core/src/main/java/com/rackspacecloud/blueflood/outputs/formats/ErrorResponse.java
```diff
@@ -31,24 +31,30 @@ public String toString() {
...
+        private Long timestamp;
         private String metricName;
         private String source;
         private String message;
 
         public ErrorData() {
         }
 
-        public ErrorData(String tenantId, String metricName, String source, String message) {
+        public ErrorData(String tenantId, String metricName, String source, String message, Long timestamp) {
```
shintasmith: Maybe more comment on what this timestamp represent? Is this ingest time? Collection time?

VinnyQ: yes, that.  Where should I put the comment?

PR #755, opened 2016-11-22
change response to BAD_REQUEST for empty aggregated metrics request

VinnyQ:
# WHAT

change response to BAD_REQUEST for empty array of aggregated metrics request

# WHY

empty array should return BAD_REQUEST (400) instead of OK (200).

# HOW

Change response in class HttpMetricsIngestionHandler.

# TEST

Run cloudmetrics-system-tests with empty array test unignored against system with this change.


VinnyQ:
@usnavi Done.  Actually, there were already a test for this, I just forgot to change the status and response message, hence the previous PR build failure.  Fixed now and PR build check should all pass.

PR #760, opened 2016-11-30
Delayed metrics reroll config

ChandraAddala:
Introducing a new configuration, DELAYED_METRICS_REROLL_GRANULARITY, using which we control the granularity upto which we re-roll only the delayed metrics instead of entire slot.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## IN PROGRESS
 * Upgraded datastax driver to 3.1.2
+* Added new configuration option DELAYED_METRICS_REROLL_GRANULARITY
 
 ## blueflood-2.0.0
 * Switched to officially building with JDK 8. **NOTE**: after this release, JDK 7 will no longer be supported.
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/rollup/SlotKey.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/rollup/SlotKey.java
@@ -149,7 +149,7 @@ public final class SlotKey {
     public Collection<SlotKey> getChildrenKeys(Granularity destGranularity) {
 
         if (!getGranularity().isCoarser(destGranularity)) {
-            throw new IllegalArgumentException("Current granularity must be coarser than the destination granularity");
+            throw new IllegalArgumentException(String.format("Current granularity [%s] must be coarser than the destination granularity [%s]", getGranularity(), destGranularity));
         }
 
         List<SlotKey> result = new ArrayList<SlotKey>();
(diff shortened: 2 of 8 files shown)


blueflood-core/src/main/java/com/rackspacecloud/blueflood/rollup/SlotKey.java
```diff
@@ -135,6 +136,33 @@ public static SlotKey parse(String string) {
     }
 
     /**
+     * Same as the method {@link #getChildrenKeys()} except this method returns only
+     * the children corresponding to the destination granularity.
+     *
+     * For example, a slot key of granularity
+     * {@link Granularity#MIN_1440 MIN_1440}, for destination granularity of
+     * {@link Granularity#MIN_60 MIN_60}, will have 6*4=24 children.
```
shintasmith: Curious, Is this a normal example? I mean, do we normally ask from 1 day (1440m) granularity to 1 hr (60m) granularity, skipping the 240m granularity? 

Also, the term "children" is not as clear, but I realized it's probably been used in other places. At first I thought 1440m is a child of the 240m. I don't know if it's possible to change to "Finer" and "Coarser"

blueflood-core/src/main/java/com/rackspacecloud/blueflood/rollup/SlotKey.java
```diff
@@ -135,6 +136,33 @@ public static SlotKey parse(String string) {
...
+     *
+     * @param destGranularity
+     * @return
+     */
+    public Collection<SlotKey> getChildrenKeys(Granularity destGranularity) {
+
+        if (!getGranularity().isCoarser(destGranularity)) {
+            throw new IllegalArgumentException("Current granularity must be coarser than the destination granularity");
```
shintasmith: Can you print the 2 granularities involved in the Exception msg?

CHANGES.md
```diff
@@ -2,6 +2,7 @@
 
 ## IN PROGRESS
 * Upgraded datastax driver to 3.1.2
+* Added new configuration option DELAYED_METRICS_REROLL_GRANULARITY
```
ChandraAddala: Once everything is merged to master, I plan on writing a wiki page just for delayed metrics explaining various configs and how to tweak them.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/rollup/SlotKey.java
ChandraAddala: There is getChildrenKeys above which would list children of all sublevels. This method uses that and only returns the level we are interested in. I just followed the terminology used there.

PR #766, opened 2016-12-15
Integrated docker maven build

SlevinBE:
The current Docker build has several problems:
- it depends on a blueflood release jar hosted on github to make a docker image. This makes it impossible to checkout any blueflood branch/tag and create a docker image out of it.
- it downloads the latest version of the Blueflood code from github. Because of this you can't have a reliable reproducible build.
- version numbers/urls need to be updated for every release in the Dockerfile and docker-entrypoint.sh, which is prone to mistakes.

This PR solves these issues by integrating the docker build into the Maven build. Basically it allows you to build a docker image of any branch/tag by executing the following command:
```
mvn clean package  docker:build -Pall-modules
```

This results in a docker image with two tags: 'latest' and the version of the maven project

The Dockerfile copies the jar which is the result of the package command into the Docker image, together with the other required files (such as ES-Setup, load.cdl). These changes make it independent from any outside resources such as github.

### DIFF
--- a/contrib/blueflood-docker/Dockerfile
+++ b/contrib/blueflood-docker/Dockerfile
@@ -13,8 +13,7 @@ COPY ES-Setup /ES-Setup
 COPY load.cdl /blueflood.cdl
 
 COPY artifacts /
-RUN (ls blueflood-all-*.jar | sed "s|blueflood-all-\(.*\)-jar-with-dependencies.jar|\1|") >> BLUEFLOOD_VERSION
-RUN mv blueflood-all-*-jar-with-dependencies.jar blueflood-all-jar-with-dependencies.jar
+RUN ln -s blueflood-all-*-jar-with-dependencies.jar blueflood-all-jar-with-dependencies.jar
 
 ENV MAX_ROLLUP_THREADS=20
 ENV MAX_TIMEOUT_WHEN_EXHAUSTED=2000
--- a/contrib/blueflood-docker/Dockerfile
+++ b/contrib/blueflood-docker/Dockerfile
@@ -9,12 +9,12 @@ RUN apt-get install -y python python-dev python-pip python-virtualenv && \
     rm -rf /var/lib/apt/lists/*
 RUN pip install cqlsh==4.1.1
 
-RUN git clone https://github.com/rackerlabs/blueflood.git
-RUN mv ./blueflood/blueflood-elasticsearch/src/main/resources ./ES-Setup
-RUN mv ./blueflood/src/cassandra/cli/load.cdl ./blueflood.cdl
-RUN rm -r ./blueflood
+COPY ES-Setup /ES-Setup
+COPY load.cdl /blueflood.cdl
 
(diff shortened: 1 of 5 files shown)


shintasmith:
@SlevinBE thank you for your contribution. I'll look through your PR shortly.

contrib/blueflood-docker/Dockerfile
```diff
@@ -9,12 +9,12 @@ RUN apt-get install -y python python-dev python-pip python-virtualenv && \
...
-RUN mv ./blueflood/blueflood-elasticsearch/src/main/resources ./ES-Setup
-RUN mv ./blueflood/src/cassandra/cli/load.cdl ./blueflood.cdl
-RUN rm -r ./blueflood
+COPY ES-Setup /ES-Setup
+COPY load.cdl /blueflood.cdl
 
-RUN curl -s -L https://github.com/rackerlabs/blueflood/releases/latest | egrep -o 'rackerlabs/blueflood/releases/download/rax-release-.*/blueflood-all-.*-jar-with-dependencies.jar' |  xargs -I % curl -C - -L https://github.com/% --create-dirs -o ./blueflood-all-2.0.0-SNAPSHOT-jar-with-dependencies.jar
+COPY artifacts /
+RUN (ls blueflood-all-*.jar | sed "s|blueflood-all-\(.*\)-jar-with-dependencies.jar|\1|") >> BLUEFLOOD_VERSION
+RUN mv blueflood-all-*-jar-with-dependencies.jar blueflood-all-jar-with-dependencies.jar
```
shintasmith: Instead of ```mv```, can you create a soft link? I.e: ```RUN ln -s blueflood-all-$BLUEFLOOD_VERSION-jar-with-dependencies.jar blueflood-all-jar-with-dependencies.jar``` ? 

It's just a nice to have. I know you already tag the image with a version. With a soft link, when you're inside the container, you know which version of blueflood you are running.

SlevinBE: oh, but that's why I'm extracting the version from the file and putting it into a BLUEFLOOD_VERSION file at line 16 ;)

SlevinBE:
I've implemented the suggestion to use a soft link to the jar file instead of writing the version to a separate file.

shintasmith:
@SlevinBE , we are always delightfully surprised when we get PRs like yours. Would you mind sharing with us what you are using/planning to use blueflood for?

Thanks!
-shinta

SlevinBE:
Klarrio is specialised in building highly scalable IoT platforms for customers by either integrating open source components or by building custom components where the open source solutions don't exist or are not sufficient.
We are evaluating Blueflood for storing IoT timeseries data, backed by Scylladb (an API-compatible Cassandra alternative) and running in a DC/OS clustered cloud environment with Docker containers being our primary deployment artifacts.

usnavi:
Thanks for the info, @SlevinBE!  We recently learned about Scylladb and am curious how it compares.  They boast quite an performance boost.

SlevinBE:
We're still in the evaluation phase with Scylladb, so can't share much yet about the performance boosts (but that's also the reason why we're trying it out). 
However, we did find this series of blog posts interesting about a company called "Outbrain" which did an evaluation over multiple months and kept a logbook. http://techblog.outbrain.com/2016/03/we-are-testing-scylladb-live-blogging-1/

PR #771, opened 2017-01-02
allowing to configure the Cassandra datacenter name

SlevinBE:
Currently the Cassandra datacenter name is hardcoded with no option to change this. This PR changes this by adding an overridable "CASSANDRA_LOCAL_DATACENTER_NAME" variable.


goru97:
:+1:

PR #772, opened 2017-01-04
Netty datastax issue

ChandraAddala:
We are having an issue when using datastax driver, where the connections remain in ESTABLISHED state. The number of connections keep on increasing until the client ultimately receives connection timeouts. 

This PR adds a handler which closes the connection after certain idle time of not having any inbound traffic. This idle time out is configurable using HTTP_CONNECTION_READ_IDLE_TIME_SECONDS.


shintasmith:
@ChandraAddala , strangely, DateTimeParserTest failed.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionServer.java
```diff
@@ -133,8 +133,11 @@ public void initChannel(SocketChannel channel) throws Exception {
     private void setupPipeline(SocketChannel channel, RouteMatcher router) {
         final ChannelPipeline pipeline = channel.pipeline();
 
-        pipeline.addLast("encoder", new HttpResponseEncoder());
-        pipeline.addLast("decoder", new HttpRequestDecoder() {
+        pipeline.addLast("logging", new LoggingHandler(LogLevel.TRACE)); //duplex handler
+        pipeline.addLast("idleStateHandler", new IdleStateHandler(HTTP_CONNECTION_READ_IDLE_TIME_SECONDS, 0, 0)); //duplex handler
```
shintasmith: 👍 nice!

shintasmith: What's a duplex handler?

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/UserDefinedEventHandler.java
```diff
@@ -0,0 +1,39 @@
...
+import io.netty.handler.timeout.IdleStateHandler;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A generic handler to handle user triggered events. Currently this handler only
+ * handles {@link IdleStateEvent} triggered by {@link IdleStateHandler}
+ *
+ */
+public class UserDefinedEventHandler extends ChannelDuplexHandler {
```
shintasmith: How about just naming the Handler specificly for what it does? Like maybe CloseOnReadIdleHandler ?

ChandraAddala: I renamed this file quite a bit now.  I am open to suggestions. But before doing that, technically speaking, the handler is quite generic. It gets called for any `userEventTriggered`. Once its triggered, we have logic to only handle `IdleStateEvent` and ignore the rest.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionServer.java
ChandraAddala: its both inbound and outbound..

PR #774, opened 2017-01-05
removing enum-related tables

usnavi:
Removing enum-related tables from this project (cassandra and elasticsearch) as enums are no longer supported.


usnavi:
@drenalin23 @izrik @shintasmith @ChandraAddala - review pls.

PR #782, opened 2017-02-07
remove log.error() on cases where it's not really system error

shintasmith:
remove calls to log.error() as an invalid user input should not be considered an error

### DIFF
--- a/blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpAggregatedMultiIngestionHandler.java
+++ b/blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpAggregatedMultiIngestionHandler.java
@@ -139,16 +139,14 @@ public class HttpAggregatedMultiIngestionHandler implements HttpRequestHandler {
             }
         } catch (JsonParseException ex) {
             log.debug(String.format("BAD JSON: %s", body));
-            log.error(ex.getMessage(), ex);
             DefaultHandler.sendErrorResponse(ctx, request, ex.getMessage(), HttpResponseStatus.BAD_REQUEST);
         } catch (InvalidDataException ex) {
             log.debug(String.format("Invalid request body: %s", body));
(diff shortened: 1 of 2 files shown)


shintasmith:
@usnavi @ChandraAddala @izrik pls review

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpAggregatedMultiIngestionHandler.java
```diff
@@ -139,16 +139,14 @@ public void handle(ChannelHandlerContext ctx, FullHttpRequest request) {
...
             DefaultHandler.sendErrorResponse(ctx, request, ex.getMessage(), HttpResponseStatus.BAD_REQUEST);
         } catch (InvalidDataException ex) {
             log.debug(String.format("Invalid request body: %s", body));
-            log.error(ex.getMessage(), ex);
             DefaultHandler.sendErrorResponse(ctx, request, ex.getMessage(), HttpResponseStatus.BAD_REQUEST);
         } catch (TimeoutException ex) {
             DefaultHandler.sendErrorResponse(ctx, request, "Timed out persisting metrics", HttpResponseStatus.ACCEPTED);
         } catch (Exception ex) {
-            log.debug(String.format("BAD JSON: %s", body));
+            log.debug(String.format("Exception processing: %s", body));
```
ChandraAddala: If we get some exception which we are not expecting, I think we need some visibility to it. Should we log error here?

shintasmith: There is already a log.error(). See the line below.

ChandraAddala: I must be blind.

PR #783, opened 2017-02-10
splitting up the locator cache into a discovery and a batch locator cache

SlevinBE:
**Issue**
I've noticed an issue when ingesting metrics with new names into blueflood. The metrics always end up in Cassandra, but sometimes the metric names are not indexed in Elasticsearch.

**Investigation**
I've investigated the issue and it appears to be a problem with the LocatorCache and the fact that discovery and batch items are written in parallel threads. 

In the [HttpMetricsIngestionServer](https://github.com/rackerlabs/blueflood/blob/blueflood-2.0.0/blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionServer.java#L238-L239) the discoveryWriter and batchWriter are applied right after each other on the metrics. When looking at such a DiscoveryWriter or BatchWriter, I noticed that they [submit their work in a ThreadPool](https://github.com/rackerlabs/blueflood/blob/blueflood-2.0.0/blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/processors/DiscoveryWriter.java#L104). So while it looks like one runs after the other, in fact they are running in parallel.

Locators are only [marked current](https://github.com/rackerlabs/blueflood/blob/blueflood-2.0.0/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/astyanax/AstyanaxWriter.java#L123) by one of the batch writers (AstyanaxWriter or DatastaxWriter), not by the discovery writer. However, both the DiscoveryWriter and the BatchWriter check whether a locator is current before writing it to their backend (Cassandra for the BatchWriter and ES for the DiscoveryWriter). 

I think a false assumption was made here (or maybe the code evolved over time to threads) that the DiscoveryWriter runs before the BatchWriter, and therefore having the BatchWriter marking a locator as current is ok since the locator has already been written to ES. However, since this is not the case (they run in parallel), it's perfectly possible that:
1. the batch writer's thread runs slightly before the discovery writer's thread and writes the metric to Cassandra and marks the locator as current
2. then the discovery writer (writing the same metrics in parallel to ES) checks whether the locator is current, which it is since the batch writer just marked it as such.
3. because the locator is current, the discovery writer skips it, causing the issue of missing metrics in ES.

**Solution**
This PR tries to solve this issue by giving each writer its own LocatorCache. So I've created a static DiscoveryLocatorCache and a BatchLocatorCache, so that each writer has its own cache.
I've chosen this approach because it makes it easier to reason about the cache since it doesn't affect two different storage backends. That way you don't have to synchronise writes to ES and Cassandra, they can each handle the writes in their own thread and just update their own cache.

The DiscoveryWriter is now marking its own locators as current after having written them to ES, instead of relying on the batch writer.

This PR has been tested on our cluster and also includes a unit test for the DiscoveryWriter.


SlevinBE:
mmm, the failed integration test (HttpMultiRollupsQueryHandlerIntegrationTest) ran fine locally multiple times :(

ChandraAddala:
The cache usually gets pretty big. By maintaining two(very similar) of these caches, we could be putting extra pressure on the JVM heap. I am wondering if there is a better way to handle this. May be decide whether the locator is current before sending it to either of these writers? Not sure if that would work. The solution you provided could be the last resort but I would like to try and see if there are other solutions.

ChandraAddala:
@SlevinBE How about we do these changes in `HttpMetricsIngestionServer`? The below code does not compile but just to give you an idea. We basically remove any locators that are in cache before sending it to `discoveryWriter`. Once we make a call to `discoveryWriter`, it wont verify the cache anymore. This way, even when `batchWriter` runs slightly before `discoveryWriter` and marks the locator as current, we wont be affected.  What do you think?

```
            final List<IMetric> willIndex = DiscoveryWriter.condense(batches);

            discoveryWriter.apply(willIndex);
            return batchWriter.apply(batches);
```

SlevinBE:
While this would indeed solve the problem in the happy path, this however assumes that both the batchWriter and discoveryWriter are able to write the metrics successfully. In case the batchWriter does write the metrics successfully to Cassandra (and marks the locator as current) but the discoveryWriter encounters an error while writing to ES, then this would mean that all metrics with the same locator coming after this batch won't be provided to the discoveryWriter since they are marked as current.

SlevinBE:
Another option would be to keep a pair of Boolean values per locator entry in the LocatorCache, one indicating whether the discovery entry is current, the other indicating whether the batch entry is current.
Based on a simple test this would only increase the memory footprint for 1 million metrics by about 24MB (the current LocatorCache implementation uses about 240MB to hold 1 million metrics)

```
public class LocatorCache {

    private final Cache<String, LocatorCacheEntry> insertedLocators;

    public synchronized void setDiscoveryLocatorCurrent(Locator loc) {}
    public synchronized void setBatchLocatorCurrent(Locator loc) {}

    private class LocatorCacheEntry {
        private Boolean discoveryCurrent = false;
        private Boolean batchCurrent = false;

        ...
    }

}
```

ChandraAddala:
SlevinBE Using the existing cache to keep a pair of boolean values sounds good to me.

SlevinBE:
@ChandraAddala I've implemented the changes to the LocatorCache to keep a composite value of two booleans in the cache.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/cache/LocatorCache.java
```diff
@@ -128,4 +159,33 @@ public synchronized void resetCache() {
...
+    @VisibleForTesting
+    public synchronized void resetInsertedLocatorsCache() {
+        insertedLocators.invalidateAll();
+    }
+
+    /**
+     * Cache entry which defines where the locator has been inserted into during the caching period.
+     */
+    private class LocatorCacheEntry {
+        private Boolean discoveryCurrent = false;
```
ChandraAddala: Can we use primitive boolean?

SlevinBE:
@ChandraAddala I've applied the requested changes.

PR #784, opened 2017-02-17
add logic to send back Keep-Alive header if httpConnectionIdleTimeout is non zero

shintasmith:
Blueflood introduced a configuration setting HTTP_CONNECTION_READ_IDLE_TIME_SECONDS that controls how long Blueflood waits before closing idle connections. Setting this to zero effectively disables it.

When this property is set to non zero value x, however, Blueflood sends back ```Connection: keep-alive``` header if the request has that same header sent. But we close the connections after x seconds.

This PR changes that behavior. If the property is set to non zero value x, the new behavior is now to send back additional ```Keep-Alive: timeout=x``` header.

Other changes are
* refactored HttpResponder so I can add unit tests
* added unit tests
* changed the default value of HTTP_CONNECTION_READ_IDLE_TIME_SECONDS to 0 (essentially disable it)

Reference:
https://www.w3.org/Protocols/HTTP/1.1/draft-ietf-http-v11-spec-01#Connection

### DIFF
--- a/blueflood-http/src/main/java/com/rackspacecloud/blueflood/http/DefaultHandler.java
+++ b/blueflood-http/src/main/java/com/rackspacecloud/blueflood/http/DefaultHandler.java
@@ -43,9 +43,11 @@ public class DefaultHandler implements HttpRequestHandler {
 
     private static final Logger log = LoggerFactory.getLogger(DefaultHandler.class);
 
+    private static final HttpResponder responder = HttpResponder.getInstance();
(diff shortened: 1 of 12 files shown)


shintasmith:
@usnavi @ChandraAddala @izrik pls review

shintasmith:
@ChandraAddala , FYI, I forgot to add the new unit test file. I just pushed HttpResponderTest.java

PR #785, opened 2017-02-20
Refactoring metric_name/search api implementation

ChandraAddala:
This refactoring is to refactor the design (and complexity) that creeped into implementing this API because of enums. I had to address these before I could integrate new implementation using the new index. In this process, I did the following change

1) Renamed `MetricToken` to `MetricName`. We were using `MetricToken` as domain object to support `/metric_name/search`.  
2) Changed the regex to only grab next level of tokens, instead of the current regex which was written to handle enums as well.
3) Changed code to remove support of enums completely. Even though some sections of the code was removed earlier, the code was still designed to handle enums. So I cleaned up some code.

In summary, a metric foo.bar.baz.qux has 4 `MetricName` objects and this is what the metric_name/search API returns, the partial/complete metric names.
```
MetricName(name=foo, isCompleteName=false)
MetricName(name=foo.bar, isCompleteName=false)
MetricName(name=foo.bar.baz, isCompleteName=false)
MetricName(name=foo.bar.baz.qux, isCompleteName=true)
```


blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/AbstractElasticIO.java
```diff
@@ -97,50 +100,51 @@
...
         try {
-            response = getMetricTokensFromES(tenant, regexForPrevToNextLevel(query));
+            response = getMetricNamesFromES(tenant, regexToGrabCurrentAndNextLevel(query));
         } finally {
-            esMetricTokensQueryTimerCtx.stop();
+            esMetricNamesQueryTimerCtx.stop();
         }
 
-        int totalTokens = getTotalTokens(query);
+        // if query = foo.bar.*, base level is 3 which is equal to the number of tokens in the query.
```
usnavi: Maybe precede this comment with "In this example..".  I was a bit confused where 3 came from for a bit.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/AbstractElasticIO.java
```diff
@@ -26,7 +29,7 @@
 
     // todo: these should be instances per client.
     protected final Timer searchTimer = Metrics.timer(getClass(), "Search Duration");
-    protected final Timer esMetricTokensQueryTimer = Metrics.timer(getClass(), "ES Metric Tokens Query Duration");
+    protected final Timer esMetricNamesQueryTimer = Metrics.timer(getClass(), "ES Metric Names Query Duration");
```
izrik: Just a thought: will changing this name affect any of our monitoring systems, e.g. Riemann?

ChandraAddala: Dont think Riemann will be affected, but I have to fix the graphs if we are plotting this.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/DiscoveryIO.java
```diff
@@ -9,6 +9,6 @@
     public void insertDiscovery(List<IMetric> metrics) throws Exception;
     public List<SearchResult> search(String tenant, String query) throws Exception;
     public List<SearchResult> search(String tenant, List<String> queries) throws Exception;
-    public List<MetricToken> getMetricTokens(String tenant, String prefix) throws Exception;
+    public List<MetricName> getMetricNames(String tenant, String prefix) throws Exception;
```
shintasmith: 👍 I always found the getMetricTokens() method name confusing. Thanks for changing!

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/AbstractElasticIO.java
```diff
@@ -97,50 +100,51 @@
...
+        List<MetricName> metricNames = new ArrayList<>();
 
-        MetricTokenListBuilder tokenInfoBuilder = new MetricTokenListBuilder();
-        //token paths matching query, which also have a next level.
-        tokenInfoBuilder.addTokenPathWithNextLevel(metricIndexData.getTokenPathsWithNextLevel());
+        //Metric Names matching query which have next level
+        metricNames.addAll(metricIndexData.getMetricNamesWithNextLevel()
+                                          .stream()
+                                          .map(x -> new MetricName(x, false))
+                                          .collect(toSet()));
```
shintasmith: Doing some Java8!!! 👍

PR #787, opened 2017-02-24
Metric Name search

ChandraAddala:
This PR is to revamp the `/metric_name/search` using new ES index inorder to improve performance. Some major changes that are part of this PR are as follows

1) Introducing a new index `metric_tokens` to store tokens based on the parent. For a metric foo.bar.baz, associating token 'baz' with parent 'foo.bar'
2) Indexing and querying using the new index, is added as an experimental feature and is controlled by the flag `EXP_TOKEN_SEARCH_IMPROVEMENTS`. Only if this flag is set to `true` the new index will come into play. If someone using blueflood does not need grafana support, they can opt out of indexing to this new index using this flag.
2) `TokenDiscoveryWriter` with separate threadpool to be used for indexing tokens.
3) `TokenDiscoveryIO ` to handle tokens, similar to `DisvoeryIO` which handles locators.
4) Having a separate indicator in `LocatorCache` to mark a locator as indexed using new index.
5) Introducing `TokenCache` to cache tokens.


blueflood-core/src/main/java/com/rackspacecloud/blueflood/types/Token.java
```diff
@@ -0,0 +1,147 @@
...
+    private final String id;
+    private final Locator locator;
+
+    public Token(Locator locator, String[] tokens, int level) {
+
+        if (level > tokens.length - 1)
+            throw new IllegalArgumentException("Invalid level for the given tokens");
+
+        this.locator = locator;
+        this.token = tokens[level];
```
shintasmith: Do we want to add more checks on the size of the array itself?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/types/Token.java
```diff
@@ -0,0 +1,147 @@
...
+ *      parent = 'a.b.c'
+ *      isLeaf = true
+ *      id = '111111:a.b.c.d:$'
+ *      tenantId = '111111'
+ *
+ */
+public class Token {
+
+    public static final String LEAF_NODE_SUFFIX = ":$";
+    public static final String SEPARATOR = ":";
```
shintasmith: Should we worry about metric names containing ```:``` ?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/cache/LocatorCache.java
```diff
@@ -103,6 +103,17 @@ public synchronized boolean isLocatorCurrentInDiscoveryLayer(Locator loc) {
     }
 
     /**
+     * Checks if Locator is recently inserted in the token discovery layer
+     *
+     * @param loc
+     * @return
+     */
+    public synchronized boolean isLocatorCurrentInTokenDiscoveryLayer(Locator loc) {
```
ChandraAddala: I will add some documentation at the class level `LocatorCache`

blueflood-core/src/test/java/com/rackspacecloud/blueflood/inputs/processors/TokenDiscoveryWriterTest.java
```diff
@@ -0,0 +1,174 @@
...
+        batch.add(Arrays.asList(new Metric(locator2, 1, 1L, new TimeValue(5, TimeUnit.SECONDS), "")));
+
+        Set<Token> tokens = TokenDiscoveryWriter.generateAndConsolidateTokens(TokenDiscoveryWriter.getLocators(batch));
+
+        assertEquals("Invalid number of tokens", 5, tokens.size());
+
+    }
+
+    @Test
+    public void testProcessTokens() throws Exception {
```
shintasmith: More meaningful method name pls? What are we testing here? Either rename method to clarify or add comments.

blueflood-core/src/test/java/com/rackspacecloud/blueflood/inputs/processors/TokenDiscoveryWriterTest.java
```diff
@@ -0,0 +1,174 @@
...
+
+        verify(discovererB, times(1)).insertDiscovery(captor.capture());
+        List<Token> actualTokensB = captor.getValue();
+
+        assertEquals("Unexpected number of tokens", 4, actualTokensB.size());
+        assertTrue(actualTokensB.containsAll(expectedTokens) && expectedTokens.containsAll(actualTokensB));
+    }
+
+    @Test
+    public void testProcessTokensSameLocatorBeingIndexed() throws Exception {
```
shintasmith: Maybe name it testProcessTokensSameLocatorBeingIndexedOnce() ?

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/Discovery.java
```diff
@@ -24,6 +24,9 @@
 import java.util.Map;
 
 public class Discovery {
+
+    protected final static String SEPARATOR = ":";
```
shintasmith: Does this separator have to be the same as Token.SEPARATOR ? If yes, then do we need to just place it in one class?

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/ESFieldLabel.java
```diff
@@ -20,6 +20,9 @@
     metric_name,
     tenantId,
     unit,
-    enum_values
+    enum_values,
```
shintasmith: Should we get rid of enum_values ?

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/ElasticTokensIO.java
```diff
@@ -0,0 +1,291 @@
...
+            BulkRequestBuilder bulk = client.prepareBulk();
+
+            for (Token token : tokens) {
+                bulk.add(createSingleRequest(token));
+            }
+
+            bulk.execute().actionGet();
+        } catch (EsRejectedExecutionException esEx) {
+            log.error(("Error during bulk insert to ES with status: [" + esEx.status() + "] " +
+                    "with message: [" + esEx.getDetailedMessage() + "]"));
```
shintasmith: We don't want to bubble up the exception?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/cache/TokenCache.java
```diff
@@ -0,0 +1,75 @@
...
+import com.google.common.cache.CacheBuilder;
+import com.rackspacecloud.blueflood.types.Token;
+import com.rackspacecloud.blueflood.utils.Metrics;
+
+import java.util.concurrent.TimeUnit;
+
+public class TokenCache {
+
+    // this collection is used to reduce the number of tokens that get written.
+    // Simply, if a token has been seen within the last 10 minutes, don't bother.
```
ChandraAddala: If we index a locator `1111:a.b.c.d` for tokens. we index documents with id's `1111:a, 1111:a.b, 1111:a.b.c, 1111:a.b.c.d:$` as tokens. Next time we get a locator`1111:a.b.c.e`, we only want to index `1111:a.b.c.e:$` as we had already indexed previous layers. Does this answer your question?

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/ElasticTokensIO.java
```diff
@@ -0,0 +1,291 @@
...
+ * {@link TokenDiscoveryIO} implementation using elastic search
+ */
+public class ElasticTokensIO implements TokenDiscoveryIO {
+
+    public static final String ES_DOCUMENT_TYPE = "tokens";
+
+    protected final Histogram batchHistogram = Metrics.histogram(getClass(), "Batch Sizes");
+    protected final Timer writeTimer = Metrics.timer(getClass(), "Write Duration");
+
+    protected final Timer esMetricNamesQueryTimer = Metrics.timer(getClass(), "ES Metric Names Query Duration");
```
shintasmith: Can we keep the writeTimer and queryTimer names similar? I am fine with just "ES Write Duration" and "ES Query Duration".

blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/processors/TokenDiscoveryWriter.java
```diff
@@ -0,0 +1,133 @@
...
+import java.util.stream.Stream;
+
+import static java.util.stream.Collectors.toSet;
+
+/**
+ * A separate discovery writer for tokens. This class is modelled after {@link DiscoveryWriter}
+ */
+public class TokenDiscoveryWriter extends FunctionWithThreadPool<List<List<IMetric>>, Void> {
+
+    private final List<TokenDiscoveryIO> tokenDiscoveryIOs = new ArrayList<>();
```
ChandraAddala: This is the extensibility that blueflood provides which was existing and I merely supported it. We can provide any implementations for `TokenDiscoveryIO`. I only provided one implementation `ElasticTokensIO` which is powered by elastic search. But if you want to provided a different implementation say for graph database, all you have to do is provide another implementation for `TokenDiscoveryIO` using graphs and we are good.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/ElasticTokensIO.java
```diff
@@ -0,0 +1,291 @@
...
+     * @param tenantId
+     * @param query
+     * @return
+     */
+    private BoolQueryBuilder buildESQuery(String tenantId, String query) {
+
+        String[] queryTokens = query.split(Locator.metricTokenSeparatorRegex);
+        String lastToken = queryTokens[queryTokens.length - 1];
+
+        BoolQueryBuilder bqb = boolQuery();
```
shintasmith: For a split second, I read this to be "BobQueryBuilder" 😂

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/ElasticTokensIO.java
```diff
@@ -0,0 +1,291 @@
...
+        /**
+         * Builds parent part of the query for the given input query glob tokens.
+         * For a given query foo.bar.*, parent part is foo.bar
+         *
+         * For example:
+         *
+         *  For query = foo.bar.*
+         *          { "term": {  "parent": "foo.bar" }}
+         *
+         *  For query = foo.*.*
```
shintasmith: I didn't even know we support this. 😊

blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/processors/TokenDiscoveryWriter.java
```diff
@@ -0,0 +1,149 @@
...
+        tokenDiscoveryIOs.add(io);
+        writeErrorMeters.put(io.getClass(),
+                             Metrics.meter(io.getClass(), "TokenDiscoveryWriter Write Errors")
+        );
+    }
+
+    public void registerIOModules() {
+        List<String> modules = Configuration.getInstance().getListProperty(CoreConfig.TOKEN_DISCOVERY_MODULES);
+
+        ClassLoader classLoader = TokenDiscoveryIO.class.getClassLoader();
```
ChandraAddala: I just answered above on why we do this. I understand it is making it complex but looks like it is the way it is designed. Provides some form for plug-ability if someone decides to use something other than elastic search.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionServer.java
```diff
@@ -73,6 +70,9 @@
     private int HTTP_CONNECTION_READ_IDLE_TIME_SECONDS =
             Configuration.getInstance().getIntegerProperty(HttpConfig.HTTP_CONNECTION_READ_IDLE_TIME_SECONDS);
 
+    public static boolean EXP_TOKEN_SEARCH_IMPROVEMENTS =
```
shintasmith: How about if we drop the EXP part? Because when it becomes non-experimental, are we going to change the config name? Also add comment that this relates to the feature of writing hierarchical tokens to metrics_token

shintasmith:
Please also update README.md

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/DiscoveryIO.java
```diff
@@ -3,12 +3,11 @@
 import com.rackspacecloud.blueflood.types.IMetric;
 import java.util.List;
 
-public interface DiscoveryIO {
+public interface DiscoveryIO extends BaseDiscoveryIO {
```
ChandraAddala: I am keeping the old way of getting tokens, which is using 'aggregated ES queries'. If someone decides to not create  separate index for token search, they can still use the old way of doing thing using only one index.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/TokenDiscoveryIO.java
```diff
@@ -0,0 +1,12 @@
+package com.rackspacecloud.blueflood.io;
+
+
+import com.rackspacecloud.blueflood.types.Token;
+
+import java.io.IOException;
+import java.util.List;
+
+public interface TokenDiscoveryIO  extends BaseDiscoveryIO {
```
ChandraAddala: The only reason `BaseDiscoveryIO` is present, it because I have `getMetricNames` that is being supported in both `DiscoverIO` and `TokenDiscoveryIO`, one using aggregated ES queries and the other using separate ES index respectively. 

I was able to leverage this common interface while writing tests. I only write one test and use both implementations so that they both work the same.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/types/Locator.java
```diff
@@ -23,8 +23,8 @@
 import org.slf4j.LoggerFactory;
 
 public class Locator implements Comparable<Locator> {
-    private static final String metricTokenSeparator;
-    private static final String metricTokenSeparatorRegex;
+    public static final String metricTokenSeparator;
+    public static final String metricTokenSeparatorRegex;
```
ChandraAddala: They are existing variables, I guess I can change them

blueflood-core/src/main/java/com/rackspacecloud/blueflood/types/Token.java
ChandraAddala: We use this separator today to define a locator which is of the form <tenantid>:<metric_name>. This will be a problem for very specific scenarios where `:$` is the end of metric token name. For example: If a locator name happens to be `111:foo.bar.baz:$.qux`, we create tokens with id's `[111:foo, 111:foo.bar, **111:foo.bar.baz:$**, 111:foo.bar.baz:$.qux:$]` No problems yet. But if there is another metric which is `111:foo.bar.baz`,  we create tokens is id's `[111:foo, 111:foo.bar, **111:foo.bar.baz:$**]`. 

For metric `111:foo.bar.baz:$.qux`, **111:foo.bar.baz:$** is not a leaf node, but for metric `111:foo.bar.baz`, **111:foo.bar.baz:$** is a leaf node.

I think we need to add documentation that for metric names, metric tokens should not end with `:$`

blueflood-core/src/main/java/com/rackspacecloud/blueflood/types/Token.java
```diff
@@ -0,0 +1,147 @@
...
+
+    private final String token;
+    private final String parent;
+    private final boolean isLeaf;
+
+    private final String id;
+    private final Locator locator;
+
+    public Token(Locator locator, String[] tokens, int level) {
```
ChandraAddala: Thats the constructor we use. It has to be public. I will add some documentation.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/Discovery.java
ChandraAddala: Yes, but `Discovery` and `Token` are these separate implementations. I thought I will keep them separate.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/ElasticTokensIO.java
ChandraAddala: They are similar to the ones we have for regular metric search.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/ElasticTokensIO.java
ChandraAddala: I would use that class, even if I dont need it. :)

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionServer.java
ChandraAddala: Its upto the team. @usnavi What do you think? Shall I drop it?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/cache/LocatorCache.java
ChandraAddala: Since writing to tokens is a separate thread pool, there is a chance, a metric is written to `metric_metadata` but not to `metric_tokens`. So tracking both of them separately.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/ElasticTokensIO.java
ChandraAddala: Good point! :)

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/ElasticTokensIO.java
ChandraAddala: If guess we have to. Someone might choose * from each drop down and plot all metrics in one graph and we support that :)

blueflood-core/src/main/java/com/rackspacecloud/blueflood/cache/TokenCache.java
ChandraAddala: In `TokenCache`, I dont cache leaf node as a leaf node is same a locator name. I use `LocatorCache` for leaf nodes and for non-leaf nodes use `TokenCache`. This logic is in `TokenDiscoveryWriter.processTokens()` method.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/TokenDiscoveryIO.java
ChandraAddala: Lets discuss about the  "aggregated" method and create a story. For now, I am going to keep it.

shintasmith: Fair enough :-)

blueflood-core/src/main/java/com/rackspacecloud/blueflood/cache/LocatorCache.java
usnavi: gotcha.  Can you put a comment stating as much?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/processors/TokenDiscoveryWriter.java
usnavi: got it.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionServer.java
usnavi: Yeah, let's drop "EXP".

blueflood-core/src/main/java/com/rackspacecloud/blueflood/cache/LocatorCache.java
```diff
@@ -10,6 +10,20 @@
...
 
+/**
+ * This class is used to cache locator's that were written recently to our persistence layers by the available writers.
+ * All the writers, check the cache to see if it is written recently(isCurrent) before writing them again.
+ *
+ * Different writers that we have:
+ *
+ *  {@link com.rackspacecloud.blueflood.inputs.processors.BatchWriter} This writes to cassandra
+ *  {@link com.rackspacecloud.blueflood.inputs.processors.DiscoveryWriter} This supports metric discovery (/metric/search)
+ *  {@link com.rackspacecloud.blueflood.inputs.processors.TokenDiscoveryWriter} This support metric tokens discovery (metric_name/search)
```
shintasmith: 👍 😘  this helps a LOT

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionServer.java
ChandraAddala: Done.

PR #788, opened 2017-02-27
Remove old perf test scripts

izrik:
This PR removes the old perf test scripts, and instructs users to go to the new repo at https://github.com/rackerlabs/raxmetrics-perf-test-scripts


contrib/grinder/README.md
```diff
@@ -1,99 +1,5 @@
 # Blueflood Grinder Integration
-##Intro
-Grinder is a distributed load testing tool described [here](http://grinder.sourceforge.net/g3/getting-started.html)
 
-This code defines implementations of grinder worker threads meant to repeatedly invoke the required number of BF api calls during each "reporting interval"
+### DEPRECATED
```
ChandraAddala: I dont think we have to do this. If its possible, mention about the new grinder script in wiki or some place appropriate.

izrik: @ChandraAddala This is an open-source repo. Open-source users cant see our internal wiki. Were you thinking the github wiki instead?

ChandraAddala: yes.

shintasmith: @ChandraAddala , I am not sure what you mean by the "this" in "I dont think we have to do this". Do you mean we don't have to do this whole PR?

I don't want 2 grinder implementation that we have to maintain in 2 different places. We should pick one that we want to support and go with it.

ChandraAddala: I meant leaving the README file with a message saying "DEPRECATED"

izrik: Maybe @usnavi can weigh in on what we should do.

shintasmith: My vote is to delete the whole ```contrib/grinder``` directory :-)

usnavi: We don't need the DEPRECATED comments--delete it all.  We should have some docs which point users to the perf repos, though.

Perhaps in the main README we have a section entitled "Related Repos" which can list all the things we use in conjunction with blueflood.  This would include (I'm sure there are others)
- perf tests repos
- carbon forwarder
- blueflood finder
- statsD plugin
- I'm sure there are more.....

We could put this in the wiki on github, but I wonder if its more noticeable in the README.

izrik: @usnavi @ChandraAddala How does the README look?

PR #789, opened 2017-03-02
Removing check for empty tokens

ChandraAddala:
I have noticed that we are getting locators like this

ingest00.iad.prod.bf.k1k.me-org.openrepose.core.filters.HeaderNormalization.header-normalization..*_GET.count

Notice how the metric name as tokens which are empty. We had a condition to check if token is not empty before inserting data. As there is a possibility of empty token, I removed that check and made sure we still index documents appropriately.


blueflood-core/src/test/java/com/rackspacecloud/blueflood/types/TokenTest.java
```diff
@@ -69,7 +69,6 @@ public void testGetTokensHappyCase() {
                 tenantID + ":" + "a.b.c.d:$"};
 
         List<Token> tokens = Token.getTokens(locator);
-        tokens.forEach(System.out::println);
```
shintasmith: good catch! 👍

PR #792, opened 2017-03-09
Code fix for a bug to handle leaf queries with no regex.

ChandraAddala:
For metric name foo.bar.baz.qux, if a metric_name/search API is used with query=foo.bar.baz.qux, I am forming the wrong ES query inside. This code is to fix that. I also added a test case to cover that scenario.


izrik:
Looks ok to me.

PR #793, opened 2017-03-13
add instrumentation to count ingested datapoints per tenant

shintasmith:
This PR adds DropWizard Meters to count the number of ingested datapoints per tenant. There will be two meters for each tenant to count:
* number of normal (on time) metrics ingested
* number of delayed metrics ingested

The metrics are collected only if the configuration property ENABLE_PER_TENANT_METRICS is set to true (by default it's false).

### DIFF
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/Instrumentation.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/Instrumentation.java
@@ -19,6 +19,7 @@ package com.rackspacecloud.blueflood.io;
 import com.codahale.metrics.Meter;
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.Timer;
+import com.datastax.shaded.netty.util.internal.StringUtil;
 import com.netflix.astyanax.connectionpool.exceptions.ConnectionException;
 import com.netflix.astyanax.connectionpool.exceptions.PoolTimeoutException;
 import com.rackspacecloud.blueflood.utils.Metrics;
@@ -38,7 +39,6 @@ public class Instrumentation implements InstrumentationMBean {
     private static final Meter batchReadErrMeter;
 
     // One-off meters
-    private static final Meter scanAllColumnFamiliesMeter;
     private static final Meter allPoolsExhaustedException;
     private static final Meter fullResMetricWritten;
     private static final Meter fullResPreaggregatedMetricWritten;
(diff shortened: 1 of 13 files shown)


blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionServer.java
```diff
@@ -99,16 +108,23 @@ public void startServer() throws InterruptedException {
...
         router.get("/v1.0", new DefaultHandler());
-        router.post("/v1.0/multitenant/experimental/metrics", new HttpMultitenantMetricsIngestionHandler(processor, timeout));
-        router.post("/v1.0/:tenantId/experimental/metrics", new HttpMetricsIngestionHandler(processor, timeout));
-        router.post("/v1.0/:tenantId/experimental/metrics/statsd", new HttpAggregatedIngestionHandler(processor, timeout));
+        router.post("/v1.0/multitenant/experimental/metrics",
+                new HttpMultitenantMetricsIngestionHandler(processor, timeout, ENABLE_PER_TENANT_METRICS));
+        router.post("/v1.0/:tenantId/experimental/metrics",
+                new HttpMetricsIngestionHandler(processor, timeout, ENABLE_PER_TENANT_METRICS));
+        router.post("/v1.0/:tenantId/experimental/metrics/statsd",
+                new HttpAggregatedIngestionHandler(processor, timeout, ENABLE_PER_TENANT_METRICS));
```
ChandraAddala: Wouldnt it be simple to just get the property value inside each handler instead of changing the constructor for each handler?

ChandraAddala: I see, its for testing. Nevermind.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/Instrumentation.java
```diff
@@ -173,4 +168,18 @@ public static void markMetricsWithShortDelayReceived() {
     public static void markMetricsWithLongDelayReceived() {
         metricsWithLongDelayReceived.mark();
     }
+
+    public static Meter getIngestedMetricsMeter(String tenantId) {
+        if ( StringUtil.isNullOrEmpty(tenantId) ) {
+            return null;
+        }
+        return Metrics.meter(Instrumentation.class, "tenants", tenantId, "Data Points Ingested");
```
ChandraAddala: I just want to double check. The meter is created only once per tenant right? Eventhough this method is called several times?

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionServer.java
shintasmith: Yeah... it's easier for testing if I make it a parameter to constructor 😊

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/Instrumentation.java
shintasmith: Yes. Internally, in Codahale/Dropwizard, the meters etc is kept in a MetricsRegistry class, which holds all these objects in a HashMap.

PR #795, opened 2017-03-23
Add per-tenant metrics for aggregated/multi ingest requests

shintasmith:
It was an oversight that this was not added in the first PR to add per-tenant metrics. This PR adds the per-tenant metrics counting to aggregated/multi ingest requests.


ChandraAddala:
Glad you found it. No one would have ever known. You think this could be the reason why the total is slightly less than what we see in raw metrics/min?

shintasmith:
@ChandraAddala , there was only one tenant doing aggregated/multi, as far as I know. It's the CDN folks. Their volume is low, not as huge as Maas. So this won't add much to the per-tenant dashboard.

PR #797, opened 2017-04-04
use LOCAL_ONE consistency

shintasmith:
Since have upgraded to newer version of cassandra-maven-plugin, we are not limited to setting our consistency to ONE due to this bug: https://issues.apache.org/jira/browse/CASSANDRA-6238.

Should not impact performance that much. This is just cleaning up work.


ChandraAddala:
That is awesome!

PR #798, opened 2017-04-04
update datastax drivers to non-shaded

shintasmith:
Changes in this PR:
* update to latest datastax driver 3.2.0
* update to non-shaded version of datastax driver so we can leverage native Epoll on Linux platform. In perf01, it has shown to slightly improve user CPU usage
* change blueflood so it knows when to use Epoll vs Nio. The NettyUtils.java class is largely stolen from datastax driver :-)

### DIFF
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/Instrumentation.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/Instrumentation.java
@@ -56,14 +56,14 @@ public class Instrumentation implements InstrumentationMBean {
         metricsWithShortDelayReceived = Metrics.meter(kls, "Metrics with short delay received");
         metricsWithLongDelayReceived = Metrics.meter(kls, "Metrics with long delay received");
 
-            try {
-                final MBeanServer mbs = ManagementFactory.getPlatformMBeanServer();
-                final String name = String.format("com.rackspacecloud.blueflood.io:type=%s", Instrumentation.class.getSimpleName());
-                final ObjectName nameObj = new ObjectName(name);
-                mbs.registerMBean(new Instrumentation() { }, nameObj);
-            } catch (Exception exc) {
-                log.error("Unable to register mbean for " + Instrumentation.class.getSimpleName(), exc);
-            }
+        try {
+            final MBeanServer mbs = ManagementFactory.getPlatformMBeanServer();
+            final String name = String.format("com.rackspacecloud.blueflood.io:type=%s", Instrumentation.class.getSimpleName());
+            final ObjectName nameObj = new ObjectName(name);
+            mbs.registerMBean(new Instrumentation() { }, nameObj);
+        } catch (Exception exc) {
+            log.error("Unable to register mbean for " + Instrumentation.class.getSimpleName(), exc);
+        }
     }
 
     private Instrumentation() {/* Used for JMX exposure */}
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DAbstractMetricsRW.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DAbstractMetricsRW.java
@@ -83,7 +83,6 @@ public abstract class DAbstractMetricsRW extends AbstractMetricsRW {
         }
 
         Timer.Context ctx = Instrumentation.getWriteTimerContext( writeContexts.get( 0 ).getDestinationCF().getName() );
-
         try {
 
             BatchStatement batch = new BatchStatement(BatchStatement.Type.UNLOGGED);
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DBasicMetricsRW.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DBasicMetricsRW.java
@@ -10,7 +10,6 @@ import com.rackspacecloud.blueflood.outputs.formats.MetricData;
 import com.rackspacecloud.blueflood.rollup.Granularity;
 import com.rackspacecloud.blueflood.types.*;
 import com.rackspacecloud.blueflood.utils.Clock;
-import com.rackspacecloud.blueflood.utils.Metrics;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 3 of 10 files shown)


blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+In the past, to resolve conflict in versions, we chose to use Datastax driver with classifier 'shaded'. This essentially shades the Netty classes used by Datastax driver to have a different package name. Datastax doc clearly indicated that if the 'shaded' jar is used, then Netty can not be configured to use native Epoll, which has some performance gains, according to documentation.
+
+Therefore, in order for us leverage native Epoll, we deliberately choose the non-shaded jar of Datastax driver. But we must do so with care. We have to ensure that the Datastax driver for Cassandra is compatible with ours, i.e.: has the same major.minor version. This is the compatiblity matrix:
+
+Cassandra Driver | Built w/ Netty | Blueflood Netty
+Version          | Version        | Version
+-----------------|----------------|----------------
+3.1.2            | 4.0.37.FINAL   | 4.0.40.FINAL
+3.2.0            | 4.0.44.FINAL   | 4.0.44.FINAL
```
ChandraAddala: I feel like we have very tight coupling with datastax driver. I dont know how to avoid it.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionServer.java
```diff
@@ -282,6 +283,21 @@ public void setHttpEventsIngestionHandler(HttpEventsIngestionHandler httpEventsI
         }
     }
 
+    private void initNetty() {
+        int acceptThreads = Configuration.getInstance().getIntegerProperty(HttpConfig.MAX_WRITE_ACCEPT_THREADS);
+        int workerThreads = Configuration.getInstance().getIntegerProperty(HttpConfig.MAX_WRITE_WORKER_THREADS);
+
+        if ( NettyUtils.get().isEpollAvailable() ) {
```
ChandraAddala: Do you think, it can be beneficial to have a feature flag to control this behavior? Think of a scenario where we just want to turn off native epoll, even when its available.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/utils/NettyUtils.java
```diff
@@ -0,0 +1,40 @@
...
+import java.util.Locale;
+
+/**
+ */
+public class NettyUtils {
+
+    private static final Logger LOG = LoggerFactory.getLogger(NettyUtils.class);
+    private boolean useEpoll = false;
+
+    private static NettyUtils INSTANCE = new NettyUtils();
```
ChandraAddala: Can you make this final?

blueflood-http/src/main/java/com/rackspacecloud/blueflood/utils/NettyUtils.java
```diff
@@ -0,0 +1,40 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.util.Locale;
+
+/**
+ */
+public class NettyUtils {
+
+    private static final Logger LOG = LoggerFactory.getLogger(NettyUtils.class);
+    private boolean useEpoll = false;
```
ChandraAddala: Can you make this local variable?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/README.md
shintasmith: Yeah. I am also torn. The improvement on user CPU is not that huge. Maybe we should just stick with shaded?

blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpMetricsIngestionServer.java
shintasmith: Yeah I can have a config for this

blueflood-http/src/main/java/com/rackspacecloud/blueflood/utils/NettyUtils.java
shintasmith: What do you mean "local variable"? Isn't it already local? Do you mean final?

blueflood-http/src/main/java/com/rackspacecloud/blueflood/utils/NettyUtils.java
shintasmith: yup

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/README.md
ChandraAddala: If the improvement is not that significant, may be we should reconsider this as we are sacrificing a lot of flexibility.

blueflood-http/src/main/java/com/rackspacecloud/blueflood/utils/NettyUtils.java
ChandraAddala: I mean move it inside the method `isEpollAvailable`

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/README.md
shintasmith: Depends on how you look at the "flexibility". By using shaded jar, we are locked into not being able to use native epoll. So in away, that's also losing some flexibility.

ChandraAddala: I agree. But if we have to upgrade datastax, we have to upgrade netty also and that could get painful in the future esp when they change major version of netty. But if you feel like we can benefit from this native epoll, I am okay with the change.

shintasmith: I made a mistake of not putting this on a branch in my fork. So let me do that, then close the PR. If we decide we want to use native epoll, the changes should be there on a branch. We may not be able to see great improvement from using native epoll until we upgrade to later version of Linux kernel. We are using Ubuntu 13.x, which is using Linux kernel 3.x. There are a lot more performance improvements and troubleshooting tools in Linux kernel 4.x, which is what Ubuntu 16.x (and maybe 14.x) is built with. 

You know what I think of shaded jars. We have 3 versions of the netty jar in JVM. One for blueflood, one for datastax, and one for elasticsearch. That's insane, to me. :-)

shintasmith:
The changes are tucked in on my branch, if we ever need it
https://github.com/shintasmith/blueflood/tree/native_epoll

PR #799, opened 2017-04-07
add ENABLE_DTX_INGEST_BATCH option

shintasmith:
This PR adds a new configuration option: ENABLE_DTX_INGEST_BATCH. When set to true, blueflood will use unlogged batch statements during Ingest, if CASSANDRA_DRIVER is set to datastax. The default behavior for this option is false: i.e: use individual asynch INSERT statement (not unlogged batch).

If this option is set to true, both Basic metrics and Preaggregated metrics will use unlogged batches.


blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/io/BasicMetricsRWIntegrationTest.java
```diff
@@ -38,15 +37,13 @@
     private static final String TENANT2 = "987654";
     private static final String TENANT3 = "123789";
 
-    protected LocatorIO locatorIO = new DLocatorIO();
-    protected DelayedLocatorIO delayedLocatorIO = new DDelayedLocatorIO();
+    protected DLocatorIO locatorIO = new DLocatorIO();
```
ChandraAddala: Not a big fan of this change. Cant you cast `locatorIO` to `DLocatorIO` where ever you need to call `getBoundStatementForLocator`? Now that I think about it, I dont know which one is less bad. Nevermind then.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DAbstractMetricsRW.java
```diff
@@ -296,22 +313,60 @@ private RollupType getRollupType( Map<Long, Object> tsRollupMap ) {
     }
 
     /**
+     * Determines if the metric is considered delayed or not
+     *
+     * @param metric
+     * @return
+     */
+    protected boolean isDelayed(IMetric metric) {
```
ChandraAddala: Nice refactoring!

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DPreaggregatedMetricsRW.java
```diff
@@ -198,5 +170,97 @@ public DAbstractMetricIO getIO( String rollupType, Granularity granularity ) {
...
+                        futureLocatorMap.get(future), granularity), ex);
+            }
+        }
+    }
+
+    private void insertMetricsInBatch(Multimap<Locator, IMetric> map, Granularity granularity) {
+        BatchStatement batch = new BatchStatement(BatchStatement.Type.UNLOGGED);
+
+        for (Locator locator : map.keySet()) {
+            for (IMetric metric : map.get(locator)) {
```
ChandraAddala: Should we have` Instrumentation` call here?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DSetIO.java
```diff
@@ -58,4 +64,26 @@ protected BluefloodSetRollup fromByteBuffer(ByteBuffer byteBuffer) {
...
+     * @param metric
+     * @param granularity
+     * @return
+     */
+    @Override
+    protected BoundStatement getBoundStatementForMetric(IMetric metric, Granularity granularity) {
+        Object metricValue = metric.getMetricValue();
+        if (!(metricValue instanceof BluefloodSetRollup)) {
+            throw new InvalidDataException(
+                    String.format("getBoundStatementForMetric(locator=%s, granularity=%s): metric value %s is not type BfSetRollup",
```
ChandraAddala: Should we write the actual type `BluefloodSetRollup` instead of `BfSetRollup` in the exception message? This comment applies to other places like counter, gauge etc as well.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DSimpleNumberIO.java
```diff
@@ -24,4 +28,41 @@ protected Object fromByteBuffer( ByteBuffer byteBuffer ) {
...
+
+    /**
+     * Constructs a {@link BoundStatement} for a particular metric.
+     * The statement could be part of a BatchStatement or it could
+     * be executed as is.
+     *
+     * @param metric
+     * @return
+     */
+    public BoundStatement getBoundStatementForMetric(IMetric metric) {
```
ChandraAddala: `@Override` is missing

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DAbstractMetricIO.java
```diff
@@ -254,4 +254,14 @@ protected void addRollupToBatch(BatchStatement batch, Locator locator,
...
+
+    /**
+     * Retrieves the {@link BoundStatement} for a particular metric and granularity.
+     * Subclasses will implement this.
+     *
+     * @param metric
+     * @param granularity
+     * @return
+     */
+    protected abstract BoundStatement getBoundStatementForMetric(IMetric metric, Granularity granularity);
```
ChandraAddala: Cant we have some default implementation here? Since each sub classes seem to have similar implementation?

blueflood-core/src/integration-test/java/com/rackspacecloud/blueflood/io/BasicMetricsRWIntegrationTest.java
shintasmith: I think this is ok. Only the DBasicMetricsRW class needs locatorIO and DelayedLocatorIO for its constructor. The ABasicMetricsRW class doesn't need those IO classes. (Don't ask me why, tho).  😬 Since DBasicMetricsRW is Datastax specific, it is ok to pass in Datastax specific IO classes to its constructor.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/IOContainer.java
```diff
@@ -85,18 +86,20 @@ static synchronized void resetInstance() {
      * @param driver
      * @param isRecordingDelayedMetrics
      */
-    private IOContainer(DriverType driver, boolean isRecordingDelayedMetrics) {
+    private IOContainer(DriverType driver,
+                        boolean isRecordingDelayedMetrics,
+                        boolean isDtxIngestBatchEnabled) {
```
shintasmith: This change actually bothers me more than the change from LocatorIO to DLocatorIO in DBasicMetricsRWIntegrationTest. But I can't think of a better way (but still simple) way to do it.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DAbstractMetricIO.java
shintasmith: I can't think of a useful way to provide default implementation here. If you look at the implementation of this method in the sub classes, each line requires knowledge of what object (BluefloodSetRollup, BasicRollup, etc) you are operating on. I briefly thought of using generic types but decided it's not worth it.

Do you have specific suggestions?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DSetIO.java
shintasmith: I will change that.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DAbstractMetricIO.java
ChandraAddala: Each of the subclasses override an abstract method called `getRollupType()` which returns (BluefloodSetRollup, BasicRollup, etc). The `getBoundStatementForMetric` calls `getRollupType()` to get the type. I dont know if that covers all the issues.

If its too much hastle, I am fine with leaving the way it is now.

PR #804, opened 2017-04-25
create properties for the datastax connection pool settings

shintasmith:
This PR changes the way we configure Datastax's Connection Pool settings. Before this PR, we:
* call setCoreConnectionsPerHost() to the value of INITIAL_CASSANDRA_CONNECTIONS
* call setMaxConnectionsPerHost() to the value we calculate based on MAX_CASSANDRA_CONNECTIONS and number of DB/C* nodes

The MAX_CASSANDRA_CONNECTIONS is also used when configuring Astyanax connection pool.

With this PR, we:
* call setCoreConnectionsPerHost() to the value of DATASTAX_CORE_CONNECTIONS_PER_HOST
* call setMaxConnectionsPerHost() to the value of DATASTAX_MAX_CONNECTIONS_PER_HOST
* call setMaxRequestsPerConnection() to the value of DATASTAX_MAX_REQUESTS_PER_CONNECTION, with a default value of 1024.

MAX_CASSANDRA_CONNECTION is now only for configuring Astyanax connection pool.

### DIFF
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/IOConfig.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/IOConfig.java
@@ -105,8 +105,16 @@ public class IOConfig {
         return maxConns / numHosts + (maxConns % numHosts == 0 ? 0 : 1);
     }
 
-    public int getInitialConn() {
-        return config.getIntegerProperty(CoreConfig.INITIAL_CASSANDRA_CONNECTIONS);
+    public int getDatastaxInitialConnectionPerHost() {
+        return config.getIntegerProperty(CoreConfig.DATASTAX_CORE_CONNECTIONS_PER_HOST);
+    }
+
+    public int getDatastaxMaxConnectionPerHost() {
+        return config.getIntegerProperty(CoreConfig.DATASTAX_MAX_CONNECTIONS_PER_HOST);
+    }
+
+    public int getDatastaxMaxRequestsPerConnection() {
+        return config.getIntegerProperty(CoreConfig.DATASTAX_MAX_REQUESTS_PER_CONNECTION);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/IOConfig.java
```diff
@@ -105,8 +105,16 @@ public int getMaxConnPerHost(int numHosts) {
         return maxConns / numHosts + (maxConns % numHosts == 0 ? 0 : 1);
     }
 
-    public int getInitialConn() {
-        return config.getIntegerProperty(CoreConfig.INITIAL_CASSANDRA_CONNECTIONS);
+    public int getDatastaxInitialConnectionPerHost() {
+        return config.getIntegerProperty(CoreConfig.DATASTAX_CORE_CONNECTIONS_PER_HOST);
+    }
+
+    public int getDatastaxMaxConnectionPerHost() {
```
ChandraAddala: Nit pick `getDatastaxMaxConnectionsPerHost` notice `Connections`

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/IOConfig.java
```diff
@@ -105,8 +105,16 @@ public int getMaxConnPerHost(int numHosts) {
         return maxConns / numHosts + (maxConns % numHosts == 0 ? 0 : 1);
     }
 
-    public int getInitialConn() {
-        return config.getIntegerProperty(CoreConfig.INITIAL_CASSANDRA_CONNECTIONS);
+    public int getDatastaxInitialConnectionPerHost() {
+        return config.getIntegerProperty(CoreConfig.DATASTAX_CORE_CONNECTIONS_PER_HOST);
```
ChandraAddala: same here.

shintasmith:
@ChandraAddala , I fixed the method names. I also added logging for the range, when we're getting timeout exception

PR #805, opened 2017-05-03
add histogram to track how many datapoints per metric we try to rollup raw to 5Min

shintasmith:
We would like to know how many points we are fetching/reading when we are trying to rollup raw points from metrics_full to 5Min granularity. This would be useful in generating the right kind of load for performance testing.

### DIFF
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/astyanax/AstyanaxReader.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/astyanax/AstyanaxReader.java
@@ -200,7 +200,8 @@ public class AstyanaxReader extends AstyanaxIO {
 
             // we only want to count the number of points we
             // get when we're querying the metrics_full
-            if ( cf == CassandraModel.CF_METRICS_FULL || cf == CassandraModel.CF_METRICS_PREAGGREGATED_FULL ) {
+            // we don't do for aggregated or other granularities
+            if ( cf == CassandraModel.CF_METRICS_FULL ) {
                 Instrumentation.getRawPointsIn5MinHistogram().update(points.getPoints().size());
             }
         } catch (RuntimeException ex) {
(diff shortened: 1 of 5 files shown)


blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DAbstractMetricIO.java
```diff
@@ -244,6 +244,12 @@ protected void addRollupToBatch(BatchStatement batch, Locator locator,
             try {
                 List<Row> rows = future.getUninterruptibly().all();
 
+                // we only want to count the number of points we
+                // get when we're querying the metrics_full
+                if ( granularity == Granularity.FULL ) {
+                    Instrumentation.getRawPointsIn5MinHistogram().update(rows.size());
+                }
```
ChandraAddala: Does this work for regular as well as pre-aggregated?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/astyanax/AstyanaxReader.java
```diff
@@ -198,6 +197,12 @@ public static AstyanaxReader getInstance() {
             for (Column<Long> col : cols) {
                 points.add(new Points.Point<T>(col.getName(), (T)col.getValue(serializer)));
             }
+
+            // we only want to count the number of points we
+            // get when we're querying the metrics_full
+            if ( cf == CassandraModel.CF_METRICS_FULL || cf == CassandraModel.CF_METRICS_PREAGGREGATED_FULL ) {
+                Instrumentation.getRawPointsIn5MinHistogram().update(points.getPoints().size());
```
ChandraAddala: Should we have separate histograms for regular vs pre-aggregated?

shintasmith: I am not sure. Do you think we should?

ChandraAddala: I am thinking, we should only do this for regular metrics. Thats the one which is likely to have higher numbers.

shintasmith:
@ChandraAddala , I have updated the code so that it will only update the histogram for query against metrics_full ColumnFamily

Issue #806, opened 2017-05-04
Maven build tests fail on Ubuntu 16.04 with OpenJDK 1.8


basdusee:
Good morning,

The tests on blueflood masterbranch are not passing on my machine when running mvn test -X integration-test.
```java
testConfigTtl_valid(com.rackspacecloud.blueflood.cache.ConfigTtlProviderTest)  Time elapsed: 0.009 sec  <<< ERROR!
java.lang.IllegalStateException: Optional.get() cannot be called on an absent value
        at com.rackspacecloud.blueflood.cache.ConfigTtlProviderTest.testConfigTtl_valid(ConfigTtlProviderTest.java:50)

testConfigTtlForStrings(com.rackspacecloud.blueflood.cache.ConfigTtlProviderTest)  Time elapsed: 0 sec  <<< FAILURE!
java.lang.AssertionError
        at com.rackspacecloud.blueflood.cache.ConfigTtlProviderTest.testConfigTtlForStrings(ConfigTtlProviderTest.java:61)

Running com.rackspacecloud.blueflood.cache.SafetyTtlProviderTest
Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.003 sec - in com.rackspacecloud.blueflood.cache.SafetyTtlProviderTest
Running com.rackspacecloud.blueflood.cache.CombinedTtlProviderTest
Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0 sec - in com.rackspacecloud.blueflood.cache.CombinedTtlProviderTest
Running com.rackspacecloud.blueflood.cache.LocatorCacheTest
Tests run: 10, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 18.011 sec - in com.rackspacecloud.blueflood.cache.LocatorCacheTest

Results :

Failed tests:
  ConfigTtlProviderTest.testConfigTtlForStrings:61
Tests in error:
  ConfigTtlProviderTest.testConfigTtl_valid:50 » IllegalState Optional.get() can...

Tests run: 734, Failures: 1, Errors: 1, Skipped: 0
```
Disabling (commenting out actually) of the  com.rackspacecloud.blueflood.cache.ConfigTtlProviderTest.testConfigTtl_valid and com.rackspacecloud.blueflood.cache.ConfigTtlProviderTest.testConfigTtlForStrings works (of course) and maven finishes the blueflood-core tests without problems.

[...]

PR #807, opened 2017-05-10
Fix failing TTL tests run on Ubuntu 16.04 (Issue #806)

shintasmith:
This PR is the fix to Issue #806 .

While I was changing the code, I noticed we still have some code related to String metrics. I removed them as well.


izrik:
lgtm

PR #810, opened 2017-05-19
create a datastax RetryPolicy that can retry n times

shintasmith:
This PR creates a custom datastax RetryPolicy that can retry requests for up to n times.


blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/IOConfig.java
```diff
@@ -124,6 +124,13 @@ public int getRequestTimeout() {
         return config.getIntegerProperty(CoreConfig.CASSANDRA_REQUEST_TIMEOUT);
     }
 
+    public int getReadTimeoutMaxRetries() { return config.getIntegerProperty(CoreConfig.DATASTAX_READ_TIMEOUT_MAX_RETRIES); }
```
ChandraAddala: Why do you want to keep these methods in `IOConfig`?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/RetryNTimes.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    @Override
+    public RetryDecision onRequestError(Statement statement, ConsistencyLevel cl, DriverException e, int nbRetry) {
+        LOG.info(String.format("Trying nextHost on requestError: stmnt %s, consistency %s, driver ex %s, nbRetry %d",
+                statement, cl, e, nbRetry));
+        return RetryDecision.tryNextHost(cl);
+    }
+
+    @Override
```
ChandraAddala: why override this?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/RetryNTimes.java
```diff
@@ -0,0 +1,93 @@
...
+    public void init(Cluster cluster) {
+    }
+
+    @Override
+    public void close() {
```
ChandraAddala: same here.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/RetryNTimes.java
```diff
@@ -0,0 +1,93 @@
...
+        this.unavailableAttempts = unavailableAttempts;
+    }
+
+    @Override
+    public RetryDecision onReadTimeout(Statement stmnt, ConsistencyLevel cl,
+                                       int requiredResponses, int receivedResponses,
+                                       boolean dataReceived, int rTime) {
+        if (dataReceived) {
+            return RetryDecision.ignore();
+        } else if (rTime < readAttempts) {
```
ChandraAddala: If we set max read attempts to 3, does this retry 3 times or 2 times?

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/IOConfig.java
shintasmith: Because it's related to how we configure IO/driver? The other configuration, such as CASSANDRA_REQUEST_TIMEOUT, etc , are fetched from here as well. I thought I put them all in same place.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/RetryNTimes.java
shintasmith: It will retry 3 times, when rTime is:
* 0 < 3
* 1 < 3
* 2 < 3

When rTime reaches 3, it will go to the else and do a rethrow()

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/RetryNTimes.java
shintasmith: It's a method in the interface, I have to provide an implementation. I guess I can remove the ```@Override``` but I was keeping it the same as the other methods.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/RetryNTimes.java
ChandraAddala: Oh I see, I felt like the first time it re-tried rTime would be 1. I just verified, rTime indicates the number of retry already performed for this operation. Thanks for clarifying.

PR #813, opened 2017-06-01
rm obsolete contrib dir

shintasmith:
delete this Docker as it's moved to https://github.com/rackerlabs/docker-graphite-api-blueflood-finder


GeorgeJahad:
the new  dockerized finder with my changes is here:
https://github.com/rackerlabs/docker-graphite-api-blueflood-finder

I've confirmed it works, so this pr can be merged

PR #815, opened 2017-12-01
Mrit cass es docker compose

mritunjaykumar:
This PR contains a docker-compose file that helps in development process by running Cassandra and Elasticsearch as containers, and leave blueflood out of it. You can run or debug blueflood java code using IntelliJ and don't worry about Cassandra and Elasticsearch as they are taken care of by docker-compose file.
This will create two folders under your HOME directory cassdata (for persistence Cassandra data) and esdata (for persistence Elasticsearch data).

### DIFF
--- a/contrib/blueflood-docker-compose/docker-compose-cass-es.yml
+++ b/contrib/blueflood-docker-compose/docker-compose-cass-es.yml
@@ -28,13 +28,13 @@ services:
       - dc1ring
 
     # Maps cassandra data to a local folder. This preserves data across
-    # container restarts. Note a folder n1data get created locally
+    # container restarts. Note a folder 'cassdata' gets created locally in HOME directory.
     volumes:
       - ${HOME}/cassdata:/var/lib/cassandra
 
     # Docker container environment variable. We are using the
     # CASSANDRA_CLUSTER_NAME to name the cluster. This needs to be the same
-    # across clusters. We are also declaring that DC1N1 is a seed node.
+    # across clusters. We are also declaring that 'cassandra' is a seed node.
     environment:
       - CASSANDRA_CLUSTER_NAME=dev_cluster
           - CASSANDRA_SEEDS=cassandra
--- a/contrib/blueflood-docker-compose/docker-compose-cass-es.yml
+++ b/contrib/blueflood-docker-compose/docker-compose-cass-es.yml
@@ -1,0 +1,61 @@
+# using Docker-Compose Version 3
+version: '3'
+
+services:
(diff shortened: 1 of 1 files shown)


contrib/blueflood-docker-compose/docker-compose-cass-es.yml
```diff
@@ -0,0 +1,61 @@
...
+    command: bash -c 'if [ -z "$$(ls -A /var/lib/cassandra/)" ] ; then sleep 0; fi && /docker-entrypoint.sh cassandra -f'
+
+    # Network for the nodes to communicate
+    networks:
+      - dc1ring
+
+    # Maps cassandra data to a local folder. This preserves data across
+    # container restarts. Note a folder 'cassdata' gets created locally in HOME directory.
+    volumes:
+      - ${HOME}/cassdata:/var/lib/cassandra
```
itzg: An alternative is to let `docker-compose` manage retaining the volumes (until you do a `docker-compose down -v`) and use a named volume like

```
    - cassdata:/var/lib/cassandra
```

and then further down in declare a volumes section:

```
volumes:
    cassdata:
```

No need to change what you have, because there are advantages to what you've specified. Just food for thought.

contrib/blueflood-docker-compose/docker-compose-cass-es.yml
```diff
@@ -0,0 +1,61 @@
...
+
+    # Docker container environment variable. We are using the
+    # CASSANDRA_CLUSTER_NAME to name the cluster. This needs to be the same
+    # across clusters. We are also declaring that 'cassandra' is a seed node.
+    environment:
+      - CASSANDRA_CLUSTER_NAME=dev_cluster
+          - CASSANDRA_SEEDS=cassandra
```
itzg: The indent on this line seems to have an extra few spaces. I'm surprised yaml/compose didn't complain.

mritunjaykumar: Just fixed it. Good catch. I am surprised too. Unless this indentation somehow by mistake got added after the run.

izrik:
@mritunjaykumar Why does it create folders in the user's home directory? Shouldn't it just keep all data in the container?

mritunjaykumar:
@izrik - Folders are created in user's home directory to make sure data persists. Purpose of this docker-compose file is to help in development process when I would like to focus on the blueflood code and not worry about persisting data that I created during my earlier dev sessions.

itzg:
I know I'm beating a dead horse, but just wanted to point out that this confusion is alleviated by using named volumes rather than host mapped directories.

PR #819, opened 2018-01-11
Add end-to-end BF + Grafana docker compose example

itzg:


itzg:
I situated this new compose file in a new directory because compose derives the project name from the current directory name and in turn the container names from the project name. The other directory resulted in container names like "blueflooddockercompose_blueflood_1" and felt too long to me :)

In the directory I created the prefix becomes "bfgrana_".

contrib/bf-grafana/docker-compose.yml
```diff
@@ -0,0 +1,42 @@
...
+          - "9180:9180"
+
+  cassandra:
+    image: cassandra:2.1
+    ports:
+          - "9042:9042"
+          - "9160:9160"
+          - "7199:7199"
+
+  cassandran:
```
GeorgeJahad: having this separate "cassandran" service seems a bit wierd.  i'm guessing you copied that from somewhere.

itzg: Yep, this file started from https://github.com/rackerlabs/blueflood/blob/master/contrib/blueflood-docker-compose/docker-compose.yml

I'm assuming they wanted to create a little two-node cassandra cluster with the "n" instance. That one should be scalable with a `docker-compose scale cassandran=5` type command...but I haven't tried that yet here.

PR #820, opened 2018-01-12
*** This is still Work-in-progress ***

mritunjaykumar:
**# What**
Following is the comment from ChandraAddala on May 11, 2017
	Upgrading ES to 2.x
 	- Upgrading ES to 2.4.4 and lucene to 5.5.2
	- Changes to mapping document: Change from index_analyzer to analyzer
 	- Changing max limit from 100,000 to 10,000; Externalized this limit.
 	- Adding ESIntegTestCase for integration tests
 	- Changes to ingest ES client to ElasticIO and ElasticTokensIO
 	- Remove references to shaded libraries of old ES
 	- Removed old test library, elasticsearch-test
 	- Common logic for ES setup is placed in HttpESIntegrationBase
 
All of the above changes were done to upgrade the ES native client and a few were done to improve on tests.

Mritunjay's past changes:
1. Merge branch 'ChandraAddala-es-upgrade' (this branch contains changes related to 'Upgrading ES to 2.x') into check-es-pull-req.
2. Added test cases for discovery class.
3. Removed redundant and unnecessary artifacts from blueflood-core/pom and blueflood-elasticsearch/pom files and tested blueflood on ES 2.4.1.

Mritunjay's new changes:
Goal of these changes is to remove ES native client dependency, because for every ES version change, we have to make code changes. 
This is much bigger cost on any upgrade. So, ES native client library is being replaced by ES REST API call.
Code changes are done for index and query for metrics and events.
Following are the major changes:
1. Addition of a helper class ElasticsearchRestHelper to help make this REST API calls to ES.
2. Removed native client use, and updated metrics query code to perform the same business logic using REST API in AbstractElasticIO and code refactor because of this.
3. Removed native client use, and updated metrics indexing code to perform the same business logic using REST API in ElasticIO and code refactor because of this.
4. Removed native client use, and updated event query and search code to perform the same business logic using REST API in EventElasticSearchIO and code refactor because of this.

**# How to test**
Right now, many test cases are failing. It’s still in “Work In Progress” state.


GeorgeJahad:
since we are no longer using the native client, do we need to pull in chandra's changes at all?  would it be cleaner to just start from your changes and skip his entirely?

mritunjaykumar:
@GeorgeJahad - You are right. Since, I started with his change, I still have his changes in the code. Let me see, if I can completely skip it and apply only the latest changes.

mritunjaykumar:
PR is closed because there is another one which has much less changes. The other PR is related to just ES REST API addition to the code to replace native ES client.

PR #821, opened 2018-01-16
Replacing Elasticsearch native client with its REST API 

mritunjaykumar:
# What
Goal of these changes is to remove ES native client dependency, because for every ES version change, we have to make code changes.
This is much bigger cost on any upgrade. So, ES native client library is being replaced by ES REST API call.
Code changes are done for index and query for metrics and events.
Following are the major changes:

Addition of a helper class ElasticsearchRestHelper to help make this REST API calls to ES.
Removed native client use, and updated metrics query code to perform the same business logic using REST API in AbstractElasticIO and code refactor because of this.
Removed native client use, and updated metrics indexing code to perform the same business logic using REST API in ElasticIO and code refactor because of this.
Removed native client use, and updated event query and search code to perform the same business logic using REST API in EventElasticSearchIO and code refactor because of this.

# How to test
Go to "blueflood home" and run "./run-test.sh"


blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/ElasticsearchRestHelper.java
```diff
@@ -0,0 +1,261 @@
...
+            response = closeableHttpClient.execute(httpPost);
+            String str = EntityUtils.toString(response.getEntity());
+            EntityUtils.consume(response.getEntity());
+            return str;
+        }
+        catch(Exception e){
+            if((e instanceof HttpResponseException) && (response != null))
+                logger.error("Status code: " + response.getStatusLine().getStatusCode() + " with msg: " + e.getMessage());
+
+            throw new RuntimeException(e.getMessage(), e);
```
itzg: It's better to "enrich" the wrapping exception with context about what operation was attempted as the first argument.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/ElasticsearchRestHelper.java
```diff
@@ -0,0 +1,261 @@
...
+
+        if (StringUtils.isNotEmpty(untilValue) && StringUtils.isNotEmpty(fromValue)) {
+            rangeQueryString = String.format("{\"range\":{\"when\":{\"from\":%d,\"to\":%d}}}",
+                    Long.parseLong(fromValue), Long.parseLong(untilValue));
+        } else if (StringUtils.isNotEmpty(untilValue)) {
+            rangeQueryString = String.format("{\"range\":{\"when\":{\"to\":%d}}}", Long.parseLong(untilValue));
+        } else if (StringUtils.isNotEmpty(fromValue)) {
+            rangeQueryString = String.format("{\"range\":{\"when\":{\"from\":%d}}}", Long.parseLong(fromValue));
+        } else {
+            // TODO: LOG THIS INVALID CASE
```
itzg: Log it now so the TODO doesn't get forgotten about.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/EventElasticSearchIO.java
```diff
@@ -43,9 +40,16 @@
     public static final String ES_TYPE = "graphite_event";
     private final Client client;
```
itzg: Should `client` simply be removed?

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/EventElasticSearchIO.java
```diff
@@ -54,49 +58,89 @@ public EventElasticSearchIO(ElasticClientManager manager) {
...
-            bulk.add(requestBuilder);
-        }
-        bulk.execute().actionGet();
+        event.put(Event.FieldLabels.tenantId.toString(), tenantId);
+
+        elasticsearchRestHelper.indexEvent(event);
         eventInsertTimerContext.stop();
     }
 
+//    @Override
```
itzg: You could delete this commented out code.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/EventElasticSearchIO.java
```diff
@@ -54,49 +58,89 @@ public EventElasticSearchIO(ElasticClientManager manager) {
...
+//        eventInsertTimerContext.stop();
+//    }
+
     @Override
     public List<Map<String, Object>> search(String tenant, Map<String, List<String>> query) throws Exception {
         final Timer.Context eventSearchTimerContext = eventSearchTimer.time();
-        BoolQueryBuilder qb = boolQuery()
-                .must(termQuery(Event.FieldLabels.tenantId.toString(), tenant));
+        String result = elasticsearchRestHelper.fetchEvents(tenant, query);
+        eventSearchTimerContext.stop();
```
itzg: This `stop` should probably be in the `finally` of a `try` around the `fetchEvents` call.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/EventElasticSearchIO.java
```diff
@@ -54,49 +58,89 @@ public EventElasticSearchIO(ElasticClientManager manager) {
...
+            if(source.has("data")) map.put("data", source.get("data").asText());
+            if(source.has("tenantId")) map.put("tenantId", source.get("tenantId").asText());
+            if(source.has("tags")) map.put("tags", source.get("tags").asText());
+
+            eventResults.add(map);
+        }
+        return eventResults;
     }
 
+//    @Override
```
itzg: You could delete this old method.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/service/ElasticIOConfig.java
```diff
@@ -19,6 +19,7 @@
 public enum ElasticIOConfig implements ConfigDefaults {
 
     ELASTICSEARCH_HOSTS("127.0.0.1:9300"),
```
itzg: You could remove `ELASTICSEARCH_HOSTS` from here and `ConfigDefaults` to avoid any accidental attempts later to use the transport connection.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/service/RemoteElasticSearchServer.java
```diff
@@ -38,17 +38,17 @@ private RemoteElasticSearchServer() {
         Configuration config = Configuration.getInstance();
         List<String> hosts = config.getListProperty(ElasticIOConfig.ELASTICSEARCH_HOSTS);
         String clusterName = config.getStringProperty(ElasticIOConfig.ELASTICSEARCH_CLUSTERNAME);
-        Settings settings = ImmutableSettings.settingsBuilder()
-                .put("cluster.name", clusterName)
-                .build();
-        TransportClient tc = new TransportClient(settings);
+//        Settings settings = ImmutableSettings.settingsBuilder()
```
itzg: Go ahead and remove the commented out code.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/AbstractElasticIO.java
```diff
@@ -51,11 +57,63 @@
...
     public List<SearchResult> search(String tenant, List<String> queries) throws Exception {
+        // TODO: Add input validation for null, empty and all whitespaces.
         String[] indexes = getIndexesToSearch();
 
-        return searchESByIndexes(tenant, queries, indexes);
+        /*
+        Here I am changing the way we are searching ES. Native client call is commented out, and
+        instead I am enabling REST call to ES. This is done to remove dependency of BF code on ES version.
+         */
+        //return searchESByIndexes(tenant, queries, indexes);
```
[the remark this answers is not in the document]
mritunjaykumar: I thought of removing that earlier, but left it because of if people have any question. No problem, I will remove the commented code.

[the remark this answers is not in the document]
mritunjaykumar: Done.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/AbstractElasticIO.java
```diff
@@ -51,11 +57,63 @@
...
+        try {
+            queryBatchHistogram.update(queries.size());
+            String response = elasticsearchRestHelper.fetch(
+                    ELASTICSEARCH_INDEX_NAME_READ, ELASTICSEARCH_DOCUMENT_TYPE, tenant, queries.get(0));
+
+            List<SearchResult> searchResults = getSearchResults(response);
+            hitsCount = searchResults.size();
+            results.addAll(searchResults);
+        } catch (JsonProcessingException e) {
+            e.printStackTrace();
```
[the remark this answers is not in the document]
mritunjaykumar: oops.. good catch. Will do that right now

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/ElasticIO.java
```diff
@@ -57,47 +50,71 @@ public ElasticIO(ElasticClientManager manager) {
...
-                    fields.put(ESFieldLabel.unit.toString(), getUnit((Metric) metric));
-                }
-
-                discovery.withSourceFields(fields);
-                bulk.add(createSingleRequest(discovery));
             }
-            bulk.execute().actionGet();
+            elasticsearchRestHelper.indexMetrics(batch);
+        } catch (IOException e) {
+            log.error(e.getMessage());
```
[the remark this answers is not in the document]
mritunjaykumar: Since, there was no logging at all earlier in any of these classes, I added that to make sure at least we are logging something. :)
I will add more to it in all of the classes metrics and event based.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/EventsIO.java
```diff
@@ -20,6 +20,6 @@
 import java.util.Map;
 
 public interface EventsIO {
-    public void insert(String tenant, List<Map<String, Object>> metrics) throws Exception;
-    public List<Map<String, Object>> search(String tenant, Map<String, List<String>> query) throws Exception;
+    void insert(String tenantId, Map<String, Object> event) throws Exception;
+    List<Map<String, Object>> search(String tenant, Map<String, List<String>> query) throws Exception;
```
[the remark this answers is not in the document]
mritunjaykumar: Changed search parameter to 'tenantId'

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/EventElasticSearchIO.java
GeorgeJahad: yes please, all the commented out code should be deleted:
1. it makes the diffs easier to read, (that way the old code is diffed against the new code, instead of both appearing.)
2. we don't want the code to be littered with multiple old versions.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/EventsIO.java
```diff
@@ -20,6 +20,6 @@
 import java.util.Map;
 
 public interface EventsIO {
-    public void insert(String tenant, List<Map<String, Object>> metrics) throws Exception;
-    public List<Map<String, Object>> search(String tenant, Map<String, List<String>> query) throws Exception;
+    void insert(String tenantId, Map<String, Object> event) throws Exception;
```
GeorgeJahad: It looks like you've changed the second parameter here from a list of maps, to a map.  any particular reason?  Isn't this api exposed to the users?

mritunjaykumar: https://github.com/rackerlabs/blueflood/blob/master/blueflood-http/src/main/java/com/rackspacecloud/blueflood/inputs/handlers/HttpEventsIngestionHandler.java#L69
At this line, we are deserializing request body into Event instance, and it accepts only one event object. There is no array here. However, we are wrapping that event object into a collection which will always have only one object. So, there was unnecessary wrapping around one object. I just removed it and made corresponding changes.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/AbstractElasticIO.java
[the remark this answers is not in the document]
mritunjaykumar: done

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/ElasticIO.java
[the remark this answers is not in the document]
mritunjaykumar: done

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/ElasticsearchRestHelper.java
mritunjaykumar: done

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/ElasticsearchRestHelper.java
mritunjaykumar: Done.. :-)

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/EventElasticSearchIO.java
mritunjaykumar: yep. all cleaned up.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/EventElasticSearchIO.java
mritunjaykumar: yes. I agree. I deleted all of the commented code.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/EventElasticSearchIO.java
mritunjaykumar: Didn't notice that. There was no try catch block earlier. So, looks like it skip through my mind. Now, I added try/catch/finally in all such methods.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/AbstractElasticIO.java
```diff
@@ -51,101 +46,57 @@
...
-
-            response = client.prepareSearch(indexes)
-                    .setRouting(tenant)
-                    .setSize(MAX_RESULT_LIMIT)
-                    .setVersion(true)
-                    .setQuery(bqb)
-                    .execute()
-                    .actionGet();
+            String response = elasticsearchRestHelper.fetch(
+                    ELASTICSEARCH_INDEX_NAME_READ, ELASTICSEARCH_DOCUMENT_TYPE, tenant, queries.get(0));
```
GeorgeJahad: It looks like the old code was using all the elements in the "queries" list, but here you only use the first.  why?

mritunjaykumar: I think, I may need to fix this. Earlier I saw that it's being called by search with query collection param and which itself is getting called by another search method with only one query parameter. Hence, I made that change. But, I just found out that it might be getting called from RollupHandler.getRollupByGranularity() method. So, I will dig into this further. Good catch.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/TokenDiscoveryIO.java
```diff
@@ -6,7 +6,7 @@
 import java.io.IOException;
 import java.util.List;
 
-public interface TokenDiscoveryIO  extends MetricNameSearchIO {
```
GeorgeJahad: If this module extends metricnameSearch, and we are no longer using that, then do we still this module?

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/AbstractElasticIO.java
```diff
@@ -51,101 +46,57 @@
...
-
-                bqb.should(boolQuery()
-                                .must(termQuery(ESFieldLabel.tenantId.toString(), tenant))
-                                .must(qb)
-                );
-            }
-
-            response = client.prepareSearch(indexes)
-                    .setRouting(tenant)
-                    .setSize(MAX_RESULT_LIMIT)
```
GeorgeJahad: What does this limit do, and does it need to be preserved in the rest client?

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/AbstractElasticIO.java
```diff
@@ -51,101 +46,57 @@
...
-                }
-
-                bqb.should(boolQuery()
-                                .must(termQuery(ESFieldLabel.tenantId.toString(), tenant))
-                                .must(qb)
-                );
-            }
-
-            response = client.prepareSearch(indexes)
-                    .setRouting(tenant)
```
GeorgeJahad: what does this routing do?

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/ElasticsearchRestHelper.java
```diff
@@ -0,0 +1,266 @@
+package com.rackspacecloud.blueflood.io;
```
GeorgeJahad: this file, blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/ElasticsearchRestHelper.java
, has a number of intellij warnings.  please take a look at those and see if they need handling.

GeorgeJahad:
you've removed quite a bit of code related to the old metrics_token/metrics_name index.  Please add a section to the PR header explaining our rational and maybe list the effected files, so people can understand what we decided.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/AbstractElasticIO.java
mritunjaykumar: Routing is used in current code. Routing restrict searches to only given node based on the specified routing.

blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/outputs/handlers/HttpRollupHandlerWithESIntegrationTest.java
```diff
@@ -91,7 +91,7 @@ public void setup() throws Exception {
         metricsRW.insertMetrics(metrics);
 
         httpHandler = new HttpRollupsQueryHandler();
-        ((ElasticIO) ModuleLoader.getInstance(DiscoveryIO.class, CoreConfig.DISCOVERY_MODULES)).setClient(esSetup.client());
+        //((ElasticIO) ModuleLoader.getInstance(DiscoveryIO.class, CoreConfig.DISCOVERY_MODULES)).setClient(esSetup.client());
```
itzg: Can this be deleted?

mritunjaykumar:
Last checked (yesterday) code (line) coverage on ElasticsearchRestHelper was at 87%. And yes it already had newer version of junit I think. This was causing some “deprecated” warnings in the code.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/TokenDiscoveryIO.java
mritunjaykumar: I didn't remove it because there are some files that are using it. So, at this point of time, I am trying to cleanup things which are around the functionality, and not go deeper into code where I don't know what they are doing and refactor them.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/AbstractElasticIO.java
mritunjaykumar: This sets the limit of the returned result size.

blueflood-elasticsearch/pom.xml
```diff
@@ -71,13 +65,6 @@
       <scope>test</scope>
     </dependency>
 
-    <dependency>
-      <groupId>junit</groupId>
-      <artifactId>junit</artifactId>
```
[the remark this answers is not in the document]
mritunjaykumar: JUnit dependency was removed because of the reason you mentioned. 

I don't have the prior coverage. But, right now code coverage on files that I touched are below:
- AbstractElasticIO               - 88% (Line Coverage)
- ElasticIO                             - 81% (Line Coverage)
- ElasticsearchRestHelper   - 92% (Line Coverage)
- EventElasticsearchIO .      - 80% (Line Coverage)

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/ElasticsearchRestHelper.java
```diff
@@ -233,13 +263,22 @@ private void index(HttpPost httpPost, String bulkString) throws IOException {
...
             response = closeableHttpClient.execute(httpPost);
 
-            logger.info("Status: " + response.getStatusLine().getStatusCode());
+            int statusCode = response.getStatusLine().getStatusCode();
+
+            if(statusCode != HttpStatus.SC_OK && statusCode != HttpStatus.SC_CREATED){
+                logger.error("index method failed with status code: {} and error: {}",
+                        response.getStatusLine().getStatusCode(),
+                        EntityUtils.toString(response.getEntity()));
+                System.out.println(EntityUtils.toString(response.getEntity()));
```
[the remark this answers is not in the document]
mritunjaykumar: removed now. I will check-it in a bit.

blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/http/HttpIntegrationTestBase.java
```diff
@@ -157,9 +156,9 @@ private static void setupElasticSearch() {
                 .withMapping("graphite_event", EsSetup.fromClassPath("events_mapping.json")));
 
         // create elaticsearch client and link it to ModuleLoader
-        elasticIO = new ElasticIO(esSetup.client());
-        eventsSearchIO = new EventElasticSearchIO(esSetup.client());
-        ((ElasticIO) ModuleLoader.getInstance(DiscoveryIO.class, CoreConfig.DISCOVERY_MODULES)).setClient(esSetup.client());
+        elasticIO = new ElasticIO();
+        eventsSearchIO = new EventElasticSearchIO();
+        //((ElasticIO) ModuleLoader.getInstance(DiscoveryIO.class, CoreConfig.DISCOVERY_MODULES)).setClient(esSetup.client());
```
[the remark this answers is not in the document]
mritunjaykumar: Done

blueflood-http/src/integration-test/java/com/rackspacecloud/blueflood/outputs/handlers/HttpRollupHandlerWithESIntegrationTest.java
mritunjaykumar: Done

mritunjaykumar:
I have incorporated code-review comments. Also, now, all of the test cases are passing.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/AbstractElasticIO.java
```diff
@@ -51,101 +46,57 @@
...
-                    .setSize(MAX_RESULT_LIMIT)
-                    .setVersion(true)
-                    .setQuery(bqb)
-                    .execute()
-                    .actionGet();
+            String response = elasticsearchRestHelper.fetch(
+                    ELASTICSEARCH_INDEX_NAME_READ, ELASTICSEARCH_DOCUMENT_TYPE, tenant, queries);
```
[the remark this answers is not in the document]
mritunjaykumar: As discussed, it's fine line that how far should I go to clean up because it might lead me into some code which is not tied to the current functionality and it *might* be a distraction.. IMHO

blueflood-elasticsearch/src/test/java/com/rackspacecloud/blueflood/io/ElasticIOIntegrationTest.java
```diff
@@ -21,143 +21,152 @@
...
-        elasticIO = new ElasticIO(esSetup.client());
 
+        elasticIO = new ElasticIO();
         elasticIO.insertDiscovery(createTestMetrics(TENANT_A));
         elasticIO.insertDiscovery(createTestMetrics(TENANT_B));
         elasticIO.insertDiscovery(createTestMetricsFromInterface(TENANT_C));
 
-        esSetup.execute(EsSetup.createIndex(ElasticTokensIO.ELASTICSEARCH_TOKEN_INDEX_NAME_WRITE)
-                               .withMapping("tokens", EsSetup.fromClassPath("tokens_mapping.json")));
+        Thread.sleep(3*1000); // Sleep a few seconds to make sure data is indexed.
```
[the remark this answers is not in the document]
mritunjaykumar: I couldn't use it because I am not using native client. When I was digging into it, it appeared that it was more tied to the native client.

blueflood-elasticsearch/src/test/java/com/rackspacecloud/blueflood/io/ElasticIOIntegrationTest.java
```diff
@@ -21,143 +21,152 @@
...
-                               .withMapping("tokens", EsSetup.fromClassPath("tokens_mapping.json")));
+    /*
+    Once done testing, delete all of the records of the given type and index.
+    NOTE: Don't delete the index or the type, because that messes up the ES settings.
+     */
+    @AfterClass
+    public static void tearDownClass() throws Exception {
+        URIBuilder builder = new URIBuilder().setScheme("http")
+                .setHost("127.0.0.1").setPort(9200)
+                .setPath("/metric_metadata/metrics/_query");
```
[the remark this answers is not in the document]
mritunjaykumar: Yes.. that's right. I think I did add that to the comment. I might not have added that into this class. 
https://github.com/rackerlabs/blueflood/blob/c5030ac09c4c7f5dc878d9b0344317ba60572c23/blueflood-elasticsearch/src/test/java/com/rackspacecloud/blueflood/io/EventElasticSearchIOTest.java#L174

blueflood-elasticsearch/src/test/java/com/rackspacecloud/blueflood/io/ElasticIOIntegrationTest.java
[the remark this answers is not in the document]
itzg: As you can see in that docs link, the `_refresh` endpoint is available via REST.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/AbstractElasticIO.java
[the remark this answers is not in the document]
itzg: Now that I see the context, I'm actually asking the opposite. I wanted to confirm your effective removal of that "token index" doesn't impact business logic of blueflood in general. I'm worried about logic that might not be testable via the ES unit tests, but might break something at an integration level.

blueflood-elasticsearch/src/test/java/com/rackspacecloud/blueflood/io/ElasticIOIntegrationTest.java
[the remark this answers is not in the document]
mritunjaykumar: let me see. I will try to dig this again.. :)

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/AbstractElasticIO.java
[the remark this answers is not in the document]
mritunjaykumar: As discussed, ALL of the integration tests (including core and http) are passing on ES versions 1.7, 2.4.1, and 5.6.5.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/ElasticsearchRestHelper.java
```diff
@@ -347,4 +349,22 @@ private String getBoolQueryString(String mustValueString, String shouldValueStri
     public void setBaseUrlForTestOnly(String url) {
         baseUrl = url;
     }
+
+    @VisibleForTesting
+    public int refreshIndex(String indexName) throws IOException {
```
itzg: Yay, so this worked for you I assume?

mritunjaykumar: Yes. It was good to switch out of hard-coded sleep. I still see many legacy *sleep* code in there. I think when we are touching that piece of functionality, we will take care of them at that time.

metrics-dev:
Yes. I will wait for George’s green signal to make sure we all are in agreement on this as this is a major code change.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/types/Locator.java
```diff
@@ -85,6 +85,9 @@ public boolean equals(Locator other) {
     }
 
     public static Locator createLocatorFromPathComponents(String tenantId, String... parts) throws IllegalArgumentException {
+        if(StringUtils.isEmpty(tenantId) || parts == null)
+            throw new IllegalArgumentException("'tenantId' is null or empty OR 'parts' is null.");
```
GeorgeJahad: why are we throwing an exception here instead of logging an error?  
what will happen to the thread if an exception is thrown?  will the server go down?
i'm fine with checking inputs when received from the outside parties, but adding exceptions to this code seems dangerous without knowing the consequences.  am i missing something?

mritunjaykumar: Based on our discussion, I have replaced exception with log.error() call.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/types/Locator.java
```diff
@@ -86,7 +86,7 @@ public boolean equals(Locator other) {
 
     public static Locator createLocatorFromPathComponents(String tenantId, String... parts) throws IllegalArgumentException {
         if(StringUtils.isEmpty(tenantId) || parts == null)
-            throw new IllegalArgumentException("'tenantId' is null or empty OR 'parts' is null.");
+            log.error("'tenantId' is null or empty OR 'parts' is null.");
```
itzg: So now it'll create a bogus Locator. Will that just cause the query to not match anything? I prefered the exception throwing, @GeorgeJahad, since that can be ultimately caught by the web handler layer and translated into a 400 back to the client. With all that said, I don't mind either way since you two have discussed this.

GeorgeJahad: I guess my philosophy is that user input should be checked when that input first enters the system, not creating exceptions throughout the system.

As I discussed with Mritunjay, unless we are sure these exceptions are properly handled somewhere, I'm concerned that they will cause the thread to halt, eventually bringing the system to its knees.

I guess the correct solution is to just be sure we have the correct exception handlers everywhere, but I'm not convinced that's trivial, and opens up a whole new issue that I really don't want to address in this PR.

If you/Jay feel strongly, I would support opening a new PR to address those issues, but this one is so complicated already I don't want to add another set of issues to it.  What do you guys think, @itzg / @mritunjaykumar ?

mritunjaykumar: I agree with @itzg . That's why I implemented that way earlier. But, @GeorgeJahad had a valid point that what if this exception blows something. So, I removed that exception just based on that fear. This check-in is massive in itself as it's literally unrooting native ES client calls, and adding another variable to the mix was kind of debatable risk particularly in this check-in. IMHO, we can take this work, when we know for certain that it has no impact on other code. Besides, AFAIK, it will never hit this code path based on whatever I have seen so far. So, I would let it be there for the next refactor.

run-test.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+BASE_DIR=`pwd`
+
+# Elasticsearch setup
+cd $BASE_DIR/blueflood-elasticsearch/src/main/resources 
+./init-es.sh
```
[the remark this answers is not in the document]
mritunjaykumar: I will check.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/AbstractElasticIO.java
mritunjaykumar: That's Elasticsearch max_result_window limit. https://www.elastic.co/guide/en/elasticsearch/reference/2.4/index-modules.html.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/AbstractElasticIO.java
```diff
@@ -37,69 +37,102 @@
...
-                bqb.should(boolQuery()
-                                .must(termQuery(ESFieldLabel.tenantId.toString(), tenant))
-                                .must(qb)
-                );
-            }
-
-            response = client.prepareSearch(indexes)
-                    .setRouting(tenant)
-                    .setSize(MAX_RESULT_LIMIT)
-                    .setVersion(true)
```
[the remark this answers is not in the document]
mritunjaykumar: I didn't see any function related to this in Elasticsearch rest api. When I was going through this, AFAIK, this was not needed. I can dig further to confirm.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/AbstractElasticIO.java
GeorgeJahad: my question is that the new code seems to have changed the max from 100,000 to 10,000.  Is that a true statement?  if so, why?

mritunjaykumar: As discussed, this limit is added in the newer version of Elasticsearch. Here is the exception that elasticsearch throws:
`QueryPhaseExecutionException[Result window is too large, from + size must be less than or equal to: [10000] but was [100000]. See the scroll api for a more efficient way to request large data sets. This limit can be set by changing the [index.max_result_window] index level parameter.]`

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/ElasticsearchRestHelper.java
```diff
@@ -0,0 +1,584 @@
...
+                if(response != null) {
+                    response.close();
+                }
+            }
+        }
+
+        return "";
+    }
+
+    private String getQueryDslString(String tenantId, List<String> queries){
```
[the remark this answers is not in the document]
mritunjaykumar: Sure.. Let's go through this tomorrow on zoom.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/ElasticsearchRestHelper.java
```diff
@@ -0,0 +1,584 @@
...
+                String str = EntityUtils.toString(response.getEntity());
+                EntityUtils.consume(response.getEntity());
+                return str;
+            } catch (Exception e) {
+                if(response == null){
+                    logger.error("fetchEvents failed with message: {}", e.getMessage());
+                    url = String.format("%s/%s/%s/_search?routing=%s&size=%d", getBaseUrl(),
+                            EventElasticSearchIO.EVENT_INDEX, EventElasticSearchIO.ES_TYPE,
+                            tenantId, MAX_RESULT_LIMIT);
+                    callQ.add(url);
```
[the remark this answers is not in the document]
mritunjaykumar: we don't need to. If we get exception in the response, we should not retry the same path.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/ElasticsearchRestHelper.java
```diff
@@ -0,0 +1,584 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(ElasticsearchRestHelper.class);
+    private final CloseableHttpClient closeableHttpClient;
+    private static String[] baseUrlArray;
+    private static int baseUrlIndex;
+    private static int MAX_CALL_COUNT = 10;
+    private static int MAX_RESULT_LIMIT =
+            Configuration.getInstance().getIntegerProperty(CoreConfig.MAX_DISCOVERY_RESULT_SIZE);
+
+    public static ElasticsearchRestHelper getInstance() {
+        return new ElasticsearchRestHelper();
```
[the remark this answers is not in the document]
mritunjaykumar: Earlier, I had only one instance of this class across the whole application. When I started on this work, I had a fear that multiple instances might interfere with the same dataset. But, when I learned more and realized that every service (ingest, query, etc) is creating its own set of instances, I was comfortable opening up and giving each service their own instance of this helper class. That way those instances will not interfere with each other, and we will have more instances to take care of ES work.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/ElasticsearchRestHelper.java
```diff
@@ -0,0 +1,584 @@
...
+            CloseableHttpResponse response = null;
+
+            try {
+                response = closeableHttpClient.execute(httpPost);
+                statusCode = response.getStatusLine().getStatusCode();
+
+                if (statusCode != HttpStatus.SC_OK && statusCode != HttpStatus.SC_CREATED) {
+                    logger.error("index method failed with status code: {} and error: {}",
+                            response.getStatusLine().getStatusCode(),
+                            EntityUtils.toString(response.getEntity()));
```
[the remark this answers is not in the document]
mritunjaykumar: Other places where I am using consume(), I am returning string from the function. So, to return string I need to make sure I read the whole stream and close it. consume() method ensures that.
https://hc.apache.org/httpcomponents-core-ga/httpcore/apidocs/org/apache/http/util/EntityUtils.html#consume(org.apache.http.HttpEntity)

Here, in this case, I am not reading response entity, so I don't need to consume the whole string. However, I do feel that I should call consume() regardless to close the response stream. Good catch.

blueflood-elasticsearch/src/test/java/com/rackspacecloud/blueflood/io/ElasticIOIntegrationTest.java
```diff
@@ -312,39 +395,103 @@ public void testDeDupMetrics() throws Exception {
...
+            }
         }
+
+        assertEquals(2, matchingTerms.size());
+        assertEquals("foo.bar", matchingTerms.get(0));
+        assertEquals("foo.bar.baz", matchingTerms.get(1));
     }
 
     @Test
-    @Parameters({"elasticIO", "elasticTokensIO"})
```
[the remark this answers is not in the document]
mritunjaykumar: I noticed that because of this, it's missing coverage on one method. I will add test cases to cover getMetricNames() which is in AbstractElasticIO class.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/ElasticsearchRestHelper.java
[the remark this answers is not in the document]
mritunjaykumar: Explanation from our chat session:
Earlier boolean expression (in native client):
(tenantId && query1) || (tenantId && query2) || (tenantId && query3)
And now it is (in ES REST API):
tenantId && (query1 || query2 || query3)

blueflood-elasticsearch/src/test/java/com/rackspacecloud/blueflood/io/ElasticIOIntegrationTest.java
```diff
@@ -157,92 +132,64 @@ public String getMethod() {
...
+        Stream<Locator> locators = metrics.stream().map(IMetric::getLocator);
+        elasticTokensIO.insertDiscovery(Token.getUniqueTokens(locators)
+                .collect(toList()));
     }
 
     @Test
     public void testNoCrossTenantResults() throws Exception {
         List<SearchResult> results = elasticIO.search(TENANT_A, "*");
-        assertEquals(NUM_DOCS + 2, results.size());
+        assertEquals(NUM_DOCS, results.size());
```
itzg: Changing the test expectations worries me.

mritunjaykumar: Just to make sure you have the context.. :-) This file is a complete redo. When you are comparing this file, please compare with the "master" branch file. There is zero change in the test cases except one and even that change is done to replace native client to the REST API call.

mritunjaykumar: Also, regarding change on expectations in the test case is tied to the test data got entered in the setup. But, at this point of time it's not relevant anymore.

blueflood-elasticsearch/src/main/resources/init-es.sh
```diff
@@ -64,13 +64,17 @@ checkFile tokens_mapping.json
 curl $AUTH -XDELETE $ELASTICSEARCH_URL'/blueflood_initialized_marker/' >& /dev/null
 curl $AUTH -XDELETE $ELASTICSEARCH_URL'/events/' >& /dev/null
 curl $AUTH -XDELETE $ELASTICSEARCH_URL'/metric_metadata/' >& /dev/null
+curl $AUTH -XDELETE $ELASTICSEARCH_URL'/metric_metadata_2/' >& /dev/null
```
GeorgeJahad: i don't understand the changes to this file.  Are they needed?

GeorgeJahad:
In the "WHAT" section above, I think the final paragraph is out of date.  Please review it and correct if needed, (it says the token index was removed.)

metrics-dev:
Removed final paragraph in the “WHAT” section which was not correct at this time.

PR #822, opened 2018-03-08
Mrit build automation

mritunjaykumar:
## Prerequisite 
Docker is running on the server.
## What it does
- Creates Cassandra and Elasticsearch containers
- Create Schema in Cassandra, and sets up elasticsearch indices
- Runs all of the test cases
- Removes the containers at the end of the test run


run-integration-test.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+BASE_DIR=`pwd`
+
+# Create Cassandra and Elasticsearch containers to run the integration test
+docker-compose -f contrib/blueflood-docker-compose/docker-compose-integration-test.yml up -d
+
+# Copy Cassandra schema file into Cassandra container
+docker cp src/cassandra/cli/load.cdl cassandra-node:/tmp/load.cdl
```
itzg: FYI you could volume-attach the file into the container, but not worth changing.

metrics-dev:
I noticed. Sorry about that. That was not the intention. ☺

PR #850, opened 2021-09-01
unsynchronizes LocatorCache

zzantozz:
The Rackspace Blueflood instance was having problems with its
ingestion rate, and lots of threads were sitting and waiting for
access to the LocatorCache, especially as the number of threads in the
thread pools were increased. Since it's already using a thread-safe
Guava cache implementation, there should be no reason for
synchronizing access to the cache overall. Looking back at the commit
history, it appears it was synchronized because of flaky tests. This
is a terrible reason to play with synchronization in Java. Rely on
high-level concurrency constructs, like the Guava cache, to control
concurrency.

This removes all instances of 'synchronized' from the LocatorCache
class and rewrites the unit tests to be much more robust and faster. A
follow-up change will add a new test of cache concurrency, but it
fails with the current implementation because of the mutable cached
object.

### DIFF
--- a/blueflood-core/pom.xml
+++ b/blueflood-core/pom.xml
@@ -374,13 +374,26 @@
         <version>3.18.2-GA</version>
         <scope>test</scope>
     </dependency>
-
+    <dependency>
+      <groupId>org.hamcrest</groupId>
+      <artifactId>hamcrest-library</artifactId>
+    </dependency>
+    <dependency>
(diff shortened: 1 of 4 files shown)


iWebi (approved):
mustNotBeSynchronized test is cool. **You shall not pass!** with synchronized

zzantozz:

PR #851, opened 2021-09-02
fixes failing integration test

zzantozz:
This creates a conventional 'resources' dir in each of the
integration-test directories and places config files in them to be
loaded for integration testing.

The HttpEventsQueryHandlerIntegrationTest failed when I ran 'mvn
verify'. The reason for failure was that it was expected to return
CORS headers and didn't. I traced this back to the fact that
HttpMetricsIngestionServerShutdownIntegrationTest ran first, which
caused the HttpResponder class to be initialized and read the default
CORS_ENABLED value of false. Later, when the setting was changed by
HttpIntegrationTestBase, which the failing class inherits from, it
didn't affect HttpResponder because it uses static fields and
initialization that can only happen once.

The simplest solution would be to have the shutdown IT inherit from
the common base class, but it starts and manipulates its own server,
which conflicts with those started by the base class.

This means the integration tests need to be configurable without
assuming all will inherit from a common base class.  Along the way, I
discovered that Maven and IntelliJ were treating the failsafe plugin's
'argLine' settings differently, and it was broken in both cases. The
sum of the changes in this commit can be described as migrating the
settings from both the HTTP IT base class and the argLine setting into
src/integration-test/resources, where they arguably belong anyway.

Note: I omitted the JVM settings in the argLine because MaxPermSize is
no longer a thing, and if we need to increase the heap space, we can
do it in something like .mvn/jvm.config.

### DIFF
--- a/blueflood-http/src/integration-test/resources/blueflood.properties
+++ b/blueflood-http/src/integration-test/resources/blueflood.properties
@@ -1,0 +1,17 @@
+# No need for zookeeper for integration testing.
+# I believe this is because zk is only/primarily used to coordinator multiple rollup nodes so a single shard doesn't
+# get rolled up multiple times, but if zk is missing, the rollup just happens anyway.
+ZOOKEEPER_CLUSTER=NONE
+
+# Some tests based on HttpIntegrationTestBase use these values to ensure CORS is enabled and working on the endpoints
+# under test.
+CORS_ENABLED=true
+CORS_ALLOWED_ORIGINS=test.domain1.com, test.domain2.com, test.domain3.com
+CORS_ALLOWED_HEADERS=XYZ, ABC
+CORS_ALLOWED_METHODS=GET, POST, PUT
+CORS_ALLOWED_MAX_AGE=6000
+
+# This is to help with Travis, which intermittently fail the following tests due
+# to getting TimeoutException. This is done here because it needs to be before
+# RollupHandler is instantiated.
+ROLLUP_ON_READ_TIMEOUT_IN_SECONDS=20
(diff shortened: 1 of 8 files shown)


blueflood-http/src/integration-test/resources/blueflood.properties
```diff
@@ -0,0 +1,17 @@
+# No need for zookeeper for integration testing.
+# I believe this is because zk is only/primarily used to coordinator multiple rollup nodes so a single shard doesn't
+# get rolled up multiple times, but if zk is missing, the rollup just happens anyway.
+ZOOKEEPER_CLUSTER=NONE
+
+# Some tests based on HttpIntegrationTestBase use these values to ensure CORS is enabled and working on the endpoints
+# under test.
+CORS_ENABLED=true
+CORS_ALLOWED_ORIGINS=test.domain1.com, test.domain2.com, test.domain3.com
+CORS_ALLOWED_HEADERS=test.domain1.com, test.domain2.com, test.domain3.com
```
iWebi, line 10: Are the values same for HEADERS,METHODS and MAX_AGE config?

zzantozz: Oh, oops, good catch. Somehow when I read it, it looked like they were all set to the same values.

zzantozz: [Fixed](https://github.com/rackerlabs/blueflood/compare/18ab8f47aa1036f2cd5c7afedaee57719ddb673a..6fd203f11c8553e8a511760080be12096265e2aa). What's odd is the tests passed with this messed up. I guess they're not very thorough.

iWebi (approved):
👍

zzantozz:

PR #852, opened 2021-09-02
fixes clojars repo http url

zzantozz:
Later versions of maven quite sensibly refuse to download from http
urls by default. This fixes a repo url that's using http.


iWebi (approved):
👍

zzantozz:

PR #854, opened 2021-09-08
stabilizes an it's timing

zzantozz:
This test grabs a start and end time at construction time and uses
that for a time window during the test.  The problem is that there's a
lot of "before" stuff that happens in the superclass. At one timing,
over 4 seconds elapsed between test construction and the execution of
the test method, which puts the metrics outside the time window being
queried.  Moving the start and end into the test ensures we're using
an appropriate time window.


iWebi (approved):
👍

PR #855, opened 2021-09-08
Revert "Remove .travis.yml do to codecov tool security breach. .travi…

zzantozz:
…s.yml is not being used"

This reverts commit a3de44599ae6028d3c493c52c7e4b3a9bee95008.

This is an attempt to re-establish the blueflood CI pipeline.  I don't
see evidence that this build was using Codecov anyway.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,53 @@
+language: java
+jdk:
+  - oraclejdk8
+
+sudo: false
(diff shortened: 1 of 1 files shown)


zzantozz:
I think I'm not going to pursue the travis build and plan to move to GH Actions instead.

PR #857, opened 2021-09-29
adds concurrency test for LocatorCache

zzantozz:
The LocatorCache is intended to support high concurrency, with
multiple database-writing threads accessing it concurrently. None of
the existing tests prove that it can handle that. This adds such a
test and fixes the bug that comes from using a mutable object (the
removed LocationCacheEntry) in a multi-threaded environment.


iWebi (approved):
👍

PR #866, opened 2022-04-21
update readme badges

zzantozz:
I'm revamping the readme, and the first thing in it is some very
out-of-date badges. This cleans them up a little and makes them work
with some basic github workflows that'll be expanded later.

In doing so, I found that code coverage is only running for
integration tests, which probably isn't right. I added it for unit
tests, as well, and the workflow is doing only unit test coverage
right now. In the future, we'll need to be careful to keep the unit
test and integration test coverage separate.


iWebi (approved):
👍

PR #868, opened 2022-04-27
removes test jar building

zzantozz:
At one point, some few test classes were shared between core and http
modules and the integration tests. Those classes (with one exception,
addressed here) were moved to main source to avoid having to manage
test jars and dependencies. There's no need to build them anymore.


iWebi (approved):
[nuke](https://emoji.gg/emoji/nuke)

iWebi (approved):

PR #879, opened 2022-06-15
adds "getting started" configs

zzantozz:
Blueflood needs some basic configuration to run. Its default config
values don't let it start. I should maybe fix that. As it is, we need
some basic config files so that anyone can easily get Blueflood
running without having to think about the details of configuring a new
instance. This adds minimal configs that let the app start and log to
stdout.

Various places in the docs refered to a basic config, and some used
the config files from the "demo" directory. I think it's better to
have dedicated, minimal configs aimed specifically at a first
startup. The demo configs could fluctuate over time if the demo app
needs to adapt to different uses. I've updated all the references I've
found to point at these, new configs.

### DIFF
--- a/contrib/getting-started/blueflood.properties
+++ b/contrib/getting-started/blueflood.properties
@@ -1,0 +1,22 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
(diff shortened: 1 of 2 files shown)


adityacb15 (approved):
Approving Blueflood PR

rohitsngh27 (approved):

PR #890, opened 2022-07-25
Es rest helper instrumentation

iWebi:
Adds instrumentation for Elasticsearch Http client pool stats
Added ElasticsearchRestHelperTest test class with tests limited to instrumentation code. Test class needs further unit tests to assert round robin retries, max retry attempt etc scenarios.
Also introduced a config property to customize graphite metrics flush interval to test infrequent metrics ingestion on local setup.
Below is a sample Grafana Histogram graph with the new metrics included in this PR.
Screen Shot 2022-07-22 at 11 06 00 AM

### DIFF
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/utils/Metrics.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/utils/Metrics.java
@@ -110,7 +110,7 @@ public class Metrics {
                     .prefixedWith(config.getStringProperty(CoreConfig.GRAPHITE_PREFIX))
                     .build(graphite);
 
-            reporter.start(30l, TimeUnit.SECONDS);
+            reporter.start(config.getIntegerProperty(CoreConfig.GRAPHITE_REPORT_PERIOD_SECONDS), TimeUnit.SECONDS);
         } else {
             reporter = null;
         }
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/utils/Metrics.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/utils/Metrics.java
@@ -159,4 +159,12 @@ public class Metrics {
     public static Counter counter(Class kls, String... names) {
         return getRegistry().counter(MetricRegistry.name(kls, names));
     }
+
+    public static void registerGauge(Class kls, Gauge gauge, String ...names) {
+        // MetricRegistry does not offer getOrAdd wrapper for Gauge
+        String name = MetricRegistry.name(kls, names);
+        if(!getRegistry().getGauges().containsKey(name)) {
+            getRegistry().register(name, gauge);
+        }
+    }
 }
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/utils/Metrics.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/utils/Metrics.java
@@ -161,10 +161,7 @@ public class Metrics {
     }
 
     public static void registerGauge(Class kls, Gauge gauge, String ...names) {
-        // MetricRegistry does not offer getOrAdd wrapper for Gauge
         String name = MetricRegistry.name(kls, names);
-        if(!getRegistry().getGauges().containsKey(name)) {
-            getRegistry().register(name, gauge);
-        }
+        getRegistry().register(name, gauge);
     }
 }
--- a/blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/ElasticsearchRestHelper.java
+++ b/blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/ElasticsearchRestHelper.java
@@ -1,13 +1,13 @@
 package com.rackspacecloud.blueflood.io;
 
-import com.codahale.metrics.Histogram;
+import com.codahale.metrics.Counter;
+import com.codahale.metrics.Gauge;
 import com.google.common.annotations.VisibleForTesting;
 import com.rackspacecloud.blueflood.service.Configuration;
 import com.rackspacecloud.blueflood.service.CoreConfig;
 import com.rackspacecloud.blueflood.service.ElasticIOConfig;
 import com.rackspacecloud.blueflood.types.*;
 import com.rackspacecloud.blueflood.utils.GlobPattern;
-import com.rackspacecloud.blueflood.utils.Metrics;
 import org.apache.commons.lang.StringUtils;
 import org.apache.http.Header;
 import org.apache.http.HttpEntity;
@@ -21,7 +21,6 @@ import org.apache.http.impl.client.HttpClients;
 import org.apache.http.impl.conn.PoolingHttpClientConnectionManager;
 import org.apache.http.message.BasicHeader;
 import org.apache.http.nio.entity.NStringEntity;
-import org.apache.http.pool.PoolStats;
 import org.apache.http.util.EntityUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -30,6 +29,7 @@ import java.io.IOException;
 import java.util.*;
 
 import static com.rackspacecloud.blueflood.utils.Metrics.counter;
+import static com.rackspacecloud.blueflood.utils.Metrics.registerGauge;
 import static java.util.stream.Collectors.joining;
 
 public class ElasticsearchRestHelper {
@@ -45,10 +45,13 @@ public class ElasticsearchRestHelper {
             new BasicHeader("Accept", "application/json"),
             new BasicHeader("Content-Type", "application/json")
     };
-
-    private Histogram availablePoolSize = Metrics.histogram(getClass(), "Available Connections");
-    private Histogram pendingPoolSize = Metrics.histogram(getClass(), "Pending Connections");
-    private Histogram leasedPoolSize = Metrics.histogram(getClass(), "Leased Connections");
+    private Map<String,Counter> errorCounters = new HashMap<String, Counter>() {{
+       put("refreshIndex", counter(getClass(), "refreshIndex Error"));
+       put("fetchEvents", counter(getClass(), "fetchEvents Error"));
+       put("fetchTokenDocuments", counter(getClass(), "fetchTokenDocuments Error"));
+       put("fetchDocuments", counter(getClass(), "fetchDocuments Error"));
+       put("index", counter(getClass(), "index Error"));
+    }};
 
     /**
      * Gets an instance of the ES rest helper for use in main source.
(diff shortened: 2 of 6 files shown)


blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/ElasticsearchRestHelper.java
```diff
@@ -42,6 +46,10 @@ public class ElasticsearchRestHelper {
             new BasicHeader("Content-Type", "application/json")
     };
 
+    private Histogram availablePoolSize = Metrics.histogram(getClass(), "Available Connections");
+    private Histogram pendingPoolSize = Metrics.histogram(getClass(), "Pending Connections");
+    private Histogram leasedPoolSize = Metrics.histogram(getClass(), "Leased Connections");
```
zzantozz, line 51: So when you asked about histogram vs gauge, I thought you were talking about the visualization. Aren't these more appropriately gauge metrics, in that they track distinct values that go up and down? A histogram metric is more geared toward collecting values into statistical buckets, I think. I'm not really sure what the difference is in how these are reported, though.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/ElasticsearchRestHelper.java
```diff
@@ -464,8 +472,10 @@ public int refreshIndex(String indexName) throws IOException {
         HttpGet httpGet = new HttpGet(url);
         CloseableHttpResponse response = null;
         try {
+            this.publishPoolStats();
             response = closeableHttpClient.execute(httpGet);
         } catch (IOException e) {
+            counter(getClass(), "refreshIndex Error").inc();
```
zzantozz, line 478: Should this be assigned to a field like the other metrics here? That seems to be the convention in the project. Not sure if there's any performance overhead to obtaining it on demand like this.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/ElasticsearchRestHelper.java
```diff
@@ -499,13 +509,15 @@ private ExecuteResponse executePost(String urlFormat, String queryDslString, Str
...
             try {
+                this.publishPoolStats();
                 response = closeableHttpClient.execute(httpPost);
                 responseCode = response.getStatusLine().getStatusCode();
                 HttpEntity entity = response.getEntity();
                 String str = EntityUtils.toString(entity);
                 EntityUtils.consume(entity);
                 return new ExecuteResponse(str, responseCode);
             } catch (Exception e) {
+                counter(getClass(), String.format("%s Error", methodName)).inc();
```
zzantozz, line 520: This could also be turned into a field that's a map of `methodName -> counter`.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/ElasticsearchRestHelper.java
```diff
@@ -464,8 +472,10 @@ public int refreshIndex(String indexName) throws IOException {
         HttpGet httpGet = new HttpGet(url);
         CloseableHttpResponse response = null;
         try {
+            this.publishPoolStats();
```
zzantozz, line 475: It doesn't feel right to be updating these metrics on each call. A gauge would sample these values at a set interval, outside the normal flow of control. Is there a reason we want to update them on every call vs periodic sampling?

blueflood-elasticsearch/src/test/java/com/rackspacecloud/blueflood/io/ElasticsearchRestHelperTest.java
```diff
@@ -0,0 +1,87 @@
...
+    // HttpClient dependency is nested inside pool manager build logic.
+    // Mocking Http response requires mocking multiple instances of Apache http client resources.
+    mockStatic(HttpClients.class);
+    when(HttpClients.custom()).thenReturn(mockBuilder);
+    when(mockBuilder.setConnectionManager(any())).thenReturn(mockBuilder);
+    when(mockBuilder.build()).thenReturn(mockHttpClient);
+    when(mockHttpResponse.getStatusLine()).thenReturn(statusLine);
+
+    resetMetricRegistry();
+    helper = ElasticsearchRestHelper.getInstance();
```
zzantozz, line 47: The `getInstance` method will hopefully change to returning a static singleton in the future. I created the `getConfigurableInstance` method for unit tests, where you always want to obtain a fresh instance, in case you need to change the app config and get a new instance of the rest helper.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/ElasticsearchRestHelper.java
iWebi: Agree. Gauge is correct choice. With few random simulated values confirmed we can view the trend as well as current value (if we need it
Screen Shot 2022-07-27 at 8 33 20 AM
)

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/ElasticsearchRestHelper.java
iWebi: Not a significant hit to performance as the metric needs to be created/lookedup only for Exception scenarios. But it does make sense make it instance field. Fixed it.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/utils/Metrics.java
```diff
@@ -159,4 +159,12 @@ public static Histogram histogram(Class kls, String... names) {
     public static Counter counter(Class kls, String... names) {
         return getRegistry().counter(MetricRegistry.name(kls, names));
     }
+
+    public static void registerGauge(Class kls, Gauge gauge, String ...names) {
+        // MetricRegistry does not offer getOrAdd wrapper for Gauge
```
zzantozz, line 164: Isn't that because it would be an error to register the same gauge more than once? With the code you've added here, if someone were to create a second gauge with the same name somewhere else, it would be silently dropped.

iWebi: MetricRegistry#getOrAdd does the same. It ignores if its the same Metric to be registered. getOrAdd has protected access specifier. So I had to write similar wrapper for Gauge.
Gauge name includes the class name to make it namespace specific. So duplicate chances are less. If it still ends up duplicate, it gets ignored. Similar to other metric types.

iWebi: aah I see the concern for Gauge. If another Gauge with different getValue implementation is added, it gets silently ignored. But the chances of collision are less due to Class name part of the Gauge name. Do you suggest raise Exception if it already exists?

zzantozz: You're right that the chances of collision are low. It seems that the MetricRegistry already prevents duplicate registrations:
```
            if (existing == null) {
                onMetricAdded(name, metric);
            } else {
                throw new IllegalArgumentException("A metric named " + name + " already exists");
            }
```
so I'd just let it do its thing and not over-code here.

zzantozz: I'm pretty sure the reason there's not a comparable "getOrAdd" function for gauges is that they're fundamentally different than other metric types. They can be looked up and updated at any time to send new data, but a gauge is always created once and passively monitors a thing. There's no reason to ever "get" a gauge after you've registered it.

iWebi: Unfortunately we can't use this as is. We have 3 instances of the ElasticsearchRestHelper. So 3 constructor calls->3 times metrics registration attempts leading to exceptions. Unless we make ElasticsearchRestHelper singleton, we can't use MetricRegistry#register as is. I have to put try/catch around every gauge registration otherwise.

zzantozz: Reminder: we decided in Zoom to go ahead and change this to a static singleton, and we'll try to get a feel for any performance issues in staging.

blueflood-integration-tests/src/integration-test/java/com/rackspacecloud/blueflood/io/ElasticIOIntegrationTest.java
```diff
@@ -67,7 +67,7 @@ public static void startElasticsearch() {
 
     @Before
     public void setup() throws Exception {
-        helper = ElasticsearchRestHelper.getConfigurableInstance();
+        helper = ElasticsearchRestHelper.getInstance();
```
zzantozz, line 70: You ran this test with the singleton change? I thought I remembered tests that seemed like they'd fail if this always returned the same object.

blueflood-elasticsearch/src/test/java/com/rackspacecloud/blueflood/io/ElasticsearchRestHelperTest.java
```diff
@@ -0,0 +1,78 @@
...
+    resetMetricRegistry();
+  }
+
+  @Test
+  public void testConstructor_registersHttpConnectionStatsGauges() {
+    // when
+    helper = ElasticsearchRestHelper.getInstance();
+    // then
+    Set<String> registeredGauges = Metrics.getRegistry().getGauges().keySet();
+    assertTrue(registeredGauges.contains(name(ElasticsearchRestHelper.class, "Available Connections")));
```
zzantozz, line 56: Hamcrest is much better for this. Something like
```
assertThat(registeredGauges, hasItem(name(ElasticsearchRestHelper.class, "Available Connections")));
```
It reads more fluently and gives much more detailed failure messages. See the ES integration test for examples.

blueflood-elasticsearch/src/test/java/com/rackspacecloud/blueflood/io/ElasticsearchRestHelperTest.java
```diff
@@ -0,0 +1,78 @@
...
+
+  @Test
+  public void testElasticsearchPostErrors_incrementsErrorCounter() throws IOException {
+    // Ensure Error counter is 0 prior to failures simulation
+    helper = ElasticsearchRestHelper.getInstance();
+    Counter fetchEventsErrorCounter = helper.getErrorCounters().get("fetchEvents");
+    Assert.assertEquals(0, fetchEventsErrorCounter.getCount());
+    when(mockHttpClient.execute(any())).thenThrow(new IOException("test exception"));
+    helper.fetchEvents("fooErrorTenant", new HashMap<>());
+    assertTrue(fetchEventsErrorCounter.getCount() > 0);
```
zzantozz, line 69: Also a case for hamcrest
```
assertThat(fetchEventsErrorCounter.getCount(), greaterThan(0));
```
Basically never use junit's assertions.

blueflood-integration-tests/src/integration-test/java/com/rackspacecloud/blueflood/io/ElasticIOIntegrationTest.java
iWebi: Yes. It ran successful. just to make sure I ran it multiple times. It did not complain.
```
Tests run: 39, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 54.647 sec - in com.rackspacecloud.blueflood.io.ElasticIOIntegrationTest

Results :

Tests run: 39, Failures: 0, Errors: 0, Skipped: 0
```

zzantozz (approved):

PR #895, opened 2022-07-26
add a minikube env/config

zzantozz:
Building from the existing Docker image and docker-compose
sub-projects, this adds a Kubernetes yaml file and instructions for
running it in minikube. It's very basic and needs some changes before
it could be deployed to another k8s cluster. Specifically:

1. The file mounts need to change to ConfigMaps, or else find another
way to configure the services, like the Grafana dashboards that are
available at startup here.

2. All the components are configured as simple Pods. They should be
Deployments or StatefulSets for a fault-tolerant, production workload.


contrib/blueflood-minikube/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+Then, if you want to run the latest Blueflood image from DockerHub, you can go ahead and start the stack with `minikube
+apply -f blueflood.yaml`. If you prefer to run a locally built version of Blueflood, you can build a Docker image inside
+minikube. You just need to set a couple of variables to do that. First, run `minikube docker-env`, and note the values
+of `DOCKER_HOST` and `DOCKER_CERT_PATH` for use in the following commands. Replace `tcp://` in `DOCKER_HOST` with
+`https://` when you use it below. Then you can build blueflood and run it with
+
+```bash
+mvn package docker:build -DdockerHost=$MODIFIED_DOCKER_HOST -DdockerCertPath=$DOCKER_CERT_PATH
+minikube apply -f blueflood.yaml
```
iWebi, line 29: Does this need to be `minikube kubectl -- apply -f blueflood.yaml` and run from ${projectRootDir}/blueflood/contrib/blueflood-minikube location? minicube did not recognize `apply` argument. Perhaps there was an alias? Also does the command need to specify path of blueflood.yaml?

iWebi (dismissed):
🥇 
Except one observation about `minikube apply` usage, everything went smooth to get this running on my machine.

contrib/blueflood-minikube/README.md
zzantozz: No you're right. That's supposed to be `kubectl`. [Fixed](https://github.com/rax-maas/blueflood/compare/5cf19a1c1ce206b8a5ab0a6f0fbdf28eb43d3f8b..01933f15e8b7e99088d344516c92e2fa745cbcc4?diff=unified). I also clarified the expected paths.

iWebi (approved):

PR #899, opened 2022-08-08
add stats to discovery metrics

zzantozz:
The current emitted metrics only track batches of metrics processed by
discovery modules. The actual count of metrics isn't available. This
can cause some misleading graphs. A thousand batches being handled
here could mean one metric per batch or a thousand metrics per
batch. It can be important to know the difference.


iWebi (approved):
PROverflowException !

PR #900, opened 2022-08-08
ensure token caching in tests

zzantozz:
I saw unexpected tokens being written under load. It seemed the caches
weren't preventing discovery writes as expected. This addition to the
test helps to verify that caching is working as expected.


iWebi (approved):
PRRateLimitOutOfBounds

PR #901, opened 2022-08-10
Added workflows for on pr action and release

rohitsngh27:
Created github workflow to be triggered when PR is created and another one a manual workflow to create an automatic blueflood release.


zzantozz (changes requested):
This commit history is unacceptable. You literally have more commits than lines changed in the PR. Most of the commits are garbage. Temporary commits while you're exploring and researching are fine, but now you need to write some good commits. Every commit needs to have a distinct purpose and a message that completely describes that purpose. Turn this change into one or more commits like that, and then open a PR for each commit. Size-wise, this is fine for a single commit and PR, but I see five or six different changes here that could be separate commits and PRs.

PR #903, opened 2022-08-10
Added github action to trigger when PR is created

rohitsngh27:
We need to create an automated check which tests for unit and integration test cases whenever a new pull request is created. This is required because automated PR checks helps in detecting any failures which new changes might bring in.
So, We are using github action `on pull_request` trigger and maven to perform tests and generate reports.


.github/workflows/on-pr.yml
```diff
@@ -0,0 +1,41 @@
+name: "Validate and plan"
```
zzantozz, line 1: This isn't a good name for this workflow. This is a copy/paste from the terraform project I pulled the skeleton from.

.github/workflows/on-pr.yml
```diff
@@ -0,0 +1,41 @@
...
+
+on:
+  pull_request:
```
zzantozz, line 4: btw, there's already a simple [pull_request worfkflow in this repo](https://github.com/rax-maas/blueflood/blob/master/.github/workflows/unit-test.yml#L6). You'll want to do something with that.

.github/workflows/on-pr.yml
```diff
@@ -0,0 +1,41 @@
...
+  test:
+    runs-on: ubuntu-18.04
+    strategy:
+      matrix:
+        java-version: [ 8 ]
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-java@v2
+        with:
+          distribution: 'adopt'
```
zzantozz, line 16: Other workflows here are using temurin. Why adopt? One way or the other, we should be consistent. I haven't kept up with the different JDKs since they split.

.github/workflows/on-pr.yml
```diff
@@ -0,0 +1,41 @@
...
+          java-version: ${{ matrix.java-version }}
+          cache: 'maven'
+      - name: Configure git user
+        run: |
+          git config user.email "actions@github.com"
+          git config user.name "GitHub Actions"
```
zzantozz, line 22: There don't seem to be any git steps in this workflow, so this step is unnecessary.

.github/workflows/on-pr.yml
```diff
@@ -0,0 +1,41 @@
...
+      - name: Build and test
+        run: |
+          mvn clean install -Dmaven.javadoc.skip --quiet
```
zzantozz, line 25: Why skip javadoc? I don't think there's anything configured to build it on install.

.github/workflows/on-pr.yml
```diff
@@ -0,0 +1,41 @@
...
+          mvn jacoco:report jacoco:report-integration
+          mvn package -P all-modules
```
zzantozz, line 27: Why run `package` again? You already did an `install`, which includes `package`. Strictly speaking, to run tests, you only need to run to the `verify` phase anyway. Also, the `all-modules` profile was removed some time ago.

zzantozz (changes requested):

zzantozz (changes requested):

.github/workflows/on-pr.yml
```diff
@@ -0,0 +1,28 @@
+name: "Run tests on pull requests"
+
+on:
+  pull_request:
+
+jobs:
+  test:
```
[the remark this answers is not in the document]
rohitsngh27: So, should I rename it to something like `build_and_test`?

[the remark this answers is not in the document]
zzantozz: Well, this workflow runs unit and integration tests on every PR. The other one runs only unit tests. If this is the workflow intended to validate a PR before merging, then the other workflow doesn't even need to run.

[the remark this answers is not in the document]
rohitsngh27: Yeaa...That's what I was thinking. Maybe I'll remove the other workflow since this one is doing the validation of PR.

zzantozz (approved):

PR #905, opened 2022-08-10
add discovery throttling

zzantozz:
The new Elasticsearch rest code seems to put more load on
Elasticsearch than the old code. Combined with heavy load, multiple
production nodes, and too-short cache TTLs, deploying new ingest code
and restarting is overloading Elasticsearch and affecting the entire
system, including other ingest nodes and query node performance.

This adds throttling to cap the number of discovery writes that will
be made per minute. Setting it lower reduces the impact on
Elasticsearch. There's both a global and per-tenant throttle.

The throttles are implemented here using Guava caches set to a
1-minute TTL. Local profiling of a similar cache suggests that the
cache itself uses approximately 70M of memory per million cache
entries, excluding the size of the entries.

### DIFF
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/processors/DiscoveryWriter.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/processors/DiscoveryWriter.java
@@ -17,6 +17,10 @@
 package com.rackspacecloud.blueflood.inputs.processors;
 
 import com.codahale.metrics.Meter;
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.rackspacecloud.blueflood.cache.LocatorCache;
 import com.rackspacecloud.blueflood.concurrent.FunctionWithThreadPool;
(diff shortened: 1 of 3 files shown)


samwar (approved):
The tests look sound, and the implementation does indeed appear to throttle the discovery input.

blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/processors/DiscoveryWriter.java
```diff
@@ -91,7 +138,11 @@ private static List<IMetric> condense(List<List<IMetric>> input) {
             }
 
             for (IMetric m : list) {
-                if (!LocatorCache.getInstance().isLocatorCurrentInDiscoveryLayer(m.getLocator())) {
+                boolean isAlreadySeen = LocatorCache.getInstance().isLocatorCurrentInDiscoveryLayer(m.getLocator());
+                boolean isThrottlingGlobally = newLocatorsThrottle.size() >= maxNewLocatorsPerMinute;
```
iWebi, line 142: global throttle check can be done once for all metrics, isn't it?

iWebi (approved):
👍

blueflood-core/src/main/java/com/rackspacecloud/blueflood/inputs/processors/DiscoveryWriter.java
zzantozz: You're right. Good catch. Fixed in #913

PR #906, opened 2022-08-12
Added manual GitHub action trigger to create release package.

rohitsngh27:
We need a trigger which can be used to create an automated release package for blueflood. We are using GitHub action with maven release plugin here to create the release package.
Maven release package helps in performing sequence of automated steps which will run tests on latest master code and then will create the tag which will have the release version which we specified when triggering the workflow. It will also increment the next snapshot version in all the moms wherever required.
A PR will be created as part of this workflow which will have the pom version changes of incremented snapshot version for new development.
Artifacts will be copied to the release package as well.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,56 @@
+name: "Manual Release workflow"
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: Release version
+        required: true
+
+jobs:
+  tagged-release:
+    runs-on: ubuntu-18.04
+    strategy:
+      matrix:
+        java-version: [ 8 ]
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          ref: master
+      - uses: actions/setup-java@v2
+        with:
+          distribution: 'temurin'
+          java-version: ${{ matrix.java-version }}
+          cache: 'maven'
+      - name: Configure Git user
+        run: |
+          git config user.email "actions@github.com"
+          git config user.name "GitHub Actions"
+          git remote -v
+      - name: Set version in env
+        run: |
+          version="${GITHUB_REF#refs/tags/}"
+          echo "RELEASE_VERSION=${{ github.event.inputs.version }}" >> $GITHUB_ENV
+      - name: Generate changelog
+        id: changelog
+        uses: metcalfc/changelog-generator@v0.4.4
+        with:
+          myToken: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build and deploy
+        run: |
+          pwd
+          git checkout -b release/${RELEASE_VERSION}
+          mvn -B release:prepare -DreleaseVersion=${RELEASE_VERSION} -Dmaven.wagon.httpconnectionManager.ttlSeconds=60
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Create pull request
+        uses: repo-sync/pull-request@v2
+        with:
+          source_branch: "release/${{ github.event.inputs.version }}"
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Copy Artifacts
+        uses: ncipollo/release-action@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -240,6 +240,27 @@
...
+            <id>sign-artifacts</id>
+            <phase>verify</phase>
+            <goals>
+              <goal>sign</goal>
+            </goals>
+          </execution>
+        </executions>
+        <configuration>
+          <skip>true</skip>
+        </configuration>
```
zzantozz, line 257: Why have the plugin here if its executions are skipped?

.github/workflows/release.yml
```diff
@@ -0,0 +1,56 @@
...
+      - uses: actions/setup-java@v2
+        with:
+          distribution: 'temurin'
+          java-version: ${{ matrix.java-version }}
+          cache: 'maven'
+      - name: Configure Git user
+        run: |
+          git config user.email "actions@github.com"
+          git config user.name "GitHub Actions"
+          git remote -v
```
zzantozz, line 29: This only shows output. It's probably left over from debugging. You probably don't want to keep it here.

.github/workflows/release.yml
```diff
@@ -0,0 +1,56 @@
...
+      - name: Set version in env
+        run: |
+          version="${GITHUB_REF#refs/tags/}"
+          echo "RELEASE_VERSION=${{ github.event.inputs.version }}" >> $GITHUB_ENV
```
zzantozz, line 33: The first line of this step is wrong now, since there's no tag. There should be no need for this step at all anymore, though, since the version is explicitly provided as input to the workflow.

.github/workflows/release.yml
```diff
@@ -0,0 +1,56 @@
...
+      - name: Generate changelog
+        id: changelog
+        uses: metcalfc/changelog-generator@v0.4.4
+        with:
+          myToken: ${{ secrets.GITHUB_TOKEN }}
```
zzantozz, line 38: You're not using the changelog anywhere. Either include it in the created PR or remove this build step.

.github/workflows/release.yml
```diff
@@ -0,0 +1,56 @@
...
+          git checkout -b release/${RELEASE_VERSION}
+          mvn -B release:prepare -DreleaseVersion=${RELEASE_VERSION} -Dmaven.wagon.httpconnectionManager.ttlSeconds=60
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Create pull request
+        uses: repo-sync/pull-request@v2
+        with:
+          source_branch: "release/${{ github.event.inputs.version }}"
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Copy Artifacts
```
zzantozz, line 51: This isn't just copying artifacts. It's creating a new github release.

.github/workflows/release.yml
```diff
@@ -0,0 +1,56 @@
...
+        uses: ncipollo/release-action@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          artifacts: /home/runner/work/blueflood/blueflood/blueflood-all/target/blueflood-all-${{ github.event.inputs.version }}-jar-with-dependencies.jar
+          tag: "blueflood-${{ github.event.inputs.version }}"
```
zzantozz, line 56: This tag doesn't exist, right? Does the release action create it?

zzantozz, line 56: Also, angry red marker for no newline.

pom.xml
```diff
@@ -272,23 +293,6 @@
     </profile>
     <profile>
       <id>release</id>
```
zzantozz, line 295: So there's no an empty profile here now. Any reason to keep it around?

zzantozz (changes requested):

pom.xml
rohitsngh27: Removed the plugin.

.github/workflows/release.yml
rohitsngh27: Removed this step and using `${{ github.event.inputs.version }}` wherever needed.

.github/workflows/release.yml
rohitsngh27: The tag is created by maven release prepare step above.

zzantozz (approved):

PR #912, opened 2022-08-15
remove excess token discovery code

zzantozz:
When someone added token indexing alongside the name indexing, they
didn't do a good job integrating it with the existing code. They
basically duplicated DiscoveryWriter and all the things it does. It's
much more efficient and logical for token discovery to live behind the
DiscoveryWriter in a new module alongside the module that provides
name discovery. DiscoveryWriter already knows how to load multiple
discovery modules and use them. This also lets all discovery
activities fall under the DiscoveryWriter and any features it
provides, like caching and throttling to reduce load on discovery
backends. As it stands now, any such improvments would have to be
duplicated in TokenDiscoveryWriter.

The main goal of this change is to remove the duplicate Token-related
classes/interfaces and move the ElasticTokensIO, which is the real
logic of the token indexing, into the DiscoveryWriter. To keep the
change as minimal as possible, I haven't reworked all the tests that
use the existing ElasticTokensIO contract, so I had to add a hacky
`insertDiscovery(List)` method there to handle either a list of
metrics or a list of tokens. Only tests will pass a list of tokens
there. At runtime, it's always a list of metrics, passed by the
DiscoveryWriter. I'll clean this up in a following change.

To implement the change, a small bit of code that manages the token
cache has to move down from TokenDiscoveryWriter to ElasticTokensIO,
and then ElasticTokensIO changes to a normal DiscoveryIO that's
plugged into the DiscoveryWriter.


blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/ElasticTokensIO.java
```diff
@@ -45,22 +48,56 @@ public ElasticTokensIO() {
...
+     * A hacky de-generified implementation of two semi-compatible methods from the interfaces implemented here. One
+     * interface expects a list of metrics and the other a list of tokens. This is from an unfortunate copy/paste
+     * incident in the dark past. The {@link com.rackspacecloud.blueflood.inputs.processors.DiscoveryWriter} calls this
+     * to create discovery data for metrics as they're ingested. Originally, it was a bad copy of that class that called
+     * this and passed a list of tokens. Now this lives behind DiscoveryWriter as a discovery module, as appropriate,
+     * and it receives a list of metrics, as it should. Tests still exist that send it a list of tokens. Once those
+     * tests are fixed, this method can change to its correct signature of accepting a list of metrics.
+     */
+    public void insertDiscovery(List things) throws IOException {
+        if (things.get(0) instanceof Token) {
```
iWebi, line 72: Existing code had a size() == 0 check. Do we need it to avoid IndexOutOfBoundsException?

blueflood-elasticsearch/src/test/java/com/rackspacecloud/blueflood/io/ElasticTokensIOTest.java
```diff
@@ -1,27 +1,39 @@
...
+    protected ElasticTokensIO elasticTokensIO;
+    private ElasticsearchRestHelper mockElasticsearchRestHelper;
 
-    @BeforeClass
-    public static void setup() throws IOException {
+    @Before
+    public void setup() throws IOException {
         elasticTokensIO = new ElasticTokensIO();
+        mockElasticsearchRestHelper = mock(ElasticsearchRestHelper.class);
+        elasticTokensIO.setElasticsearchRestHelper(mockElasticsearchRestHelper);
```
iWebi, line 36: TokenCache.getInstance().resetCache(); may be needed to make tests happy.

iWebi (changes requested):
Once tests pass, we can merge this.

blueflood-elasticsearch/src/main/java/com/rackspacecloud/blueflood/io/ElasticTokensIO.java
iWebi: never mind. Usages of this method always have non-empty list input.

zzantozz: Good call. It does work fine right now, but I'm all for guarding against future misuse. [Fixed](https://github.com/rax-maas/blueflood/compare/61529b6eb6fb894c2ff360bb3b628dd73b71edf3..c009776e134915e5417c6642419dd20ea740f894#diff-4e793beb5eafe1d535bc7b3ec8a9bd32a6745f0b51a846e91a38bf0f7401da9bR161-R165).

blueflood-elasticsearch/src/test/java/com/rackspacecloud/blueflood/io/ElasticTokensIOTest.java
zzantozz: [Fixed](https://github.com/rax-maas/blueflood/compare/61529b6eb6fb894c2ff360bb3b628dd73b71edf3..c009776e134915e5417c6642419dd20ea740f894#diff-4e793beb5eafe1d535bc7b3ec8a9bd32a6745f0b51a846e91a38bf0f7401da9bR35). Tests should be happy now.

iWebi (approved):

PR #918, opened 2022-08-29
Ignoring Before and AfterClass annotations

richarxt:
Comment out BeforeClass and AfterClass annotations because it is still running initialization thought we had Ignored the class.


richarxt:
It is a miss..it runs fine.

PR #920, opened 2022-09-06
let env vars set config properties

zzantozz:
As it stands, configuration items have default values which can be overriden with a config file, and both can be overriden by setting system properties. For containerization, it'll be important to also let environment variables be used. This adds environment variables as the highest-precedence configuration override, so we now have four levels to find a config item. In precedence order, they are: environment variables, system properties, config file, defaults.


iWebi (approved):
JUnit means loooooong method names :)

zzantozz:
lol, yeah. I thought some of those names were a little excessive, but they're certainly descriptive!

PR #924, opened 2022-09-06
allow docker image to take config files

zzantozz:
Through working with the image and deploying to Kubernetes, it became apparent that this Docker image was made with limited, dev-style deployments in mind. In particular, we don't want to have to configure everything via environment variables. We want to have a config file that holds most of our Blueflood settings and override a few things with the environment.

This makes the Docker image accept an existing file location for the Blueflood config and the log4j config. If either is missing, it falls back to its original behavior of: 1) generating a Blueflood config file by dumping all environment variables to a file and 2) writing a static log4j config file where only the log level of the Blueflood classes is configurable by the user.


contrib/blueflood-docker/docker-entrypoint.sh
```diff
@@ -70,13 +70,19 @@ if [ "$INIT_ELASTICSEARCH" != false ]; then
...
-printenv > blueflood.conf
+if [ -z "$BLUEFLOOD_CONF_LOCATION" ]; then
+  BLUEFLOOD_CONF_LOCATION="./blueflood.conf"
+  printenv > blueflood.conf
+fi
 
-cat > blueflood-log4j.properties << EOL
+if [ -z "$BLUEFLOOD_LOG4J_CONF_LOCATION" ]; then
+  BLUEFLOOD_LOG4J_CONF_LOCATION="./blueflood-log4j.properties"
+  : ${BLUEFLOOD_LOG_LEVEL=INFO}
```
iWebi, line 80: 👍  learnt something new. colon syntax

iWebi (approved):

contrib/blueflood-docker/docker-entrypoint.sh
zzantozz: That's something I learned about recently too :D

PR #929, opened 2022-09-15
Added changelogs as part of GitHub release workflow.

rohitsngh27:
Created changelog using a metcalfc/changelog-generator and then adding it to a file. Also including it as part of github release which will be created when the workflow runs.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -26,10 +26,19 @@ jobs:
         run: |
           git config user.email "actions@github.com"
           git config user.name "GitHub Actions"
+      - name: Generate changelog
+        id: changelog
+        uses: metcalfc/changelog-generator@v0.4.4
+        with:
+          myToken: ${{ secrets.GITHUB_TOKEN }}
       - name: Build and deploy
         run: |
           pwd
           git checkout -b release/${{ github.event.inputs.version }}
+          echo "${{ steps.changelog.outputs.changelog }}" > CHANGELOG.md
+          git add CHANGELOG.md
+          git commit -m "Adding changelogs"
+          git push origin release/${{ github.event.inputs.version }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -44,3 +53,4 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
           artifacts: /home/runner/work/blueflood/blueflood/blueflood-all/target/blueflood-all-${{ github.event.inputs.version }}-jar-with-dependencies.jar
           tag: "blueflood-${{ github.event.inputs.version }}"
+          body: "${{ steps.changelog.outputs.changelog }}"
```
zzantozz, line 56: Something here doesn't look right. Either the changelog file will always have only the latest changes in it every time, or every release will have the entire history of the project in it. The changelog file and release body should be treated differently.

rohitsngh27: Changelog file does have the latest changes in it. What do we want in release body?

zzantozz: Only the changes for that release, right?

rohitsngh27: which means all the commits since last release?

zzantozz: Well, with the way we're going now, yes. What's important is that a release notifies users what's new in the release.

rohitsngh27: So I tested it on my fork and release body does have the commits list since last release.

zzantozz: So do you want to go ahead and merge this like this and then figure out where to go from there?

zzantozz (approved):

.github/workflows/release.yml
rohitsngh27: Sure... I will figure out a way.

PR #939, opened 2022-09-16
roll out k8s resources - example overlay

zzantozz:
This is one in a series of changes that adds Kubernetes resources under contrib/blueflood-k8s. These resources make it simple to deploy Blueflood to any Kubernetes cluster.

This adds an "example" Kustomize overlay and documentation for deploying and initializing Cassandra and Elasticsearch clusters for Blueflood in Kubernetes.

It's a *LARGE CHANGE*, but most of it is config files which were copied out of the Cassandra and Elasticsearch containers. They're only changed slightly to make the clusters deploy. Most all of the files in the config directories are just the defaults. The important parts are the README instructions and kustomization.yaml files.


iWebi (approved):
💯

PR #943, opened 2022-09-22
de-static-ify DatastaxIO

zzantozz:
Much of Blueflood's core code is built as static singletons, but DatastaxIO is just pure static. It connects to the database at class initialization time, making it impossible to touch this thing without having a database up and running. I want to add the ability to initialize the database schema from code, but I can't do that because this requires the keyspace to already exist just to load the class!

The only important method here is `getSession()`, which all other code uses to get access to the database. As long as that method's contract is preserved, we can make changes to the rest of the class around it.

This changes DatastaxIO to be a static singleton in line with the rest of Blueflood and leaves the constructor public so that other code can get access to the database when necessary. There are no unit tests of this class, but it's the central access point for all database traffic that uses the Datastax driver, so integration tests will certainly fail if there's a problem here.

### DIFF
--- a/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxIO.java
+++ b/blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxIO.java
@@ -35,23 +35,60 @@ import java.util.Set;
  * to use datastax driver can use this class to get a {@link com.datastax.driver.core.Session}
  * object for the read/write statements to Cassandra.
  */
-public class DatastaxIO {
+public class DatastaxIO implements AutoCloseable {
     private static final Logger LOG = LoggerFactory.getLogger(DatastaxIO.class);
-    private static final IOConfig ioconfig = IOConfig.singleton();
     private static final String metricsPrefix = "datastax.";
 
-    private static  Cluster cluster;
-    private static Session session;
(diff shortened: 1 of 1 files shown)


blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxIO.java
```diff
@@ -172,12 +211,26 @@ public void onTimerRemoved(String name) {
         });
     }
 
+    @Override
     public void close() { //Not to be used with time-series data.
         session.close();
```
iWebi, line 216: Not sure if we need null check here as session/cluster is not guaranteed to be initialized when class is loaded.

iWebi (approved):

blueflood-core/src/main/java/com/rackspacecloud/blueflood/io/datastax/DatastaxIO.java
zzantozz: `cluster` and `session` are final and set by the constructor. There's no way they could be null by the time you could call `close()`. Actually, things would crash before you ever got that far because `cluster` is used by the monitoring and logging stuff during the constructor call. If it's null, it fails then. I saw it happen while I was working this change out.

PR #966, opened 2024-01-31
Update the ubuntu image for github action

richarxt:
github action doesn't support ubuntu 18.04. So we updated teh image to latest ubuntu version provided the github action.


zzantozz:
This needs attention

